PR #2, opened 2017-03-20
Add get_env and file permission utilities

mikaelarguedas:
This pushes some basic function to c_utilities. These functions could then be use to replace their rcl/rmw implementations (c.f. https://github.com/ros2/rmw/pull/87).

Labeling this as in progress because no tests attached.

Feel free to give feedback on the PR even if still in progress!


dirk-thomas:
The functions should have a detailed docblock in the header files since at least some of them need to clarify memory allocation and ownership of pointers.

Karsten1987:
any updates on that?

mikaelarguedas:
didn't have time to look into adding dummy files with different permissions for testing. Do you need it soon ? functions themselves are working AFAIK (that's what I'm relying on for the sros demo) you can use this branch if you need to leverage them

wjwwood:
As for documentation (@dirk-thomas mentioned this above), you can probably base it on the existing "get env" function that was to be migrated to rmw:

https://github.com/ros2/rmw/pull/87/files#diff-edc2fac2b38b652a58db78bce99d39ccR26

Karsten1987:
I am going to touch quite some RMW code and I was wondering if I should take care of these changes in the same time.

regarding `get_env`, is there a difference between  https://github.com/ros2/rmw/blob/8d116dca3d932a50badd7f97915577ef2477da32/rmw/include/rmw/impl/getenv.h#L46 and this https://github.com/ros2/rmw_implementation/blob/master/rmw_implementation/src/functions.cpp#L30

wjwwood:
I think the idea was to use the common one once it is available:

https://github.com/ros2/rmw_implementation/pull/17#discussion_r94687328

mikaelarguedas:
@Karsten1987 Feel free to push on this branch any changes to get it in and we can then start refactoring every different implementation of these functions to use this common one instead. If you don't have time I'll try to address this on Monday so that you can use it in your changes. You can use this branch for testing your rmw changes in the meantime

include/c_utilities/get_env.h
```diff
@@ -0,0 +1,35 @@
...
+{
+#endif
+
+#include <stdbool.h>
+
+#include "c_utilities/visibility_control.h"
+
+C_UTILITIES_PUBLIC
+bool
+get_env(const char * env_name, const char ** env_value);
```
wjwwood: Also there are no tests for this. They could be scavenged from:

https://github.com/ros2/rmw/pull/87/files#diff-4389729cf437a8deb24b70ef12a2b57e

include/c_utilities/file_permissions.h
```diff
@@ -0,0 +1,51 @@
...
+{
+#endif
+
+#include <stdbool.h>
+
+#include "c_utilities/visibility_control.h"
+
+C_UTILITIES_PUBLIC
+bool
+is_directory(const char * abs_path);
```
wjwwood: It's not a great start for this general purpose library to not have any docstrings or tests for these functions... I really think sometime should be spent on this while we're looking at these functions and their design is fresh in our minds.

src/file_permissions.c
```diff
@@ -0,0 +1,110 @@
...
+bool
+is_file_readable_and_writable(const char * file_abs_path)
+{
+  struct stat buf;
+  if (!file_exists(file_abs_path)) {
+    return false;
+  }
+  stat(file_abs_path, &buf);
+#ifdef WIN32
+  // NOTE(marguedas) on windows all files are readable
```
wjwwood: If this is the case, then why does `is_file_readable` do any checking for the Windows case?

mikaelarguedas: sorry haven't updated the comments yet. This comment is actually totally wrong. Windows files are always readable if they are writable

mikaelarguedas:
@wjwwood Sorry I had these commits lying around but didn't have time to push them because not toally ready. This is still in progress though and I'll change the label / tag the team when it's ready for review

src/file_permissions.c
wjwwood: I see, makes sense.

wjwwood:
@mikaelarguedas I'm not in a hurry, but @Karsten1987 just asked if this supersedes https://github.com/ros2/rmw/pull/87, which I think it should, but first it needs to be fixed up. That's why I went ahead and did a review.

test/test_filesystem.cpp
```diff
@@ -18,60 +18,79 @@
...
 
 static char cwd[1024];
+#ifdef WIN32
+const char delimiter = '\\';
+#else
+const char delimiter = '/';
+#endif
 
 TEST(test_filesystem, utilities_exists) {
   utilities_get_cwd(cwd, 1024);
```
dirk-thomas: The return value of the function call should be checked.

I would propose to even make that a "requirement" (using `__attribute__((warn_unused_result))` / `_Check_return_` on the function declaration).

Same below as well as for other functions with a return value.

mikaelarguedas: Good point, added in https://github.com/ros2/c_utilities/pull/2/commits/87873ccd7555e93e397b04ce2522b529c63c1d57

dirk-thomas: I guess only `utilities_get_cwd` has arguments which are used to output information from the functions. For the other functions the return value is the only output. I don't think we need the a attribute for those. Sorry for being unclear in the previous comment.

src/get_env.c
```diff
@@ -0,0 +1,60 @@
...
+  if (!env_name) {
+    return "argument env_name is null";
+  }
+  if (!env_value) {
+    return "argument env_value is null";
+  }
+  *env_value = NULL;
+#ifdef WIN32
+  size_t required_size;
+  errno_t ret = getenv_s(&required_size, __env_buffer, sizeof(__env_buffer), env_name);
```
dirk-thomas: Does `getenv_s` populate `__env_buffer` with an empty string in case the environment variable doesn't exist?

test/test_filesystem.cpp
mikaelarguedas: no worries I just did copy paste without any brain cell involved in the process :S https://github.com/ros2/c_utilities/pull/2/commits/87873ccd7555e93e397b04ce2522b529c63c1d57

src/get_env.c
wjwwood: Yes, that's how it is different from unix, because on Windows there's no difference between "not set" and "set to empty string". I think this is covered by this test:

https://github.com/ros2/c_utilities/pull/2/files#diff-7ca3b7c1d3db145dbb22cc9364c97cdaR39

mikaelarguedas: Based on the fact that this test passed: https://github.com/ros2/c_utilities/blob/8255f35c36f5ab7c73005839e05b3f319c49867a/test/test_get_env.cpp#L37-L39

I saw that in some places we use [dupenv_s](https://msdn.microsoft.com/en-us/library/ms175774.aspx). Is it for thread safety of just to not have to handle memory  allocation ? I don't mind switching `getenv_s` to `dupenv_s` if there is a strong preference

wjwwood: We could also provide a `utilities_dup_env()`. I was being conservative and trying to avoid memory allocation if possible. In cases where you know that the variable you're reading shouldn't be changing, like `ROS_DOMAIN_ID` I think you can safely use `utilities_get_env()`, but in other cases you might want the copy. On the other hand, `dupenv_s` is just for convenience, however, if does seem to behave differently from `get_env`:

> If the variable is not found, then buffer is set to NULL, numberOfElements is set to 0, and the return value is 0 because this situation is not considered to be an error condition.

-- https://msdn.microsoft.com/en-us/library/ms175774.aspx#Remarks

The `getenv_s` documentation is frustratingly unclear about this point, but I feel confident that I landed on the above behavior of `getenv_s` through experimentation and ultimately that test I pointed out.

mikaelarguedas: Ok, I'll stick with `get_env_s` here for now and will add dupenv_s later on if it proves necessary when refactoring.
Placing this in review then

mikaelarguedas:
Note: while tested locally on Linux, there is no automated test for non readable files. To test this, the test suite should create them, set the right permissions, run the test and clean them up after the fact

mikaelarguedas:
CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2380/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1837/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2496/)

include/c_utilities/filesystem.h
```diff
@@ -0,0 +1,97 @@
...
+/// Check if the provided path points to a directory.
+/**
+ * \param[in] abs_path Absolute path to check.
+ * \return bool True if directory False otherwise
+ */
+C_UTILITIES_PUBLIC
+bool
+utilities_is_directory(const char * abs_path);
+
+/// Check if the provided path points to a directory.
```
dirk-thomas: Copy-n-paste.

include/c_utilities/get_env.h
```diff
@@ -0,0 +1,52 @@
...
+
+/// Retrieve the value of the given environment variable if it exists, or "".
+/* The returned c-string is only valid until the next time this function is
+ * called, because it is a direct pointer to the static storage.
+ * The variable env_value populated by this function should never have free() called on it.
+ * If the environment variable is not set, an empty string will be returned.
+ *
+ * Environment variables will be truncated at 2048 characters on Windows.
+ *
+ * This function may allocate heap memory.
```
dirk-thomas: That doesn't seem to be the case?

include/c_utilities/get_env.h
```diff
@@ -0,0 +1,52 @@
...
+ * This function is not thread-safe.
+ * This function is not lock-free.
```
dirk-thomas: Does it actually use a lock?

mikaelarguedas: Good point, all were copy-paste from https://github.com/ros2/rmw/pull/87 that I didn't update. Thanks for pointing it out !

include/c_utilities/filesystem.h
```diff
@@ -0,0 +1,107 @@
...
+/// Return current working directory.
+/**
+ * \param[in] buffer Allocated string to store current directory path to
+ * \param[in] max_length maximum length to be stored in buffer
+ * \return bool True if success False otherwise
+ */
+C_UTILITIES_PUBLIC
+C_UTILITIES_WARN_UNUSED
+bool
+utilities_get_cwd(char * buffer, size_t max_length);
```
wjwwood: For all these functions which take in a `char *` of some type or another. What happens if I pass `NULL`? I think it's worth just putting a in each docblock that says `False` is returned in that case, since that's the easiest solution.

A more extreme, but precise, solution would be to return a return code rather than a bool, so that the user could know if it failed because of the check or because of some other logistical issue, like a null pointer being passed.

include/c_utilities/filesystem.h
```diff
@@ -0,0 +1,107 @@
...
+utilities_exists(const char * abs_path);
+
+/// Check if the provided path points to a file/folder readable by current user.
+/**
+ * \param[in] abs_path Absolute path to check.
+ * \return bool True if the file is readable False otherwise
+ */
+C_UTILITIES_PUBLIC
+bool
+utilities_is_readable(const char * file_abs_path);
```
wjwwood: Several of these functions seem to use `file_abs_path` while their documentation says `abs_path`. I don't know which is right at this point, but I would have guessed the more generic `abs_path` is the intended name, given the fact that most of the functions operate on both files and folders.

include/c_utilities/filesystem.h
```diff
@@ -0,0 +1,107 @@
...
+
+/// Concatenate path adding the right delimiter according to the platform.
+/**
+ * \param[in] lhp, left hand path
+ * \param[in] rhp, right hand path
+ * \return cons char * concatenated path
+ */
+C_UTILITIES_PUBLIC
+const char *
+utilities_join_path(const char * lhp, const char * rhp);
```
wjwwood: nitpick: use the full names `left_hand_path` and `right_hand_path`. It's a one time cost of writing these out, but might be the difference in the user needing to lookup the docs versus just understanding the signature. It doesn't place a typing burden on user.

src/filesystem.c
```diff
@@ -0,0 +1,162 @@
...
+  }
+#ifdef WIN32
+  return (buf.st_mode & S_IFREG) == S_IFREG;
+#else
+  return S_ISREG(buf.st_mode);
+#endif
+}
+
+bool
+utilities_exists(const char * file_abs_path)
```
wjwwood: Also, might ought to be changed to `abs_path`.

include/c_utilities/filesystem.h
mikaelarguedas: good point. For every other function, the check on `stat` will cause a `return false` so I think only this function needs to be modified.

I didn't see the need to go all the way to create error codes for these because the scope of these function is pretty limited. But if you think otherwise I can add some

include/c_utilities/filesystem.h
mikaelarguedas: sounds good, that was a shameless copy paste of concat

include/c_utilities/filesystem.h
wjwwood: No, I think documentation should be sufficient.

include/c_utilities/get_env.h
```diff
@@ -24,11 +24,29 @@ extern "C"
...
+ * #include <c_utilities/get_env.h>
+ * const char * env_value;
+ * const char * error_str;
+ * error_str = utilities_get_env("SOME_ENV_VAR", &env_value);
+ * if (error_str != NULL) {
+ *   fprintf(stderr, "Error getting env var: %s\n", error_str);
+ * }
+ * printf("Valued of 'SOME_ENV_VAR': %s\n", env_value);
+ * ```
+ *
```
mikaelarguedas: thanks @wjwwood for improving the doc!

wjwwood: Yeah, np. Please read over it, I'm prone to typos.

include/c_utilities/filesystem.h
mikaelarguedas: sounds good https://github.com/ros2/c_utilities/pull/2/commits/91728406de92b0cff0d6151b8d2880d356e289cb

wjwwood:
I dunno if it needs CI again or not, up to you.

mikaelarguedas:
Can't hurt:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2383/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1840/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2504/)

Thanks for reviewing

PR #8, opened 2017-04-02
c89 compliance: move declaration outside for loop

codebot:
to resolve discussion in https://github.com/ros2/c_utilities/issues/7

### DIFF
--- a/src/cmdline_parser.c
+++ b/src/cmdline_parser.c
@@ -19,7 +19,8 @@
 bool cli_option_exist(char ** begin, char ** end, const char * option)
 {
   // return std::find(begin, end, option) != end;
-  for (size_t i = 0; i < (size_t)(end - begin); ++i) {
+  size_t i;
+  for (i = 0; i < (size_t)(end - begin); ++i) {
     if (strcmp(begin[i], option) == 0) {
       return true;
     }


dirk-thomas:
I don't see the reason for this change. The [previous code](https://github.com/ros2/c_utilities/commit/f9ad68640780c001ae082eedb0fe73eb3f8d8a09#diff-8c323f88b7e0fb9896b393c22967ae2c) compiles perfectly fine [on Windows](http://ci.ros2.org/job/ci_windows/2498/). And the compiler as well as the warning should be selected for the non-test code anyway (not only for the tests). So I would propose to revert his change and address [this comment](https://github.com/ros2/c_utilities/issues/7#issuecomment-291198285) instead.

dirk-thomas:
Yes, the standard version as well as the warning flags should be set outside of the test block (as it is done in all other ROS 2 packages).

wjwwood:
We don't set C11 because windows doesn't fully support it. Neither rmw nor rcl set it. They set c++11 for their cpp test files only. Personally, I think this is the right change.

dirk-thomas:
Windows supports the use case (to declare the variable within the loop) perfectly well. If in the future a new C11 specific is being used we and it turns out it doesn't work on Windows that simply can't be used. I don't see a reason why we should not utilize the features which are available all targeted platforms?

wjwwood:
The issue is not whether or not windows support is this. It is Linux, and I don't think we should turn on c11 on Linux just for this feature because that risks us using a feature in c11 that windows doesn't support. Is it possible you could just set c99?

I don't see the issue, what Morgan proposed here works everywhere.

dirk-thomas:
I am not proposing anything new here. Please see the [current CMake file](https://github.com/ros2/c_utilities/blob/c49b7da2e04083ac764d9160ec75d28218f53529/CMakeLists.txt#L35): it already uses C11. The flags just need to be set globally instead of in the test block. Simply because enabling / disabling tests should neither affect the C standard we use to build the code nor affect the warning level. I created PR #9 for this.

mikaelarguedas:
Agreed that the C flags (if set) should be set everywhere and not only when building tests.
Regarding the variable declaration within loop, glad to see that windows supports it now, it was definitely not the case before. Which is why all of the C code doesn't use it. Happy to start using it if it's now supported everywhere.
Regarding using the flag itself I'm happy to use it if the features we use from it are supported by MSVC

Karsten1987:
Sorry for the trouble here. That was definitely a mistake from my side. I should have set the C flags globally and only C++ in the tests. Must have been wrongly copy&pasted. 

Can we officially support C99 on all platforms? I think MSVC doesn't support all of it but has adopted a few things: https://msdn.microsoft.com/en-us/library/hh409293.aspx

```
99 Conformance Visual Studio 2015 fully implements the C99 Standard Library, with the exception of any library features that depend on compiler features not yet supported by the Visual C++ compiler (for example, <tgmath.h> is not implemented).
```

Can we instead of setting `c11` set `c99` ?

dirk-thomas:
> Can we instead of setting `c11` set `c99`?

Why should be restrict the features we can use as long as they are available all targeted platforms?

wjwwood:
> Why should be restrict the features we can use as long as they are available all targeted platforms?

I don't want to be cross-checking every feature of C11 with whether or not Windows supports it. Both `rmw` and `rcl` avoid setting C11 on purpose, I would have advised this package do the same had I noticed that it was using C11. The difference in C99 and C11 on Windows is that C99 is partially supported, but C11 is completely unsupported and they have hinted that they will never update their compiler to work with newer C standards.

Karsten1987:
I am not saying we should restrict ourselves, but we could have a statement for a minimum requirement, such that we require a C99 compatible compiler.

PR #9, opened 2017-04-03
set compiler flags globally

dirk-thomas:
Address the following comments:
* https://github.com/ros2/c_utilities/issues/7#issuecomment-291198285
* https://github.com/ros2/c_utilities/pull/8#issuecomment-291198373

CI builds:
* [Build Status](http://ci.ros2.org/job/ci_linux/2382/)
* [Build Status](http://ci.ros2.org/job/ci_osx/1839/) the compiler warning has been addressed in the second commit
* [Build Status](http://ci.ros2.org/job/ci_windows/2503/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ find_package(ament_cmake REQUIRED)
 
 include_directories(include)
 
+if(NOT WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
 set(utilities_sources
   src/cmdline_parser.c
   src/concat.c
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Could this start leveraging the `CMAKE_C_STANDARD` feature ? If we start supporting more compilers for various platforms it can be good to let cmake figure out the flags for us rather than calling out a specific flag explicitly. c.f.https://github.com/ros2/c_utilities/commit/db24473a0d67bc6f063bb6cca3c2524fbf159829

dirk-thomas:
@mikaelarguedas That sounds good. I would suggest doing that in a separate PR though.

mikaelarguedas:
sounds good. I'll update the `compile_options` branch here to reflect the changes of this PR. Hopefully I'll have time to test this soon on all repos. :+1: for merging as is. The linux job failed because of the time drift on the slaves but all test passed.

PR #10, opened 2017-04-05
cleaner way of (de-)init string_array + doc + tests

Karsten1987:


include/c_utilities/types/string_array.h
```diff
@@ -31,14 +31,38 @@ typedef struct C_UTILITIES_PUBLIC_TYPE string_array_t
   char ** data;
 } string_array_t;
 
+/// Returns an empty string array struct
+/*
+ * This function returns an empty and zero initialized string array struct.
+ * Calling utilities_string_array_fini on this instance leads to undefined
+ * behavior.
```
wjwwood: As I understood it, this should not be the case. Instead this is what I expected:

```c
string_array_t foo;
utilities_string_array_fini(foo);  // undefined behavior

string_array_t bar = utilities_get_zero_initialized_string_array();
utilities_string_array_fini(bar);  // ok

string_array_t baz = utilities_get_pre_initialized_string_array(10);
utilities_string_array_fini(baz);  // also, ok
```

src/string_array.c
```diff
@@ -35,6 +35,9 @@ utilities_get_pre_initialized_string_array(size_t size)
   static string_array_t array = {0, NULL};
   array.size = size;
   array.data = (char **)malloc(array.size * sizeof(char *));
+  for (size_t i = 0; i < size; ++i) {
+    array.data[i] = NULL;
+  }
```
wjwwood: I think using `calloc` would be easier/better than iterating over it again.

src/string_array.c
```diff
@@ -47,19 +50,13 @@ utilities_string_array_fini(string_array_t * array)
...
-      free(array->data[i]);
-      array->data[i] = NULL;
-    } else {
-      ret = UTILITIES_RET_WARN;
-    }
+    free(array->data[i]);
+    array->data[i] = NULL;
   }
 
   if (array->data) {
```
wjwwood: This should be checked before you loop over `array->data` in the previous for loop.

include/c_utilities/types/string_array.h
wjwwood: The tests seem to confirm my expectation, maybe there is just a typo in the docs here.

CMakeLists.txt
```diff
@@ -77,6 +77,16 @@ if(BUILD_TESTING)
     target_link_libraries(test_concat c_utilities)
   endif()
 
+  ament_add_gtest(test_string_array
+    test/test_string_array.cpp
+  )
+  if(TARGET test_string_array)
+    target_include_directories(test_string_array PUBLIC
+      ${c_utilities_INCLUDE_DIRS}
```
dirk-thomas: Can you please double check the content of this variable. I doubt that it contains anything and also since `include_directories` is called before shouldn't be necessary.

src/string_array.c
Karsten1987: can you quickly double check that the new changes correspond to what you had in mind?

Karsten1987:
thanks for the review! Would you lovely people take another quick minute to re-review?

dirk-thomas:
These [two compiler warnings](http://ci.ros2.org/job/ci_windows/2522/warnings41Result/) on Windows seem to come from this PR.

Karsten1987:
I'll take care of it asap.

PR #12, opened 2017-04-06
use _strdup() on windows to avoid compiler warning

gerkey:
Aiming to squash these warnings:

http://ci.ros2.org/job/ci_windows/2524/warnings41Result/

Following guidance from here:

http://stackoverflow.com/questions/7582394/strdup-or-strdup#7582741

### DIFF
--- a/test/test_string_array.cpp
+++ b/test/test_string_array.cpp
@@ -16,6 +16,10 @@
 
 #include "c_utilities/types/string_array.h"
 
+#ifdef _WIN32
+  #define strdup _strdup
+#endif
+
 TEST(test_string_array, boot_string_array) {
   // UNDEFIEND BEHAVIOR
   // string_array_t sa00;


gerkey:
* http://ci.ros2.org/job/ci_linux/2407/
* http://ci.ros2.org/job/ci_linux-aarch64/43/
* http://ci.ros2.org/job/ci_osx/1856/
* http://ci.ros2.org/job/ci_windows/2526/

Karsten1987:
thanks for catching up on that. We can merge this once CI passed.

However, I am not so sure if this is the appropriate long-term solution or rather put that in a utility function. I've seen that fastrtps does similar things and I don't really see the point in copying these defines all over our source files.

wjwwood:
The linux-aarch64 and macOS test failures look unrelated. Since this happened on my watch as Build cop, and since it's still affecting CI jobs, I'm going to ahead and merge this.

wjwwood:
Thanks @gerkey for fixing this so quickly.

gerkey:
@Karsten1987 I agree with your concern about replicating defines like this everywhere. In this case, I'm not expecting widespread use of `strdup` in our code base, so it doesn't worry me that much. But if we do start to see these things pile up, then we should collect them somewhere.

mikaelarguedas:
agreed with @Karsten1987 and @gerkey , it would be sensible to have it in `c_utilities` for reuse.
I'm actually surprised that this warning didn't show up before  [e.g. here](https://github.com/ros2/rmw_connext/blob/2f7630d17deee3cdc3e7e16d8c538d62afff832e/rmw_connext_shared_cpp/src/shared_functions.cpp#L322)
Give that we already define it in some places of our codebase ([in the middle of a file](https://github.com/ros2/rmw_fastrtps/blob/d353b3ce34849147b0bddcc184810b04cebfb736/rmw_fastrtps_cpp/src/functions.cpp#L2327) which is even weirder) +1 for having it in this repo and reuse it where needed

PR #15, opened 2017-04-12
just some small doc fixups

wjwwood:
I just ran across these while adding some new stuff to this package.


Karsten1987:
great. Thanks for reviewing :)

PR #16, opened 2017-04-12
Allocator and error handling in c utilities

### LINKED ISSUE #6: pass custom allocator struct to utilities function
all functions should be somehow capable of handling custom memory allocators in the style of RCL.
https://github.com/ros2/rcl/blob/master/rcl/include/rcl/allocator.h

wjwwood:
This should help with ros2/rmw#94 and #6.

Pull requests to remove the duplicate functionality in `rmw` and `rcl` will follow, but the pr for gmock (https://github.com/ament/ament_cmake/pull/91) can be merged at anytime.

I tried to take care to have a "move" commit and then a "change" commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,9 @@ include_directories(include)
 
 if(NOT WIN32)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  # The warning flags are added per target, because otherwise the local
+  # gmock library that is built will have compiler warnings.
+  set(c_utilities_extra_warning_flags "-Wall -Wextra -Wpedantic")
 endif()
 
 set(utilities_sources
@@ -29,6 +31,7 @@ add_library(
   c_utilities
   SHARED
   ${utilities_sources})
+set_target_properties(c_utilities PROPERTIES COMPILE_FLAGS ${c_utilities_extra_warning_flags})
 
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
@@ -66,8 +69,22 @@ if(BUILD_TESTING)
     set(SKIP_TEST_IF_WIN32 "SKIP_TEST")
   endif()
 
+  macro(c_utilities_custom_add_gtest target)
+    ament_add_gtest(${target} ${ARGN})
+    if(TARGET ${target})
+      set_target_properties(${target} PROPERTIES COMPILE_FLAGS ${c_utilities_extra_warning_flags})
+    endif()
+  endmacro()
+
+  macro(c_utilities_custom_add_gmock target)
+    ament_add_gmock(${target} ${ARGN})
+    if(TARGET ${target})
+      set_target_properties(${target} PROPERTIES COMPILE_FLAGS ${c_utilities_extra_warning_flags})
+    endif()
+  endmacro()
+
   # Gtests
-  ament_add_gtest(test_allocator test/test_allocator.cpp
+  c_utilities_custom_add_gtest(test_allocator test/test_allocator.cpp
     ENV ${extra_test_env}
     APPEND_LIBRARY_DIRS ${extra_lib_dirs}
     ${SKIP_TEST_IF_WIN32}
@@ -76,7 +93,7 @@ if(BUILD_TESTING)
     target_link_libraries(test_allocator c_utilities ${extra_test_libraries})
   endif()
 
-  ament_add_gmock(test_error_handling test/test_error_handling.cpp
+  c_utilities_custom_add_gmock(test_error_handling test/test_error_handling.cpp
     # Append the directory of libc_utilities so it is found at test time.
     APPEND_LIBRARY_DIRS "$<TARGET_FILE_DIR:c_utilities>"
   )
@@ -84,35 +101,35 @@ if(BUILD_TESTING)
     target_link_libraries(test_error_handling ${PROJECT_NAME})
   endif()
 
-  ament_add_gtest(test_split
+  c_utilities_custom_add_gtest(test_split
     test/test_split.cpp
   )
   if(TARGET test_split)
     target_link_libraries(test_split c_utilities)
   endif()
 
-  ament_add_gtest(test_find
+  c_utilities_custom_add_gtest(test_find
     test/test_find.cpp
   )
   if(TARGET test_find)
     target_link_libraries(test_find c_utilities)
   endif()
 
-  ament_add_gtest(test_concat
+  c_utilities_custom_add_gtest(test_concat
     test/test_concat.cpp
   )
   if(TARGET test_concat)
     target_link_libraries(test_concat c_utilities)
   endif()
 
-  ament_add_gtest(test_string_array
+  c_utilities_custom_add_gtest(test_string_array
     test/test_string_array.cpp
   )
   if(TARGET test_string_array)
     target_link_libraries(test_string_array c_utilities)
   endif()
 
-  ament_add_gtest(test_get_env test/test_get_env.cpp
+  c_utilities_custom_add_gtest(test_get_env test/test_get_env.cpp
     ENV
       EMPTY_TEST=
       NORMAL_TEST=foo
@@ -120,15 +137,9 @@ if(BUILD_TESTING)
   )
   if(TARGET test_get_env)
     target_link_libraries(test_get_env ${PROJECT_NAME})
-    if(UNIX AND NOT APPLE)
-      target_link_libraries(test_get_env pthread)
-    endif()
-    if(NOT WIN32)
-      set_target_properties(test_get_env PROPERTIES COMPILE_FLAGS "-std=c++11")
-    endif()
   endif()
 
-  ament_add_gtest(test_filesystem
+  c_utilities_custom_add_gtest(test_filesystem
     test/test_filesystem.cpp
   )
   if(TARGET test_filesystem)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ if(BUILD_TESTING)
   endif()
 
   find_package(ament_cmake_gmock REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
@@ -66,7 +67,7 @@ if(BUILD_TESTING)
   endif()
 
   # Gtests
-  ament_add_gmock(test_allocator test/test_allocator.cpp
+  ament_add_gtest(test_allocator test/test_allocator.cpp
     ENV ${extra_test_env}
     APPEND_LIBRARY_DIRS ${extra_lib_dirs}
     ${SKIP_TEST_IF_WIN32}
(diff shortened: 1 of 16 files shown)


wjwwood:
I still haven't figure out what to do with the memory tools that have been copied here. They are also used in `rcl` even after removing the allocator stuff later. I don't really want to create a new package for it, so I might expose the source files from this repository via a cmake macro, since they're useful for testing C or C++ in other packages.

CMakeLists.txt
```diff
@@ -42,40 +44,69 @@ if(BUILD_TESTING)
...
+
+  ament_add_gmock(test_split
     test/test_split.cpp
   )
   if(TARGET test_split)
     target_link_libraries(test_split c_utilities)
   endif()
 
-  ament_add_gtest(test_find
+  ament_add_gmock(test_find
```
dirk-thomas: Why are the existing unmodified gtests changed into gmocks?

wjwwood: I thought I should only use gmock or gtest, but not both. I can change them back if that's not the case.

wjwwood: I never heard back from you on this, so I just did it anyways as not to delay this further: a642c12

wjwwood:
With respect to how to de-duplicate the `memory_tools` here and in `rcl`, I will address this in a follow on pull request after we do the renaming (https://github.com/ros2/c_utilities/issues/17). For two reasons, first I want to merge these pr's before doing the renaming and de-duplicating `memory_tools` is going to take a lot of time, and second because I want to refactor memory tools even more to be namespaced under this package's name and I'd rather not do it twice.

wjwwood:
CI (fixed cmake linter error in 5a422fb):

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2469/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/82/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1923/) (known compiler warning which should be fixed in latest version of Fast-RTPS)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2587/)

I'm running new CI with some fixes, but this (and all related prs) can go in review now.

wjwwood:
Updated CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2470/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/83/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1926/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2590/) (known flaky test)

src/error_handling.c
```diff
@@ -0,0 +1,259 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <c_utilities/error_handling.h>
+#include <c_utilities/macros.h>
+
+#ifndef UTILITIES_REPORT_ERROR_HANDLING_ERRORS
+#define UTILITIES_REPORT_ERROR_HANDLING_ERRORS 1
```
mikaelarguedas: A comment here explaining what this is used for would be useful)

src/error_handling.c
```diff
@@ -0,0 +1,259 @@
...
+#endif
+  }
+#endif
+  __utilities_error_state->file = file;
+  __utilities_error_state->line_number = line_number;
+  if (__utilities_error_is_set(old_error_state)) {
+#if UTILITIES_REPORT_ERROR_HANDLING_ERRORS
+    // Only warn of overwritting if the new error string is different from the old ones.
+    if (error_string != old_error_string && error_string != old_error_state->message) {
+      fprintf(
```
mikaelarguedas: should this use SAFE_FWRITE_TO_STDERR like the other error messages?

src/error_handling.c
```diff
@@ -0,0 +1,259 @@
...
+  }
+  return utilities_get_error_string();
+}
+
+void
+__utilities_reset_error_string(char ** error_string_ptr, utilities_allocator_t allocator)
+{
+  char * error_string = *error_string_ptr;
+  if (error_string) {
+    if (error_string) {
```
mikaelarguedas: maybe you meant:
```c++
if(error_string_ptr) {
  char * error_string = *error_string_ptr;
  if(error_string) {
```

src/error_handling.c
```diff
@@ -0,0 +1,259 @@
...
+  *error_string_ptr = NULL;
+}
+
+void
+__utilities_reset_error(utilities_error_state_t ** error_state_ptr_ptr)
+{
+  if (error_state_ptr_ptr) {
+    utilities_error_state_t * error_state_ptr = *error_state_ptr_ptr;
+    if (error_state_ptr) {
+      utilities_allocator_t allocator = error_state_ptr->allocator;
```
mikaelarguedas: what happens if allocator is null? wouldnt deallocate not be defined?

test/memory_tools/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+if(APPLE)
+  # Create an OS X specific version of the memory tools that does interposing.
+  # See: http://toves.freeshell.org/interpose/
+  add_library(${PROJECT_NAME}_memory_tools_interpose SHARED memory_tools_osx_interpose.cpp)
+  target_link_libraries(${PROJECT_NAME}_memory_tools_interpose ${GMOCK_LIBRARIES})
+  list(APPEND extra_test_libraries ${PROJECT_NAME}_memory_tools_interpose)
+  list(APPEND extra_test_env
+    DYLD_INSERT_LIBRARIES=$<TARGET_FILE:${PROJECT_NAME}_memory_tools_interpose>)
+endif()
+if(UNIX AND NOT APPLE)
```
mikaelarguedas: Nit: elseif(UNIX)

test/memory_tools/test_memory_tools.cpp
```diff
@@ -0,0 +1,133 @@
...
+  });
+  set_on_unexpected_free_callback(on_unexpected_free);
+  void * mem = nullptr;
+  void * remem = nullptr;
+  // First try before enabling, should have no effect.
+  mem = malloc(1024);
+  ASSERT_NE(mem, nullptr);
+  remem = realloc(mem, 2048);
+  ASSERT_NE(remem, nullptr);
+  if (!remem) {free(mem);}
```
mikaelarguedas: Question: redundant condition? previous assert would have raised if this condition was not true, same below

test/memory_tools/test_memory_tools.cpp
```diff
@@ -0,0 +1,133 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "./memory_tools.hpp"
+
+/* Tests the allocatation checking tools.
```
mikaelarguedas: typo allocatation => allocation

test/test_error_handling.cpp
```diff
@@ -0,0 +1,89 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+
+#include "gmock/gmock.h"
```
mikaelarguedas: nit: should thi be included as system header?

include/c_utilities/error_handling.h
```diff
@@ -0,0 +1,133 @@
...
+ * The returned pointer is valid until UTILITIES_SET_ERROR_MSG,
+ * utilities_set_error_state, or utilities_reset_error are called in the same thread.
+ *
+ * \return The current error string, with file and line number, or "error not set" if not set.
+ */
+C_UTILITIES_PUBLIC
+const char *
+utilities_get_error_string_safe(void);
+
+/// Resets the error state by clearing any previously set error state.
```
mikaelarguedas: Typo/Consistency : Resets => Reset

src/error_handling.c
mikaelarguedas: or simply fwrite like above to be consistent

wjwwood: This error message uses formatting (like `%s`), fwrite cannot do that. I only fallback to fwrite when memory allocation fails, because apparently it can still work on some systems even when malloc fails.

src/error_handling.c
wjwwood: I think so, I guess it was wrong in `rmw` as well.

src/error_handling.c
wjwwood: `error_state_ptr` may be null, but `error_state_ptr->allocator` cannot be null. `error_state_ptr->allocator->deallocate` can be null though. The assumption is (and I think it holds) that if `error_state_ptr` is not null, then the allocator will be valid as well. I can make it more defensive though.

test/test_error_handling.cpp
wjwwood: Some of the other files are combining it with the system headers, but it's not a system header imo, i.e. it's not part of the standard library. So I don't think so, but I can I guess.

test/memory_tools/test_memory_tools.cpp
wjwwood: b0c9939

include/c_utilities/error_handling.h
wjwwood: aa607bf

src/error_handling.c
wjwwood: cbf5eeb

src/error_handling.c
wjwwood: cbf5eeb

test/memory_tools/test_memory_tools.cpp
wjwwood: d5035f5

test/memory_tools/CMakeLists.txt
wjwwood: 7620056

src/error_handling.c
wjwwood: ae4100a

wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2485/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/87/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1939/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2596/)

I'm not sure about the issues on macOS, because I cannot reproduce them locally. I wasn't getting those errors until very recently. I'll look into it before merging.

wjwwood:
Looks like something is amiss with mini1, dosa reports green for macOS:

[Build Status](http://ci.ros2.org/job/ci_osx/1941/)

I'll merge now.

test/memory_tools/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+    DYLD_INSERT_LIBRARIES=$<TARGET_FILE:${PROJECT_NAME}_memory_tools_interpose>)
+elseif(UNIX)
+  # On Linux like systems, add dl and use the normal library and DL_PRELOAD.
+  list(APPEND extra_test_libraries dl)
+  list(APPEND extra_test_env DL_PRELOAD=$<TARGET_FILE:${PROJECT_NAME}_memory_tools>)
+endif()
+list(APPEND extra_lib_dirs $<TARGET_FILE_DIR:${PROJECT_NAME}_memory_tools>)
+target_link_libraries(${PROJECT_NAME}_memory_tools ${extra_test_libraries})
+target_compile_definitions(${PROJECT_NAME}_memory_tools
+  PRIVATE "RCL_MEMORY_TOOLS_BUILDING_DLL")
```
dirk-thomas: This `RCL` prefix hasn't been changed to `RCUTILS`.

The same for all the prefixes in the `memory_tools.hpp`.

wjwwood: Thanks for catching that, see: https://github.com/ros2/rcutils/pull/24

PR #20, opened 2017-04-25
Namespace expansion

wjwwood:
This pull request is in support of implementing namespace expansion in rcl. It adds two string utility functions (`rcutils_strdup` and `rcutils_format_string`). It also adds a new data structure which is a map where both the keys and values are c strings.

This is still a work in progress along with the changes in rcl that are needed to expand topic strings.

Connects to ros2/rcl#132

### DIFF
--- a/include/rcutils/types/string_map.h
+++ b/include/rcutils/types/string_map.h
@@ -1,0 +1,308 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__TYPES__STRING_MAP_H_
+#define RCUTILS__TYPES__STRING_MAP_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include <string.h>
+
+#include "rcutils/allocator.h"
+#include "rcutils/types/rcutils_ret.h"
+#include "rcutils/macros.h"
+#include "rcutils/visibility_control.h"
+
+struct rcutils_string_map_impl_t;
+
+typedef struct RCUTILS_PUBLIC_TYPE rcutils_string_map_t
+{
+  struct rcutils_string_map_impl_t * impl;
+} rcutils_string_map_t;
+
+/// Return an empty string map struct.
+/*
+ * This function returns an empty and zero initialized string map struct.
+ *
+ * Example:
+ *
+ * ```c
+ * // Do not do this:
+ * // rcutils_string_map_t foo;
+ * // rcutils_string_map_fini(&foo); // undefined behavior!
+ *
+ * // Do this instead:
+ * rcutils_string_map_t bar = rcutils_get_zero_initialized_string_array();
+ * rcutils_string_array_fini(&bar); // ok
+ * ```
+ * */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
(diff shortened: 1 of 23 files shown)


wjwwood:
I've now moved `isalnum_no_locale()` here as well as added a new string function `rcutils_repl_str()` based on this Public Domain code:

http://creativeandcritical.net/str-replace-c

If anyone has a chance to see how I've placed it in our code and give feedback, I'd appreciate it.

src/string_map.c
```diff
@@ -111,8 +111,8 @@ rcutils_string_map_get_capacity(const rcutils_string_map_t * string_map, size_t
     string_map->impl, "invalid string map",
     return RCUTILS_RET_STRING_MAP_INVALID, rcutils_get_default_allocator())
   RCUTILS_CHECK_ARGUMENT_FOR_NULL(
-    capacity, RCUTILS_RET_INVALID_ARGUMENT, rcutils_get_default_allocator())
-  *capacity = string_map->impl->capacity;
+    capacity, RCUTILS_RET_INVALID_ARGUMENT, rcutils_get_default_allocator()) *
+  capacity = string_map->impl->capacity;
```
dirk-thomas: This is really hard to read. Can you work around this by doing something like this instead?

```
{
  *capacity = string_map->impl->capacity;
}
```

or 

```
// comment to keep style checker sane
*capacity = string_map->impl->capacity;
```

wjwwood: Thanks I didn't notice that when reviewing what it did. Unfortunately the comment didn't help, it still though it was part of the previous statement and put a space between `*` and `capacity`. I also couldn't put a semicolon on the macro because that would trigger the unnecessary semicolon warning from the compiler.

So I just turned uncrustify off on that line: 38cbbca

include/rcutils/types/string_map.h
```diff
@@ -0,0 +1,401 @@
...
+rcutils_ret_t
+rcutils_string_map_set_no_resize(
+  rcutils_string_map_t * string_map,
+  const char * key,
+  const char * value);
+
+/// Unset a key value pair in the map.
+/**
+ * The key needs to be a null terminated c string.
+ * If the given key is not found, RCUTILS_RET_STRING_KEY_NOT_FOUND is returned.
```
Karsten1987: what's happening with the memory?

test/test_string_map.cpp
```diff
@@ -0,0 +1,1439 @@
...
+    EXPECT_EQ(RCUTILS_RET_STRING_MAP_ALREADY_INIT, ret) << rcutils_get_error_string_safe();
+    rcutils_reset_error();
+  }
+
+  // double init
+  {
+    rcutils_string_map_t string_map = rcutils_get_zero_initialized_string_map();
+    ret = rcutils_string_map_init(&string_map, 10, allocator);
+    EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string_safe();
+    ret = rcutils_string_map_init(&string_map, 10, allocator);
```
Karsten1987: what happens if you re-initialize the map with a diff capacity?

include/rcutils/types/string_map.h
wjwwood: Which memory?

test/test_string_map.cpp
wjwwood: You get `RCUTILS_RET_STRING_MAP_ALREADY_INIT` if you initialize a string map twice. Is that what you're talking about?

include/rcutils/types/string_map.h
Karsten1987: sorry, to be more precise. Maybe just mention that you deallocate memory when you call unset.

test/test_string_map.cpp
Karsten1987: i see. I just wondered whether the result is different when calling it twice but with different capacity values.

include/rcutils/types/string_map.h
wjwwood: I think this is covered by the set function which says that the key and value are copied and managed by the string map:

https://github.com/ros2/rcutils/pull/20/files#diff-d7eafeb8de6275c09bdb50b98d178534R236

test/test_string_map.cpp
wjwwood: It shouldn't be different, since no memory is allocated in either case.

PR #22, opened 2017-05-25
Use namespaces

wjwwood:
Connects to ros2/rcl#137

### DIFF
--- a/test/test_error_handling.cpp
+++ b/test/test_error_handling.cpp
@@ -87,3 +87,18 @@ TEST(test_error_handling, recursive) {
     "Expected ', at' in the error string twice but got it '" << count << "': " << err_msg;
   rcutils_reset_error();
 }
+
+TEST(test_error_handling, copy) {
+  rcutils_reset_error();
+  const char * test_message = "test message";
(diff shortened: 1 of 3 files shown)


test/test_error_handling.cpp
```diff
@@ -87,3 +87,18 @@ TEST(test_error_handling, recursive) {
...
+  const char * test_message = "test message";
+  RCUTILS_SET_ERROR_MSG(test_message, rcutils_get_default_allocator());
+  using ::testing::HasSubstr;
+  ASSERT_THAT(rcutils_get_error_string_safe(), HasSubstr(", at"));
+  rcutils_error_state_t copy;
+  rcutils_ret_t ret = rcutils_error_state_copy(rcutils_get_error_state(), &copy);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
+  ASSERT_STREQ(test_message, copy.message);
+  ASSERT_STREQ(__FILE__, copy.file);
+  rcutils_error_state_fini(&copy);
```
Karsten1987: maybe assert that the original message is still untouched after the copy state fini?

wjwwood: Check 1eb376b to see if that's what you're talking about.

PR #24, opened 2017-05-31
replace RCL prefixes with RCUTILS

wjwwood:
See: https://github.com/ros2/rcutils/pull/16#pullrequestreview-41341338


wjwwood:
Only running Linux CI because of the nature of the change:

[Build Status](http://ci.ros2.org/job/ci_linux/2652/)

PR #25, opened 2017-06-01
move time files from rcl

dirk-thomas:
Previous CI builds looked good - the latest set addresses some linter warnings:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2661/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/212/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2114/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2779/)

Please review since this is necessary to move forward on the logging API.


Karsten1987:
I don't feel really strong about it, but by just quickly browsing through the files, I have the feeling that this PR deserves a different place than rcutils. Does it make sense to create a ros-time related package? From what I've seen this PR is the complete implementation of time and thus goes beyond only utility functions. But I might be wrong.

mikaelarguedas:
I tend to agree with @Karsten1987 maybe the time utilities deserve a specific package within the rcutils repository. But same as previous comments I don't feel very strongly about it.

wjwwood:
I would also agree that putting the `*_ros_time_source_t` and related functions in here is not quite right. Ideally (IMO) these pr's would have:

- moved everything _but_ the ros time source stuff here (https://github.com/ros2/rcutils/pull/23#issuecomment-305320934)
- create aliases from the `rcutils_time*` types and functions to `rcl_time*` equivalents (removes the need for https://github.com/ros2/rclcpp/pull/332)
  - For me the fact that the generic "time source" types and functions live in `rcutils` is an implementation detail of `rcl`.
- add ros time stuff amongst the `rcl` aliases to `rcutils`, extending what's already there
  - this provides a "unified" interface for time for users of `rcl` (like `rclcpp` and `rclpy`)

But that's a lot of work, so maybe there is a compromise to be made, like putting the ros time source stuff back into `rcl` and just leaving the discrepancy in namespace. The other middle ground of moving everything to `rcutils` including the ros time source, is not as good of a compromise though I think.

> I tend to agree with @Karsten1987 maybe the time utilities deserve a specific package within the rcutils repository.

I don't see a need for another package just for time. I think that everything except the "ros time" time source could easily belong in `rcutils`. For those in favor of that, what does a package just for time get us?

Karsten1987:
My thinking behind this is that I believe that such a time package will
become bigger in the future (with sim time etc.) and having this a separate
package could allow ros2 agnostic packages such as TF2 use the same time
format than the ros2 stack without strongly depending on rcl*.

mikaelarguedas:
> I don't see a need for another package just for time. I think that everything except the "ros time" time source could easily belong in rcutils. For those in favor of that, what does a package just for time get us?

It was more going in the direction of having the "ros time stuff" dissociated from the common utilities that are purely ros agnostic (as you expressed it much more clearly than me in your comment) than wanting the time utilities to have their own package. The additional package seemed to be the least amount of work allowing to not mix them in the same package. but I'm in favor of whatever compromise allowing this to move forward

wjwwood:
There is also the other alternative I proposed (https://github.com/ros2/rcutils/pull/23#issuecomment-305322542), which I think might be the best of the alternatives, though maybe I didn't describe it properly.

Basically you would:

- change the logging macros that need time to just take a function pointer that returns the current time
  - this decouples them from the "time source" concept
  - a time source could still be wrapped in a function which is passed to the macro
  - convenience macros could avoid the caller from needing to pass a function in common cases, i.e.:
    - `LOG_THROTTLE(msg, period, get_now)` with convenience macros like `LOG_THROTTLE_SYSTEM_TIME(msg, period)` (implicit `get_now` == `get_system_time_now()`)
- move just the functions that get the current system or steady time from `rcl` (reuse them in `rcl` to implement time source concept)
  - for example:
    - https://github.com/ros2/rcl/blob/master/rcl/src/rcl/time_unix.c and
    - https://github.com/ros2/rcl/blob/master/rcl/src/rcl/time_win32.c
- leave the rest of the "time stuff" in `rcl` and use the time functions in rcutils to implement them, i.e. `uint64_t rcutils_get_system_time_now()` and `uint64_t rcutils_get_steady_time_now()`

I think this requires more nuance than just moving everything, but it also provides the cleanest break between the logging macros and the time source concept.

wjwwood:
> My thinking behind this is that I believe that such a time package will
become bigger in the future (with sim time etc.) and having this a separate
package could allow ros2 agnostic packages such as TF2 use the same time
format than the ros2 stack without strongly depending on rcl*.

I think that's an admirable goal, but I see it as a separate decision from how to let the logging macros live in `rcutils` without moving ROS specific concepts into `rcutils` from `rcl`.

Also, anything using sim time will need a node to initialize sim time. It's possible that the interface, behind which sim time is implemented, could be generic and not tied to `rcl*`, but then I'd say it could just live in `rcutils` as a "general purpose C time source abstraction interface".

dirk-thomas:
> change the logging macros that need time to just take a function pointer that returns the current time
> this decouples them from the "time source" concept
> a time source could still be wrapped in a function which is passed to the macro

The wrapping function can't be as simple as `double get_now()` imo. What does it do when the internal time source returns an error?

Anyway since we need a decision quickly on this topic I will schedule an in-person meeting for this today.

wjwwood:
> The wrapping function can't be as simple as double get_now() imo. What does it do when the internal time source returns an error?

We're sort of having the same conversation in two places, but as I said there you can have the `get_now()` function optionally return an error too and the logging macro can abort or exit or printf to stderr or whatever it wants if the get_now function fails.

dirk-thomas:
The conclusion of the ad hoc meeting was to only move the functionality to retrieve the current steady time from `rcl` to `rcutils` since it is required for the throttle macros. For consistency the same function for getting the system time is being moved (but not being used within `rcutils` at the moment). Any additional types and structures (e.g. time source, time source type, ROS time, etc.) are not being moved and stay in `rcl`.

I will update this as related PRs accordingly soon.

dirk-thomas:
New CI with updated PRs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2667/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/217/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2120/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2785/)

PR #26, opened 2017-06-06
Explicitly set C11 via the CMake variable

esteve:
The version of C standard wasn't explicit, so I added `CMAKE_C_STANDARD` and set it to 99 (if I'm not mistaken that's the version that ROS2 is targeting).

I also added some logic that enables switching to C11 to enable `_Thread_local` instead of `__thread`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,13 @@ find_package(ament_cmake REQUIRED)
 
 include_directories(include)
 
+# Default to C99
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 99)
+endif()
+
 if(NOT WIN32)
+  add_definitions(-D_POSIX_C_SOURCE=199309L)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
   # The warning flags are added per target, because otherwise the local
   # gmock library that is built will have compiler warnings.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,9 @@ find_package(ament_cmake REQUIRED)
 
 include_directories(include)
 
-# Default to C99
+# Default to C11
 if(NOT CMAKE_C_STANDARD)
-  set(CMAKE_C_STANDARD 99)
+  set(CMAKE_C_STANDARD 11)
 endif()
 
 if(NOT WIN32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,6 @@ if(NOT CMAKE_C_STANDARD)
 endif()
 
 if(NOT WIN32)
-  add_definitions(-D_POSIX_C_SOURCE=199309L)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
   # The warning flags are added per target, because otherwise the local
   # gmock library that is built will have compiler warnings.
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Thanks @esteve, we currently don't use the CMAKE_X_STANDARD but that would be a great improvement and more platform/compiler agnostic way to do in every package rather than passing `-std` flags everywhere

include/rcutils/macros.h
```diff
@@ -57,7 +57,12 @@ extern "C"
   #endif
 #else
 // Some other non-Windows, non-cygwin, non-apple OS
-  #define RCUTILS_THREAD_LOCAL _Thread_local
+  #if  __STDC_VERSION__ >= 201112L
+// C11 enabled
```
dirk-thomas: Please move this comment at the end of the previous line (or indent it to the level of the current block).

CMakeLists.txt
```diff
@@ -6,7 +6,13 @@ find_package(ament_cmake REQUIRED)
 
 include_directories(include)
 
+# Default to C99
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 99)
+endif()
+
 if(NOT WIN32)
+  add_definitions(-D_POSIX_C_SOURCE=199309L)
```
dirk-thomas: What is this needed for?

esteve: This was needed for `clock_gettime`, but only if `std=c99` was used instead. `CMAKE_C_STANDARD` already causes CMake to set the appropriate flags so `_POSIX_C_SOURCE` is not needed

include/rcutils/macros.h
esteve: How so? Doesn't this follow the same style as https://github.com/ros2/rcutils/blob/master/include/rcutils/macros.h#L32 https://github.com/ros2/rcutils/blob/master/include/rcutils/macros.h#L35 et al?

esteve:
I just rebased this to resolve any potential conflicts, let me know if there's any further work that needs to be done.

mikaelarguedas:
Thanks @esteve ! I'm trying out the same logic on every package and will run CI with all of them including your branches on rmw and rcl

esteve:
@mikaelarguedas no problem, great idea to use `CMAKE_C_STANDARD`, `CMAKE_CXX_STANDARD` and `add_compile_options()` in all the packages.

dirk-thomas:
Please see #36 for switching to use C11.

mikaelarguedas:
@esteve would you mind changing this one to C11 as well ? It proved to generate warnings on OSX because of the new variadic logging macros. thanks!

esteve:
@mikaelarguedas done, I've pushed https://github.com/ros2/rcutils/pull/26/commits/4224fa1cd1087846c812d62465fb296220ff77ac to switch to C11

include/rcutils/macros.h
```diff
@@ -57,7 +57,12 @@ extern "C"
...
+    #define RCUTILS_THREAD_LOCAL _Thread_local
+  #else
+    #define RCUTILS_THREAD_LOCAL __thread
+  #endif
```
dirk-thomas: Do we still need / want to keep this with C11 being selected? I would be in favor or removing the change and keeping it as-is.

mikaelarguedas: I think it's safe to remove this given that C11 will be selected by default from now on

dirk-thomas: The remaining patch is already part of #35. Should we close this than?

mikaelarguedas: that works, when do you expect to land #35 ?

dirk-thomas: Asap.

dirk-thomas:
@esteve Thank you for working on this. We well merge #35 soon so I will close this.

esteve:
@dirk-thomas no problem, happy to see this getting merged one way or another :-)

PR #28, opened 2017-06-06
increase test timeouts for OS X

### LINKED ISSUE #27: log throttle tests are failing on macOS
See: http://ci.ros2.org/job/nightly_osx_release/481/testReport/

These changes were merged yesterday: https://github.com/ros2/rcutils/pull/23

I don't have an example of these tests passing on the above link nightly (this is the first night), so I can't say if it is flakey or not. I assume so since I assume there was a passing CI job for macOS before merging. I don't see the same failure on Linux or Windows atm.

dirk-thomas:
Not that a single run says much... [Build Status](http://ci.ros2.org/job/ci_osx/2134/)

Fixes #27.

### DIFF
--- a/test/test_logging_macros.cpp
+++ b/test/test_logging_macros.cpp
@@ -130,9 +130,9 @@ TEST_F(TestLoggingMacros, test_logging_skipfirst) {
 
 TEST_F(TestLoggingMacros, test_logging_throttle) {
   for (int i : {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}) {
-    RCUTILS_LOG_ERROR_THROTTLE(RCUTILS_STEADY_TIME, 30 /* ms */, "throttled message %d", i)
+    RCUTILS_LOG_ERROR_THROTTLE(RCUTILS_STEADY_TIME, 50 /* ms */, "throttled message %d", i)
     using namespace std::chrono_literals;
-    std::this_thread::sleep_for(20ms);
+    std::this_thread::sleep_for(30ms);
   }
   EXPECT_EQ(g_log_calls, 5u);
   EXPECT_EQ(g_last_log_event.level, RCUTILS_LOG_SEVERITY_ERROR);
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
can we just run the same CI but with a retest-until-fail 50 rather than a restest until pass ? (just to have an idea of how flaky it is because as you said one run doesnt say much)

dirk-thomas:
Sure: [Build Status](http://ci.ros2.org/job/ci_osx/2137/)

test/test_logging_macros.cpp
```diff
@@ -143,9 +143,9 @@ TEST_F(TestLoggingMacros, test_logging_throttle) {
 TEST_F(TestLoggingMacros, test_logging_skipfirst_throttle) {
   for (int i : {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}) {
     RCUTILS_LOG_FATAL_SKIPFIRST_THROTTLE(
-      RCUTILS_STEADY_TIME, 30 /* ms */, "throttled message %d", i)
+      RCUTILS_STEADY_TIME, 50 /* ms */, "throttled message %d", i)
     using namespace std::chrono_literals;
-    std::this_thread::sleep_for(20ms);
+    std::this_thread::sleep_for(30ms);
```
wjwwood: Another option is to put the whole test in a loop, retrying a limited number of times. Presumably if it passes any of the attempts then there is no issue. This test is really testing the logic, not the timeliness, of the `sleep_for` function.

Yet another option is to measure the time you actually sleep for (measure around the `sleep_for` function call) and accumulate the total since the first call. Then each time you update the total you can figure out if the next `i` should be in the output or not. This would at least be slightly more robust to a overrun on `sleep_for` (remember it only guarantees that it returns after the period, not how close to it).

dirk-thomas:
Since it passed with `--retest-until-fail 50` I will go ahead and merge this as-is. Please feel free to improve the test logic if you think that makes sense.

PR #31, opened 2017-06-14
Declare C standard explicitly

allenh1:
Failed to compile on my system without explicitly declaring the std=gnu99

(`_Thread_local` is not in ISO C99, but *is* in GNU99)


allenh1:
Woops, duplicate of #31

mikaelarguedas:
duplicate of #26

allenh1:
^^ that one

PR #33, opened 2017-06-15
Refactor get topic names and types

wjwwood:
Connects to ros2/ros2#361


include/rcutils/types/string_array.h
```diff
@@ -31,6 +31,7 @@ typedef struct RCUTILS_PUBLIC_TYPE rcutils_string_array_t
 {
   size_t size;
   char ** data;
+  rcutils_allocator_t allocator;
```
Karsten1987: is that a rule that the types should hold the allocator object?
If so, ignore the comments on the `rmw` PR.

src/split.c
```diff
@@ -59,6 +65,7 @@ rcutils_split(
       ++string_array->size;
     }
   }
+  // TODO(wjwwood): refactor this function so it can use rcutils_string_array_init() instead
```
Karsten1987: will you include this TODO in your PR?

include/rcutils/types/string_array.h
wjwwood: It's not a rule, but it prevents the caller from needing to keep the allocator around and pass it to the `fini` function. This also helps ensure that the same allocator is used with `init` and `fini`.

src/split.c
wjwwood: No, I wasn't planning on it. The split functions logic involves resizing the `size` argument dynamically as it goes. I didn't want to figure out how to change it so it could figure out the end size once and not need to change storage as it goes.

PR #34, opened 2017-06-15
use -Wpedantic for C

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2740/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/276/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2208/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2872/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ find_package(ament_cmake REQUIRED)
 include_directories(include)
 
 if(NOT WIN32)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
   # The warning flags are added per target, because otherwise the local
   # gmock library that is built will have compiler warnings.
   set(rcutils_extra_warning_flags "-Wall -Wextra -Wpedantic")


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ find_package(ament_cmake REQUIRED)
 include_directories(include)
 
 if(NOT WIN32)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
   # The warning flags are added per target, because otherwise the local
   # gmock library that is built will have compiler warnings.
   set(rcutils_extra_warning_flags "-Wall -Wextra -Wpedantic")
```
clalancette: I know you didn't add this line of code, but I'm not sure how to square the comment about gmock with the addition of -Wpedantic to the overall CMAKE_C_FLAGS.  Do you understand what that comment is talking about, and why we need `rcutils_extra_warning_flags` ?

dirk-thomas: This is legacy stuff imo. I saw it when writing the logging macros. I will hopefully find some time soon to get rid of it.

clalancette: Ah, OK, thanks.  Would you mind opening an issue about that cleanup, just in case you don't get to it?  Otherwise, this is fine and I'll approve.

dirk-thomas: I would rather not open a ticket for this. The code works as-is. If I would open tickets for every little thing we could improve the backlog would grow exponentially.

dirk-thomas: See #36.

PR #35, opened 2017-06-15
add c test using logging macros

### LINKED ISSUE #32: rcutils logging macro warning
I was using the new logging macros like this:

```c
#include "rcutils/logging_macros.h"
// ...
    RCUTILS_LOG_ERROR(
      "failed to clean up on error (leaking memory): '%s'", rcutils_get_error_string_safe());
```

But I get a warning like this:

```
In file included from /Users/william/ros2_ws/src/ros2/rcutils/src/split.c:26:
/Users/william/ros2_ws/build/rcutils/include/rcutils/logging_macros.h:948:13: warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]
    format, ## __VA_ARGS__)
            ^
/Users/william/ros2_ws/build/rcutils/include/rcutils/logging_macros.h:93:68: warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]
      rcl_log(&__rcutils_logging_location, severity, name, format, ## __VA_ARGS__); \
                                                                   ^
2 warnings generated.
```

I looked into it briefly, but I didn't dig too deep. Am I doing something wrong or is this just a warning we should fix in the logging macros?

The warning only references the line on which the header is included, but if I remove all uses of the logging macro (but leave the include) the warning goes away.

dirk-thomas:
Fixes #32.

For now the PR only adds a test which reproduced the problem:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2745/)
* OS X [Build Status](http://ci.ros2.org/job/ci_osx/2214/)

Looking into a fix...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,12 @@ if(BUILD_TESTING)
     APPEND_LIBRARY_DIRS ${extra_lib_dirs})
   target_link_libraries(test_logging_macros ${PROJECT_NAME})
 
+  add_executable(test_logging_macros_c test/test_logging_macros.c)
+  target_link_libraries(test_logging_macros_c ${PROJECT_NAME})
+  ament_add_test(test_logging_macros_c
+    COMMAND "$<TARGET_FILE:test_logging_macros_c>"
+    GENERATE_RESULT_FOR_RETURN_CODE_ZERO)
+
   # This subdirectory extends both extra_test_libraries, extra_test_env, and extra_lib_dirs.
   add_subdirectory(test/memory_tools)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,15 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rcutils)
 
+# Default to C11
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 11)
+endif()
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
One simple solution to fix this is to switch to C11. I won't look into alternative solutions for now until we reconsidered if switching is an option...

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2751/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/284/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2221/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2879/)

dirk-thomas:
Since this overlaps with other PR atm I would wait until they are merged, rebase/update and run CI again.

dirk-thomas:
Commit 94e2f1c folds the `format` arg into the vargs to make a compiler stop :sob: While it would be good to keep the explicit argument for documentation as well as completion we don't know a better way around for now...

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2761/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/293/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2892/)

PR #38, opened 2017-06-24
setup doxygen, add a mainpage, and fixup docs

wjwwood:
Connects to https://github.com/ros2/ros2/issues/370


wjwwood:
Yeah, I intentionally left out a few items from the overview that I thought were less useful or otherwise not well enough documented or were only used by `rcutils` itself.

PR #39, opened 2017-06-27
add filename and line number to error message

dirk-thomas:
Otherwise the location from where the new error message is coming from is unknown.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2827/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/331/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2283/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2963/)

Ready for review.

### DIFF
--- a/src/error_handling.c
+++ b/src/error_handling.c
@@ -152,12 +152,14 @@ rcutils_set_error_state(
         "\n"
         "with this new error message:\n"
         "\n"
-        "  '%s'\n"
+        "  '%s, at %s:%zu'\n"
         "\n"
         "rcutils_reset_error() should be called after error handling to avoid this.\n"
         "<<<\n",
         old_error_string,
-        error_string);
+        error_string,
+        file,
+        line_number);
     }
 #endif
     __rcutils_reset_error(&old_error_state);


nuclearsandwich:
Is this targeting beta2? At some point we need to put a pin in things or the arm packages are never going to finish in time to be properly QA'd before release.

wjwwood:
I think this could definitely go in afterwards. It is only a slight improvement to the print output in error conditions.

dirk-thomas:
This is certainly only an improvement which could wait but there will be other patches with crucial fixes which should go in. Therefore I think it makes sense to merge that in order to include the improvement.

nuclearsandwich:
So I can track both at once, where is the other "crucial fix" that was mentioned?

dirk-thomas:
> So I can track both at once, where is the other "crucial fix" that was mentioned?

Please just watch the GitHub notifications for merged PRs. There have been quite a few today. We also try to reference them in the testing spreadsheet - not 100% sure if they are complete there though.

PR #44, opened 2017-08-06
nanoseconds should always be uint64_t

Karsten1987:
connects to ros2/rclcpp#352

### DIFF
--- a/include/rcutils/time.h
+++ b/include/rcutils/time.h
@@ -27,11 +27,11 @@ extern "C"
 #include "rcutils/visibility_control.h"
 
 /// Convenience macro to convert seconds to nanoseconds.
-#define RCUTILS_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))
+#define RCUTILS_S_TO_NS(seconds) ((uint64_t)(seconds) * (1000 * 1000 * 1000))
 /// Convenience macro to convert milliseconds to nanoseconds.
-#define RCUTILS_MS_TO_NS(milliseconds) (milliseconds * (1000 * 1000))
+#define RCUTILS_MS_TO_NS(milliseconds) ((uint64_t)(milliseconds) * (1000 * 1000))
 /// Convenience macro to convert microseconds to nanoseconds.
-#define RCUTILS_US_TO_NS(microseconds) (microseconds * 1000)
+#define RCUTILS_US_TO_NS(microseconds) ((uint64_t)(microseconds) * 1000)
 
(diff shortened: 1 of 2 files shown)


allenh1:
But what about systems that don't have `uint64_t`? I think bash uses `unsigned long` for nanoseconds. Maybe we should adopt that behavior?

mikaelarguedas:
given that `rcl_wait` uses an int64_t for it's timeout (negative values meaning wait forever) and that the docblocks examples use these macros to specify the timeout, it looks risky to me to cast this as a uint64_t. see https://github.com/ros2/rcl/pull/86#discussion_r86605398 for an analog discussion

wjwwood:
> But what about systems that don't have `uint64_t`?

I don't know of any systems which cannot represent `uint64_t` (either natively on the CPU or through emulation). I think it's an unreasonable constraint on the use of the 64-bit type. We've discussed this in the past internally as well as externally:

https://discourse.ros.org/t/of-clocks-and-simulation-betimes-and-otherwise/1587/7

> I think bash uses unsigned long for nanoseconds. Maybe we should adopt that behavior?

Bash probably uses it in combination with a 32-bit seconds field too, but I don't know since you didn't link to what you're referring. I don't think it makes sense to use `unsigned long` for the nanoseconds representation of a time point (note only nanoseconds, not a nanoseconds + seconds style structure), but maybe you meant something else.

> given that `rcl_wait` uses an `int64_t` for it's timeout (negative values meaning wait forever)

That's because it is a duration. In `rcl` we represent durations as signed 64-bit integers so you can safely do arithmetic on them. Time points, however, use unsigned integers because they are not often the result of arithmetic but rather a component of it, and because using an unsigned values give a much longer scope for the time point since the unix epoch. Basically it is a trade-off between representing more time points in the future versus representing time points before 1970. I made the decision for the latter in `rcl`.

> and that the docblocks examples use these macros to specify the timeout, it looks risky to me to cast this as a `uint64_t`

Indeed, not all representations of time that could be used with these macros must be unsigned or even integers. You could perfectly well do this:

```c
double nanoseconds = RCL_S_TO_NS(1.3);
```

You might need to consider the consequences of doing that (loss of floating point precision, etc) but it's totally valid.

wjwwood:
Edited the previous post, meant `former` when I said `latter`.

Karsten1987:
closing in favor of commit: https://github.com/ros2/rclcpp/pull/351/commits/89b6b6f2a6bf2e6187a07d2d2a73a886491e05fe

PR #48, opened 2017-08-10
Make sure to free string_map->impl in fini.

clalancette:

connects to ros2/ros2#387

CI jobs are in the above issue


clalancette:
CI is green for this, approved, merging.

PR #49, opened 2017-08-25
Factor out python data used in logging_macros empy template

dhood:
In wrapping rcutils logging macros in rclpy, I found I needed access to most of the same data used in the logging_macros empy template, so I moved the data to a python module for use in this package and others.

To use the module at build time, I force it onto the path [like this](https://github.com/ros2/rcutils/compare/master...common_python_logic#diff-16a8c3749d32285b558137f65c5c60d2R223) (the `rclpy_module_path` variable is set in the CMakeLists [here](https://github.com/ros2/rcutils/compare/master...common_python_logic#diff-af3b638bc2a3e6c650974192a53c7291R62)). If there's a cleaner way to do this, please LMK.

There is no change to the generated `logging_macros.h` with this PR.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/471/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2477/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3163/)


dirk-thomas:
While the template doesn't have too much history it might still be nice to preserve it (by using a separate commit to copy the existing file to the new location without changing the content).

dhood:
this commit only moves the code out without modifying it: https://github.com/ros2/rcutils/pull/49/commits/7893fe44526b556e9dfa18f4d0ce64ca8765e286 
I changed the whitespace is all

dirk-thomas:
> this commit only moves the code out without modifying it: 7893fe4
I changed the whitespace is all

The fact that the file was changed (different content hash) makes it impossible to follow the history for the new file: see https://github.com/ros2/rcutils/commits/2b34c9ea066d0d07b25e093e3568782e63e1de7b/rcutils/logging.py

dhood:
@dirk-thomas as you know, I added some commits (https://github.com/ros2/rcutils/pull/49/commits/8a6ff6db3ec877ad1501f7190b1f74fdadb5dadb and https://github.com/ros2/rcutils/pull/49/commits/9ff5f76663218a3dead2912d7b30ab9f02d2feb3) to maintain the history of the old file. Summarising our discoveries: the history of the old file isn't showing with the 'history' display on github but it is showing with the 'blame' display. Achieving this required the original file to be moved around as well, as part of a workaround; just duplicating the file wasn't enough to get the old history to show in the copied file (`logging.py`).

resource/logging_macros.h.em
```diff
@@ -219,152 +219,9 @@ typedef bool (* RclLogFilter)();
...
-
-def get_macro_arguments(suffix):
-  args = []
-  for k, default_value in default_args.items():
-    value = feature_combinations[suffix].args.get(k, default_value)
-    args.append(value)
-  return ''.join([a + ', ' for a in args])
+import sys
+sys.path.insert(0, rcutils_module_path)
+from rcutils.logging import feature_combinations, get_macro_arguments, get_macro_parameters, severities
```
dirk-thomas: Use separate lines for each `import`.

resource/logging_macros.h.em
```diff
@@ -374,7 +231,7 @@ def get_macro_arguments(suffix):
 // empty logging macros for severity @(severity) when being disabled at compile time
 @[ for suffix in feature_combinations]@
 /// Empty logging macro due to the preprocessor definition of RCUTILS_LOG_MIN_SEVERITY.
-# define RCUTILS_LOG_@(severity)@(suffix)(@(get_macro_parameters(suffix))format, ...)
+# define RCUTILS_LOG_@(severity)@(suffix)(@(''.join([p + ', ' for p in get_macro_parameters(suffix).keys()]))format, ...)
```
dirk-thomas: Why move this logic from the function (where it was only written once) here as well as below where it is now duplicated?

dhood: because the way the data is being formatted is very specific to this template file (I might still use the logic in `get_macro_arguments` in rclpy, for example, but I need to format the data in a different way). there could be a function local to this file that formats the string but I didn't think it was worthwhile

dhood:
conflicts resolved

resource/logging_macros.h.em
dhood: I squashed a change for this into https://github.com/ros2/rcutils/pull/49/commits/b71991f4a98340a4831fea52820da136ecfd4a37

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3096/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/475/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2481/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3168/) (unrelated flaky test)

dhood:
note for future self: in the end this wasn't needed for ros2/rclpy#103 because it reimplemented the logging features in pythonland rather than templating calls to the different macros.

PR #52, opened 2017-08-31
Don't require logging users to manually initialise before setting severity

dhood:
If this is appropriate, does it also belong on the getter so you don't get the uninitialised value?


dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3142/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/490/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2499/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3195/)

PR #53, opened 2017-09-08
Configurable console output format from environment variable

dhood:
no tests yet but you can do something like:

```
$ RCUTILS_CONSOLE_OUTPUT_FORMAT="[{}{name}] [{severity}]: {line_number} {message} {not_token" ros2 run demo_nodes_cpp add_two_ints_client

[{}rmw_fastrtps_cpp] [DEBUG]: 149 ************ Client Details ********* {not_token
[{}rmw_fastrtps_cpp] [DEBUG]: 152 Sub Topic add_two_intsReply {not_token
[{}rmw_fastrtps_cpp] [DEBUG]: 155 Sub Partition rr {not_token
[{}rmw_fastrtps_cpp] [DEBUG]: 158 Pub Topic add_two_intsRequest {not_token
[{}rmw_fastrtps_cpp] [DEBUG]: 161 Pub Partition rq {not_token
```

It still needs to use a dynamic buffer, or if we're happy with static then this is ready for review


src/logging.c
```diff
@@ -17,10 +17,14 @@ extern "C"
...
 #endif
 
+#include <string.h>
+
 #include "rcutils/allocator.h"
+#include "rcutils/get_env.h"
 #include "rcutils/logging.h"
 
 bool g_rcutils_logging_initialized = false;
+const char * g_rcutils_logging_output_format_string = "[{severity}] [{name}]: {message} {line_number}";
```
mikaelarguedas: I guess it's location as a whole and not only line_number

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
-      stream, "[%s] [%s]: %s (%s() at %s:%zu)\n", severity_string, name, message_buffer,
-      location->function_name, location->file_name, location->line_number);
+    // TODO before merge
+    //location = empty_location;
+  }
+
+  // Process the format string looking for known tokens
+  char token_start_delimiter = '{';
+  char token_end_delimiter = '}';
+  char output_buffer[1024];
```
mikaelarguedas: should all of this be const ?

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] == token_start_delimiter) {
+      char token[20];
+      memset(token, '\0', sizeof(token));
+      size_t j = i + 1;
+      // Look for a token end delimeter
+      for (j = i + 1; j < size && str[j] != token_end_delimiter; j++);
+      if (j >= size) {
+          // No end delimiters found; there won't be any more tokens so shortcut the rest of the checking
+          strcat(output_buffer, str + i);
```
mikaelarguedas: `strcat` is not safe, you can use either a custom snprintf call or rcultils_concat, same everywhere below

src/logging.c
dirk-thomas: For the default I would suggest to leave the `line_number` out. Without the filename it isn't very helpful anyway.

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+          break;
+      }
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i+1 , token_len);
```
mikaelarguedas: Nit: space between delimiters, also I think strncpy throws warning on windows like strcat

src/logging.c
```diff
@@ -31,6 +35,13 @@ void rcutils_logging_initialize()
   if (!g_rcutils_logging_initialized) {
     g_rcutils_logging_output_handler = &rcutils_logging_console_output_handler;
     g_rcutils_logging_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
+    // Check for the environment variable to disable output formatting
+    const char * output_format;
+    const char * ret;
+    ret = rcutils_get_env("RCUTILS_CONSOLE_OUTPUT_FORMAT", &output_format);
+    if (!ret && strcmp(output_format, "") != 0) {
+      g_rcutils_logging_output_format_string = output_format;
```
dirk-thomas: According to the docblock of `rcutils_get_env` this won't work:

> The c-string which is returned in the env_value output parameter is only valid until the next time this function is called, because it is a direct pointer to the static storage.

So this needs to copy the string.

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+          // No end delimiters found; there won't be any more tokens so shortcut the rest of the checking
+          strcat(output_buffer, str + i);
+          break;
+      }
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i+1 , token_len);
+        char token_buffer[1024];
+        int n = 0;
+        bool known_token = false;
+        if (strcmp("severity", token) == 0)
```
mikaelarguedas: Nit: style to make uncrustify happy, same everywhere below

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+        {
+          n = sprintf(token_buffer, "%s", severity_string);
+          known_token = true;
+        }
+        if (!known_token && strcmp("line_number", token) == 0)
```
mikaelarguedas: for safety and performance `strncmp`

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+          break;
+      }
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i+1 , token_len);
+        char token_buffer[1024];
+        int n = 0;
+        bool known_token = false;
+        if (strcmp("severity", token) == 0)
+        {
+          n = sprintf(token_buffer, "%s", severity_string);
```
mikaelarguedas: sprintf not safe, use `snprintf`, same everywhere below

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+  char token_start_delimiter = '{';
+  char token_end_delimiter = '}';
+  char output_buffer[1024];
+  memset(output_buffer, '\0', sizeof(output_buffer));
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] == token_start_delimiter) {
+      char token[20];
```
dirk-thomas: Can you please add a comment why you chose 20.

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+  // Process the format string looking for known tokens
+  char token_start_delimiter = '{';
+  char token_end_delimiter = '}';
+  char output_buffer[1024];
+  memset(output_buffer, '\0', sizeof(output_buffer));
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] == token_start_delimiter) {
```
dirk-thomas: I would invert the check since allows you to indent the whole following block:

```
if (str[i] != token_start_delimiter) {
  strncat(output_buffer, str + i, 1);
  continue;
}

char token[20];
...
```

src/logging.c
mikaelarguedas: Should this 20 come from somewhere rather than being a magic number here ?

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
...
+          n = sprintf(token_buffer, "%s", message_buffer);
+          known_token = true;
+        }
+        if (!known_token && strcmp("name", token) == 0)
+        {
+          n = sprintf(token_buffer, "%s", name);
+          known_token = true;
+        }
+        if (known_token) {
+          strncat(output_buffer, token_buffer, n);
```
dirk-thomas: This should check not to exceed the length of the allocated `output_buffer`.

src/logging.c
dirk-thomas: By using `else if` here you can maybe get rid of `known_token` all together?

src/logging.c
```diff
@@ -30,7 +35,14 @@ void rcutils_logging_initialize()
 {
   if (!g_rcutils_logging_initialized) {
     g_rcutils_logging_output_handler = &rcutils_logging_console_output_handler;
-    g_rcutils_logging_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
+    g_rcutils_logging_severity_threshold = RCUTILS_LOG_SEVERITY_DEBUG;
```
mikaelarguedas: Why do we change the default threshold here ?

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+        // there won't be any more tokens so shortcut the rest of the checking
+        snprintf(output_buffer, sizeof(output_buffer), "%s", str + i);
+        break;
+      }
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i + 1, token_len);
+      char token_buffer[1024];
+      int n = 0;
+      bool known_token = false;
+      if (strcmp("severity", token) == 0) {
```
mikaelarguedas: for performance I'd recommend strncmp given that it's comparing to a string of known length, same everywhere below

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+  char output_buffer[1024];
+  memset(output_buffer, '\0', sizeof(output_buffer));
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] == token_start_delimiter) {
+      char token[20];
+      memset(token, '\0', sizeof(token));
+      size_t j = i + 1;
```
mikaelarguedas: Nit: this is reassigned to the same value just below(line 168)

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+      // Look for a token end delimeter
```
mikaelarguedas: typo delimiter

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+        snprintf(output_buffer, sizeof(output_buffer), "%s", str + i);
+        break;
+      }
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i + 1, token_len);
+      char token_buffer[1024];
+      int n = 0;
+      bool known_token = false;
+      if (strcmp("severity", token) == 0) {
+        n = snprintf(token_buffer, sizeof(token_buffer), "%s", severity_string);
```
mikaelarguedas: rcutils_snprintf would allow to handle different platforms more sefaly, same everywhere below

src/logging.c
```diff
@@ -17,10 +17,15 @@ extern "C"
...
 
+#include <string.h>
+
 #include "rcutils/allocator.h"
+#include "rcutils/get_env.h"
 #include "rcutils/logging.h"
 
 bool g_rcutils_logging_initialized = false;
+const char * g_rcutils_logging_output_format_string = \
+  "[{severity}] [{name}]: {message} {line_number}";
```
mikaelarguedas: original comment still applies, I think that should be called `location` rather than `line_number` if kept

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+      size_t token_len = j - i - 1;  // not including delimeters
+      strncpy(token, str + i + 1, token_len);
+      char token_buffer[1024];
+      int n = 0;
+      bool known_token = false;
+      if (strcmp("severity", token) == 0) {
+        n = snprintf(token_buffer, sizeof(token_buffer), "%s", severity_string);
+        known_token = true;
+      }
+      if (!known_token && strcmp("line_number", token) == 0) {
```
mikaelarguedas: else if here and everywhere below if the tests are actually exclusive

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+      char token[20];
+      memset(token, '\0', sizeof(token));
+      size_t j = i + 1;
+      // Look for a token end delimeter
+      for (j = i + 1; j < size && str[j] != token_end_delimiter; j++) {
+      }
+      if (j >= size) {
+        // No end delimiters found;
+        // there won't be any more tokens so shortcut the rest of the checking
+        snprintf(output_buffer, sizeof(output_buffer), "%s", str + i);
```
mikaelarguedas: `rcutils_snprintf` everywhere to handle different platforms more safely

src/logging.c
```diff
@@ -135,12 +147,62 @@ void rcutils_logging_console_output_handler(
...
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] == token_start_delimiter) {
+      char token[20];
+      memset(token, '\0', sizeof(token));
+      size_t j = i + 1;
+      // Look for a token end delimeter
+      for (j = i + 1; j < size && str[j] != token_end_delimiter; j++) {
```
mikaelarguedas: could `rcutils_find` do the job ?

src/logging.c
dirk-thomas: You could `strncat` instead but at the end of the day `N` needs to come from somewhere - using any of the suggested functions instead doesn't make it safer imo. So getting `N` in the beginning by calling `strlen` is ok if the result is checked against a large upper bound.

src/logging.c
dirk-thomas: Why would `strncmp` be "safer" here? What are you passing as `N`? The string literal first argument already limits the length to 12.

mikaelarguedas: you're right it is a string literal so the max length is already forced, habbit of using `strncmp` everywhere :S

src/logging.c
mikaelarguedas: discard this, duplicated and not relevant

src/logging.c
dhood: not sure that it really gains anything here but I might have overlooked something

src/logging.c
```diff
@@ -135,12 +148,68 @@ void rcutils_logging_console_output_handler(
...
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  for (size_t i = 0; i < size; ++i) {
+    if (str[i] != token_start_delimiter) {
+      strncat(output_buffer, str + i, 1);
+      continue;
+    }
+    // Found a token start delimiter: determine if there's a token or not.
+    char token[20];  // the largest known token
```
mikaelarguedas: Still not sure what token we're referring to here. Maybe declaring this with a define next to where the tokens are defined would be more future-proof ? Same for the 1024 above and below, maybe a define would be more appropriate

dhood: thanks, I've seen you comment I just haven't gotten to it yet. I moved this to in progress and will ping once the comments have been addressed

mikaelarguedas: Sounds good, it's just because I saw a commit addressing review comment modifying this line that I responded

src/logging.c
dhood: I've tried to do this dynamically in 1162ac4a0681e46a14a006aebaafd9666d1f758c but FWIU I'll need to deallocate it myself, and I don't see anywhere appropriate to do that since there isn't a fini for the logging module. So I'm going to switch this to a large static allocation that will truncate.

src/logging.c
dhood: absolutely; it wasn't clear on my part that they weren't all addressed yet

dhood:
Hmmm.. I've added dynamic allocation for the output buffer in f922e7be84bf580892fa8e7ea680560aa1a9e634 (this fixes the large message test), but I wasn't able to de-allocate the buffer without memory errors until I deallocated it after the buffer for the "message" expansion (in 8e723ce620f51df7b1a75e04a6fa078ec38feba5). I'm not 100% about the subtleties of the ordering to know if it was a coincidence or not: any input would be welcome.

src/logging.c
```diff
@@ -73,9 +93,15 @@ void rcutils_log(
 }
 
 void rcutils_logging_console_output_handler(
-  rcutils_log_location_t * location,
+  rcutils_log_location_t * location_,
   int severity, const char * name, const char * format, va_list * args)
 {
+  rcutils_log_location_t * location = location_;
```
dirk-thomas: Why are two separate variables needed here? Just having an argument name `location` and optionally pointing it to the dummy_location should be fine.

src/logging.c
```diff
@@ -134,17 +160,102 @@ void rcutils_logging_console_output_handler(
...
 
-  if (!location) {
-    fprintf(stream, "[%s] [%s]: %s\n", severity_string, name, message_buffer);
-  } else {
-    fprintf(
-      stream, "[%s] [%s]: %s (%s() at %s:%zu)\n", severity_string, name, message_buffer,
-      location->function_name, location->file_name, location->line_number);
+  // Process the format string looking for known tokens
+  char token_start_delimiter = '{';
+  char token_end_delimiter = '}';
```
dirk-thomas: Those two variables should be `const`.

src/logging.c
```diff
@@ -31,6 +38,19 @@ void rcutils_logging_initialize()
   if (!g_rcutils_logging_initialized) {
     g_rcutils_logging_output_handler = &rcutils_logging_console_output_handler;
     g_rcutils_logging_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
+
+    int max_output_format_len = sizeof(g_rcutils_logging_output_format_string);
+    memset(g_rcutils_logging_output_format_string, '\0', max_output_format_len);
```
dirk-thomas: It is not necessary to overwrite all characters with `\0`. The first one would be sufficient to make it a null terminated string.

Same below.

src/logging.c
```diff
@@ -134,17 +160,102 @@ void rcutils_logging_console_output_handler(
...
+      // there won't be any more tokens so shortcut the rest of the checking
+      // TODO(dhood) before merge: dynamically allocate more space if required.
+      strncat(output_buffer, token_buffer, n);
+      break;
+    }
+    // Found what looks like a token; determine if it's recognized.
+    size_t token_len = j - i - 1;  // not including delimiters
+    strncpy(token, str + i + 1, token_len);
+    if (strcmp("severity", token) == 0) {
+      n = rcutils_snprintf(token_buffer, sizeof(static_token_buffer), "%s", severity_string);
```
dirk-thomas: Using `snprintf` for existing strings isn't necessary. It could simply `memcpy` the characters. It would also be easier to directly copy them to the `output_buffer` rather than using another buffer (`token_buffer`).

Same below - also for the other token types.

src/logging.c
```diff
@@ -134,17 +160,102 @@ void rcutils_logging_console_output_handler(
...
+        // TODO(dhood): uncomment this once it's tested.
+        /*
+        if (output_buffer != static_output_buffer) {
+          allocator.deallocate(output_buffer, allocator.state);
+        }
+        */
+        output_buffer = (char *)dynamic_output_buffer;
+        // printf("New output buffer size: %zu\n", output_buffer_size);
+      }
+      strncat(output_buffer, token_buffer, n);
```
allenh1: Windows will complain about the `strcpy` here (at least it has for me in the past). `memcpy` does the trick.

allenh1: That comment was supposed to go on line 234.

src/logging.c
dhood: the reason for the intermediate token buffer instead of directly copying to the output buffer was to avoid the need for checking if dynamic allocation is necessary in multiple places. I'll make a macro for doing that check so we can go straight into the output buffer and remove the token buffer completely.

src/logging.c
dhood: thanks for the tip!

src/logging.c
dhood: I thought there'd be a chance the user was counting on it to stay null and didn't want to change it under their nose, but I suppose I could have changed it back afterwards. I avoided modifying it in 3049cb0afec856136e0348d7ef04614c0ed915c1 (many other things have changed in the meantime though)

CMakeLists.txt
```diff
@@ -139,6 +139,12 @@ if(BUILD_TESTING)
     WORKING_DIRECTORY "$<TARGET_FILE_DIR:test_logging_long_messages>"
     TIMEOUT 10)
 
+  ament_add_nose_test(test_logging_output_format
+    "test/test_logging_output_format.py"
+    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
+    WORKING_DIRECTORY "$<TARGET_FILE_DIR:test_logging_long_messages>"
```
dhood: this test re-uses the long_message test binary and just modifies the output format from an environment variable. In the custom output, the long message is output twice, to test both dynamic allocation and re-allocation.

src/logging.c
```diff
@@ -31,6 +41,27 @@ void rcutils_logging_initialize()
...
+      }
+      memcpy(g_rcutils_logging_output_format_string, output_format, chars_to_copy);
+      g_rcutils_logging_output_format_string[chars_to_copy] = '\0';
+    } else {
+      if (NULL != ret) {
+        fprintf(
+          stderr,
+          "Failed to get output format from env. variable: %s. Using default output format.\n",
+          ret);
+      }
```
dhood: It might make more sense to return early here on error, but since there isn't a way to signal the error, the logging calls will still be called and continually try to re-initialise the logging system. So, I chose to just fallback to the default format.

src/logging.c
```diff
@@ -72,6 +103,44 @@ void rcutils_log(
...
+/*!
+  RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER(
+   n, output_buffer_size, allocator, output_buffer, static_output_buffer)
+  Macro for (re)allocating a dynamic buffer if the ouput buffer doesn't have enough space for an
+  additional n characters and null terminator. Whether to allocate or re-allocate is determined by
+  if the output_buffer points to the original static_output_buffer or not. If a new buffer is
+  allocated it will be padded with NUL.
+*/
+
+#define RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER( \
```
dhood: Pay attention for bad practices here because before this I didn't have much experience writing macros like this.

src/logging.c
```diff
@@ -72,6 +103,44 @@ void rcutils_log(
...
+    do { \
+      output_buffer_size *= 2; \
+    } while (required_output_buffer_size >= output_buffer_size); \
+    if (output_buffer == static_output_buffer) { \
+      void * dynamic_output_buffer = allocator.allocate(output_buffer_size, allocator.state); \
+      if (!dynamic_output_buffer) { \
+        fprintf(stderr, "failed to allocate buffer for logging output\n"); \
+        goto cleanup; \
+      } \
+      memset(dynamic_output_buffer, '\0', output_buffer_size); \
```
dhood: I chose to memset the buffer because there are other parts of the code where I'd have to manually copy the null terminator across, and I felt this was safer. I can switch to adding the null terminators manually if we review it extra carefully.

src/logging.c
```diff
@@ -104,47 +173,149 @@ void rcutils_logging_console_output_handler(
       return;
   }
 
-  char buffer[1024];
-  char * message_buffer = buffer;
+  // Declare variables that will be needed for cleanup ahead of time.
```
dhood: if this makes the code less readable we can revert https://github.com/ros2/rcutils/pull/53/commits/18777e23224232338ece56d0d1c4de590b469803 and not use `cleanup` on message format failure.

dhood:
this PR has been updated to dynamically allocate enough memory needed at each write to the output buffer. I've added some comments of my own to assist reviewers.

example test output (no warnings on windows): http://ci.ros2.org/job/ci_windows/3339/testReport/(root)/test_logging_output_format/test_logging_output_format/

include/rcutils/logging.h
```diff
@@ -188,23 +188,25 @@ void rcutils_log(
  * The messages with a severity `DEBUG` and `INFO` are written to `stdout`.
  * The messages with a severity `WARN`, `ERROR`, and `FATAL` are written to
  * `stderr`.
- * For each message the severity and name is prepended and the location
+ * The console output format of the logged message can be configured through
+ * the `RCUTILS_CONSOLE_OUTPUT_FORMAT` environment variable.
+ * By default, the severity and name is prepended and the location
```
dirk-thomas: The location seem to be appended in any case (not only when available).

src/logging.c
```diff
@@ -17,10 +17,20 @@ extern "C"
...
 
+#include <string.h>
+
 #include "rcutils/allocator.h"
+#include "rcutils/find.h"
+#include "rcutils/get_env.h"
 #include "rcutils/logging.h"
+#include "rcutils/snprintf.h"
+
+#define RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN 1024
```
dirk-thomas: This constant seems to be used for two purposes: for the max length of the environment variable as well as for the max length of every formatted message.

I would argue it is only necessary for the second case. Why do we need to limit the length of the environment variable (which is already limited by the operating system)?

src/logging.c
```diff
@@ -72,6 +103,44 @@ void rcutils_log(
...
+  Macro for (re)allocating a dynamic buffer if the ouput buffer doesn't have enough space for an
+  additional n characters and null terminator. Whether to allocate or re-allocate is determined by
+  if the output_buffer points to the original static_output_buffer or not. If a new buffer is
+  allocated it will be padded with NUL.
+*/
+
+#define RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER( \
+    n, output_buffer_size, allocator, output_buffer, static_output_buffer) \
+  size_t required_output_buffer_size = strlen(output_buffer) + n + 1; \
+  if (required_output_buffer_size >= output_buffer_size) { \
```
dirk-thomas: Why does it have to reallocate if `required_output_buffer_size` == `output_buffer_size`?

Same three lines below.

src/logging.c
dirk-thomas: Depending on how new data is copied into the buffer I think adding the `\0` there is safer rather than relying on the buffer to be filled with zeros. I don't know if this could happen but if a long message is copied into the buffer first and afterwards a shorter message is copied in (without a resize happening) it must set the trailing `\0` anyway. Therefore I would prefer adding the trailing `\0` whenever a message is written to the buffer. It also keep the logic local and the reader of the code doesn't need to consider in what state the buffer is currently in.

src/logging.c
```diff
@@ -72,6 +103,44 @@ void rcutils_log(
...
+      void * dynamic_output_buffer = allocator.allocate(output_buffer_size, allocator.state); \
+      if (!dynamic_output_buffer) { \
+        fprintf(stderr, "failed to allocate buffer for logging output\n"); \
+        goto cleanup; \
+      } \
+      memset(dynamic_output_buffer, '\0', output_buffer_size); \
+      memcpy(dynamic_output_buffer, output_buffer, strlen(output_buffer)); \
+      output_buffer = (char *)dynamic_output_buffer; \
+    } else { \
+      void * new_dynamic_output_buffer = (char *)allocator.reallocate( \
```
dirk-thomas: Could this be of type `char *`?

src/logging.c
```diff
@@ -104,47 +173,149 @@ void rcutils_logging_console_output_handler(
...
+  // Start with a fixed size output buffer and if during token expansion we need longer, we'll
+  // dynamically allocate space.
+  output_buffer = static_output_buffer;
+  memset(static_output_buffer, '\0', sizeof(static_output_buffer));
+  size_t output_buffer_size = sizeof(static_output_buffer);
+  const char * str = g_rcutils_logging_output_format_string;
+  size_t size = strlen(g_rcutils_logging_output_format_string);
+
+  // Walk through the format string and expand tokens when they're encountered.
+  size_t n = 0;
```
dirk-thomas: `n` seems to be only used once within the loop. Can it be declared there instead?

src/logging.c
```diff
@@ -104,47 +173,149 @@ void rcutils_logging_console_output_handler(
...
+      token_expansion = name;
+    } else if (strcmp("message", token) == 0) {
+      token_expansion = message_buffer;
+    } else if (strcmp("function_name", token) == 0) {
+      token_expansion = location ? location->function_name : "\"\"";
+    } else if (strcmp("file_name", token) == 0) {
+      token_expansion = location ? location->file_name : "\"\"";
+    } else if (strcmp("line_number", token) == 0) {
+      if (location) {
+        char line_number_expansion[10];  // Allow 9 digits for the expansion (otherwise, truncate).
```
dirk-thomas: I am certain this will be enough to cover line numbers. I wonder though if we need to limit it artificially. Could the code instead us `SIZE_MAX` and derive the maximum number of digits from it? Then there is no need to truncate and the size of the buffer should be only slightly bigger than currently. But it will allow to get rid of the case to handle the truncation.

src/logging.c
dhood: alright, I'll switch to that approach and will call attention to that commit for extra-review.

src/logging.c
dhood: I think there's been a misinterpretation on the second use of the constant: there isn't currently any limit to the length of a formatted message. When `RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN` is used [here](https://github.com/ros2/rcutils/pull/53/files#diff-f909d0b1a298531f76132943dc55f86aR251) it's to put a conservative limit on the length of an individual token (not its expansion).

To your point about limiting the length of the environment variable, since the contents of the environment variable [needs to be copied in](https://github.com/ros2/rcutils/pull/53/files#diff-f909d0b1a298531f76132943dc55f86aR53), I set a limit so we could declare the buffer statically to avoid the need to dynamically allocate in `rcutils_logging_initialize`, since there's no logging fini at the moment. Provided this reasoning makes sense, how about I just increase the limit to 2048 to match the [max returned env var length on windows](https://github.com/ros2/rcutils/blob/450002515fe45a64e57e90257fc2f9ac68c1f6de/src/get_env.c#L26)?

dirk-thomas: Thanks for clarifying.

> how about I just increase the limit to 2048 to match the max returned env var length on windows?

Sounds good to me.

include/rcutils/logging.h
dhood: thanks, 83a458935bae4e1dc8a4a4b1ae95943237e78f98

src/logging.c
dhood: OK, 41e93ab3ff769f4aeb42cc6d0f5364a8e3b66590

src/logging.c
dhood: It doesn't need to how the code is now, thanks! 136223ff4fc32e52ad7810d7441fb9fff0040a49

src/logging.c
dhood: TBH in general I was copying https://github.com/ros2/rcutils/pull/23/files#diff-f909d0b1a298531f76132943dc55f86aR124 and went a bit off-script. In cd67099b586950a6e32544d6e122bae65b2e2a6c I removed this cast to match the line I just linked to, but I'm also happy to update all of them to not use `void *`

src/logging.c
dhood: Yep, I forgot that we prefer that f07539581e94dfb32c02c6e849fc44c30e7e3ca8

src/logging.c
dhood: 3ae0a9b9b473f694252d30701b82d1bb2f971d45 and optionally 1705ae1ead44352790aed2dd701dd0cd67c988b9 are the commits that remove the use of memset in favour of manually copying null terminators into `output_buffer`

src/logging.c
dhood: the truncation is handled by `rcutils_snprintf` and FWIU we'll need to use that function regardless. Because of that, while I like the idea of determining the max number of digits, I don't think it will gain us any simplification in the code, so to me it doesn't seem it'd be worth the calculation. Would you agree?

dirk-thomas: Sure, good enough as-is. If someone comes with a source file with more lines we question if there is any value in giving him the right line number since his editor might not be able to load that file anyway :wink:

dhood: sounds like a plan!

dhood:
Since this PR adds a new test, these jobs run it 100 times to check it's not flaky:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3312/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/577/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2642/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3350/)

And this one runs it once on Windows debug since that catches issues sometimes: [Build Status](http://ci.ros2.org/job/ci_windows/3351/)

dhood:
@ros2/team another set of eyes on this would be appreciated

src/logging.c
```diff
@@ -104,47 +172,153 @@ void rcutils_logging_console_output_handler(
...
+        1, output_buffer_size, allocator, output_buffer, static_output_buffer)
+      memcpy(output_buffer + old_output_buffer_len, str + i, 1);
+      output_buffer[old_output_buffer_len + 1] = '\0';
+      i++;
+      continue;
+    }
+    size_t n = strlen(token_expansion);
+    RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER(
+      n, output_buffer_size, allocator, output_buffer, static_output_buffer)
+    memcpy(output_buffer + old_output_buffer_len, token_expansion, n + 1);
```
wjwwood: Looking at this, I wonder if you might be better off using `repl_str()`: https://github.com/ros2/rcutils/blob/6d6d5507362415fd377098c0b5973b91072dfafa/include/rcutils/repl_str.h#L127-L132

Did you consider it? Was there some reason to not use it for replacing tokens like `{name}` with an actual value? (there very well might have been)

test/test_logging_output_format.py
```diff
@@ -0,0 +1,75 @@
...
+    handler = create_handler(name, launch_descriptor, output_file)
+    assert handler, 'Cannot find appropriate handler for %s' % output_file
+    launch_descriptor.add_process(
+        cmd=[executable],
+        env=env_edge_cases,
+        name=name,
+        exit_handler=ignore_exit_handler,
+        output_handlers=[ConsoleOutput(), handler],
+    )
+
```
clalancette: I suggest adding a test here that has no tokens whatsoever, just to make sure that piece works.

dhood: sure thing! ce21158d6e5a31efc653c2e9abb1bfb922c56a8b

src/logging.c
dhood: TBH I hadn't seen it, but I've taken a look and don't think we'd want to pull it in "as is". We'd need at least 6 calls to it, and each allocates memory, while we had been trying to avoid allocation if possible in the logging calls themselves. Certainly it'd be possible to have a version of `rcutils_repl_str` that only allocates when necessary, as with `RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER`, and if/when we have another instance of needing a `rcutils_repl_str` that only allocates when necessary, I think that's what we should consider switching to

src/logging.c
```diff
@@ -72,6 +103,43 @@ void rcutils_log(
...
+      void * dynamic_output_buffer = allocator.allocate(output_buffer_size, allocator.state); \
+      if (!dynamic_output_buffer) { \
+        fprintf(stderr, "failed to allocate buffer for logging output\n"); \
+        goto cleanup; \
+      } \
+      memcpy(dynamic_output_buffer, output_buffer, old_output_buffer_len); \
+      output_buffer = (char *)dynamic_output_buffer; \
+    } else { \
+      void * new_dynamic_output_buffer = allocator.reallocate( \
+        output_buffer, output_buffer_size, allocator.state); \
```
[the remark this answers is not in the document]
dhood: Yep, if allocation of `new_dynamic_output_buffer` fails, `cleanup` will handle deallocation of `output_buffer` so I think we're good: https://github.com/ros2/rcutils/pull/53/files#diff-f909d0b1a298531f76132943dc55f86aR330

src/logging.c
```diff
@@ -31,6 +41,27 @@ void rcutils_logging_initialize()
...
+    // Check for the environment variable for custom output formatting
+    const char * output_format;
+    const char * ret = rcutils_get_env("RCUTILS_CONSOLE_OUTPUT_FORMAT", &output_format);
+    if (NULL == ret && strcmp(output_format, "") != 0) {
+      size_t chars_to_copy = strlen(output_format);
+      if (chars_to_copy > RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN - 1) {
+        chars_to_copy = RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN - 1;
+      }
+      memcpy(g_rcutils_logging_output_format_string, output_format, chars_to_copy);
+      g_rcutils_logging_output_format_string[chars_to_copy] = '\0';
```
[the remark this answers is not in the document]
dhood: windows complains about `strncpy` (and fwiu there isn't a standardised min available for use), so I've stuck with `memcpy` + manual truncation here

src/logging.c
```diff
@@ -72,6 +103,43 @@ void rcutils_log(
...
 }
 
+/*!
+  RCUTILS_LOGGING_ENSURE_LARGE_ENOUGH_BUFFER(
+   n, output_buffer_size, allocator, output_buffer, static_output_buffer)
+  Macro for (re)allocating a dynamic buffer if the ouput buffer doesn't have enough space for an
+  additional n characters and null terminator. Whether to allocate or re-allocate is determined by
+  if the output_buffer points to the original static_output_buffer or not. If a new buffer is
+  allocated it will be padded with NUL.
+*/
```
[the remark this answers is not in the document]
dhood: I had NUL on purpose referring to the character, but this sentence doesn't actually apply anymore anyway so I'm glad you mentioned it. df3bb88d935a75776f90f0cf13cb3fca1e806175 updates the docblock

src/logging.c
```diff
@@ -135,12 +146,64 @@ void rcutils_logging_console_output_handler(
   }
 
   if (!location) {
-    fprintf(stream, "[%s] [%s]: %s\n", severity_string, name, message_buffer);
-  } else {
-    fprintf(
-      stream, "[%s] [%s]: %s (%s() at %s:%zu)\n", severity_string, name, message_buffer,
-      location->function_name, location->file_name, location->line_number);
+    // TODO before merge
```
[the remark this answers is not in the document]
dhood: not sure why this showed up when you were reviewing but this had been resolved already :+1:

src/logging.c
```diff
@@ -17,10 +17,20 @@ extern "C"
...
+#include "rcutils/find.h"
+#include "rcutils/get_env.h"
 #include "rcutils/logging.h"
+#include "rcutils/snprintf.h"
+
+#define RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN 2048
 
 bool g_rcutils_logging_initialized = false;
+char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
+static const char * rcutils_default_output_format =
```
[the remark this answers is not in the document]
dhood: cheers bb3e6139eb0767ec6b885c0a9e51f168bc7c059d

src/logging.c
dhood: If we were to switch to a function I think the readability of the code would stay the same, wouldn't it? (we'd remove trailing `\`, use pointers for buffer sizes, but otherwise the code would stay the same). The duplication of code at each of the callers (e.g. error checking) makes me think we wouldn't gain much from switching, but I might have missed something

dhood:
More CI after recent reviews:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3328/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/590/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2656/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3368/)

src/logging.c
wjwwood: Ok, that's fine (I thought that might be the case).

Btw, is that an opportunity for optimization, could you avoid expanding tokens which do not change over time, like the name or severity? Perhaps that's too much of a premature optimization.

In the other extreme, is it that big of a deal to allocate so many times during a logging call? I mean obviously we'd love have an allocation free logging function, but that might not be realistic unless we just truncate them to a fixed length. By comparison using a std::string in normal cases might result in many separate memory allocations and that seems to work ok in most cases despite the overhead.

By the way, the "performance" section of the website where I got the `repl_str()` function is pretty interesting:

http://creativeandcritical.net/str-replace-c#performance

I think the one we have isn't great for small strings with lots of matches, but honestly it probably doesn't matter that much. When someone needs to make giga-logs per second, they can come and help us optimize it :smile:.

src/logging.c
clalancette: In general, macros are difficult to read and reason about, and difficult to debug in a debugger later.  As an instance of the difficulty of reasoning about it, the fact that the macro accesses the `size` variable, which is in the "outer scope" of the generated code, means that the reader has to go find size in the correct scope to know what is going on with it (this is similar to using global variables in some respects).  If we changed it to a function I don't think the code size would change very much, but I think it would be a lot easier to read and reason about.

All of that being said, this seems to be a common pattern in this codebase.  If others are OK with the macros, then we can go ahead with it.

src/logging.c
dhood: Yeah the logging is only guaranteed to be allocation free at the moment if the expanded output is <1024, it doesn't truncate (but did at some point in this PR). From the existing code I was under the impression that we valued only allocating when necessary, but your point is fair about `std::string`; I am probably taking it to the extreme and `repl_str` wouldn't be that bad. If I were to do it again from scratch I think it would be worth the extra `realloc`s for readability, but for now we can benefit from the slight optimisation. Thanks for pointing it out so I don't unnecessarily optimise for memory allocation in the future.

And I like the idea about saving the expanded token string.. we don't have the infrastructure for even caching the severity threshold just yet to avoid getting into this function unnecessarily, but once the logger concept is there for that then we might be able to avoid recalculating it

src/logging.c
dhood: ok, I understand your point better now when considering the reasoning about the code. I still don't know that too much would change for this particular case due to the nature of the arguments (e.g. `output_buffer_size`) being both inputs and outputs, but I'll keep your point about "reasonability" in mind for the future when making the decision for one or the other

PR #57, opened 2017-10-10
Programmatic logger severity thresholds

dhood:
~~still in progress~~ description below.

### DIFF
--- a/include/rcutils/find.h
+++ b/include/rcutils/find.h
@@ -23,9 +23,33 @@ extern "C"
 #include "rcutils/types.h"
 #include "rcutils/visibility_control.h"
 
+/// Return the first index of a character in a string.
+/**
+ * Search in a string for the first occurence of a delimiter.
+ *
+ * \param[in] str null terminated c string to search
+ * \param[in] delimiter the character to search for
+ * \returns the index of the first occurence of the delimiter if found, or
+ * \returns string_length if the delimiter is not found
+ */
 RCUTILS_PUBLIC
 size_t
 rcutils_find(const char * str, char delimiter);
+
+/// Return the first index of a character in a string of specifed length.
+/**
+ * Identical to rcutils_find_first() but without relying on the string to be a
+ * null terminated c string.
+ *
+ * \param[in] str string to search
+ * \param[in] delimiter the character to search for
+ * \param[in] string_length length of the string to search
+ * \returns the index of the first occurence of the delimiter if found, or
+ * \returns string_length if the delimiter is not found
+ */
+RCUTILS_PUBLIC
+size_t
+rcutils_findn(const char * str, char delimiter, size_t string_length);
 
 /// Return the last index of a character in a string.
 /**
--- a/include/rcutils/find.h
+++ b/include/rcutils/find.h
@@ -37,7 +37,7 @@ RCUTILS_PUBLIC
 size_t
 rcutils_find(const char * str, char delimiter);
 
-/// Return the first index of a character in a string of specifed length.
+/// Return the first index of a character in a string of specified length.
 /**
  * Identical to rcutils_find_first() but without relying on the string to be a
  * null terminated c string.
--- a/include/rcutils/find.h
+++ b/include/rcutils/find.h
@@ -30,7 +30,8 @@ extern "C"
  * \param[in] str null terminated c string to search
  * \param[in] delimiter the character to search for
  * \returns the index of the first occurence of the delimiter if found, or
- * \returns string_length if the delimiter is not found
+ * \returns `string_length` if the delimiter is not found, or
+ * \returns `0` for invalid arguments
  */
 RCUTILS_PUBLIC
 size_t
@@ -45,7 +46,8 @@ rcutils_find(const char * str, char delimiter);
  * \param[in] delimiter the character to search for
  * \param[in] string_length length of the string to search
  * \returns the index of the first occurence of the delimiter if found, or
- * \returns string_length if the delimiter is not found
+ * \returns `string_length` if the delimiter is not found, or
+ * \returns `0` for invalid arguments
  */
 RCUTILS_PUBLIC
 size_t
@@ -58,7 +60,8 @@ rcutils_findn(const char * str, char delimiter, size_t string_length);
  * \param[in] str null terminated c string to search
  * \param[in] delimiter the character to search for
  * \returns the index of the last occurence of the delimiter if found, or
- * \returns string_length if the delimiter is not found
+ * \returns `string_length` if the delimiter is not found, or
+ * \returns `0` for invalid arguments
  */
 RCUTILS_PUBLIC
 size_t
@@ -73,7 +76,8 @@ rcutils_find_last(const char * str, char delimiter);
  * \param[in] delimiter the character to search for
  * \param[in] string_length length of the string to search
  * \returns the index of the last occurence of the delimiter if found, or
- * \returns string_length if the delimiter is not found
+ * \returns `string_length` if the delimiter is not found, or
+ * \returns `0` for invalid arguments
  */
 RCUTILS_PUBLIC
 size_t
--- a/include/rcutils/find.h
+++ b/include/rcutils/find.h
@@ -27,9 +27,33 @@ RCUTILS_PUBLIC
 size_t
 rcutils_find(const char * str, char delimiter);
 
+/// Return the last index of a character in a string.
+/**
+ * Search in a string for the last occurence of a delimiter.
+ *
+ * \param[in] str null terminated c string to search
+ * \param[in] delimiter the character to search for
+ * \returns the index of the last occurence of the delimiter if found, or
+ * \returns string_length if the delimiter is not found
+ */
 RCUTILS_PUBLIC
 size_t
 rcutils_find_last(const char * str, char delimiter);
+
+/// Return the last index of a character in a string of specifed length.
+/**
+ * Identical to rcutils_find_last() but without relying on the string to be a
+ * null terminated c string.
+ *
+ * \param[in] str string to search
+ * \param[in] delimiter the character to search for
+ * \param[in] string_length length of the string to search
+ * \returns the index of the last occurence of the delimiter if found, or
+ * \returns string_length if the delimiter is not found
+ */
+RCUTILS_PUBLIC
+size_t
+rcutils_find_lastn(const char * str, char delimiter, size_t string_length);
 
 #if __cplusplus
 }
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -21,6 +21,8 @@
 
 #include "rcutils/visibility_control.h"
 
+#define RCUTILS_LOGGING_ROOT_LOGGER_NAME ""
+
 #if __cplusplus
 extern "C"
 {
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -236,7 +236,8 @@ int rcutils_logging_get_logger_severity_thresholdn(const char * name, size_t nam
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | No, for logger that has already had its threshold set
- *                    | Yes, for logger with previously unset threshold
+ *                    | Yes, for logger with previously unset threshold, if more
+ *                    | storage is required
  * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
@@ -257,7 +258,7 @@ void rcutils_logging_set_logger_severity_threshold(const char * name, int severi
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param The name of the logger, must be null terminated c string.
+ * \param The name of the logger, must be null terminated c string or NULL.
  * \param The severity level.
  *
  * \return True if the logger is enabled for the severity; false otherwise.
@@ -300,9 +301,9 @@ int rcutils_logging_get_logger_effective_threshold(const char * name);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param location The pointer to the location struct
+ * \param location The pointer to the location struct or NULL
  * \param severity The severity level
- * \param name The name of the logger
+ * \param name The name of the logger, must be null terminated c string or NULL
  * \param format The format string
  * \param ... The variable arguments
  */
@@ -333,7 +334,7 @@ void rcutils_log(
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param location The pointer to the location struct
+ * \param location The pointer to the location struct or NULL
  * \param severity The severity level
  * \param name The name of the logger, must be null terminated c string
  * \param format The format string for the message contents
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -187,7 +187,7 @@ void rcutils_logging_set_severity_threshold(int severity);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param name The name of the logger
+ * \param name The name of the logger, must be null terminated c string
  * \return The severity threshold if it has been set, or
  * \return `RCUTILS_LOG_SEVERITY_UNSET` if unset
  */
@@ -207,10 +207,10 @@ int rcutils_logging_get_logger_severity_threshold(const char * name);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param name The name of the logger.
- * \param name_length Logger name length.
+ * \param name The name of the logger
+ * \param name_length Logger name length
  * \return The severity threshold if it has been set, or
- * \return `RCUTILS_LOG_SEVERITY_UNSET` if unset.
+ * \return `RCUTILS_LOG_SEVERITY_UNSET` if unset
  */
 RCUTILS_PUBLIC
 int rcutils_logging_get_logger_severity_thresholdn(const char * name, size_t name_length);
@@ -220,12 +220,13 @@ int rcutils_logging_get_logger_severity_thresholdn(const char * name, size_t nam
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
- * Allocates Memory   | Yes
+ * Allocates Memory   | No, for logger that has already had its threshold set
+ *                    | Yes, for logger with previously unset threshold
  * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param name The name of the logger.
+ * \param name The name of the logger, must be null terminated c string.
  * \param severity The severity threshold to be used.
  */
 RCUTILS_PUBLIC
@@ -241,7 +242,7 @@ void rcutils_logging_set_logger_severity_threshold(const char * name, int severi
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param The name of the logger.
+ * \param The name of the logger, must be null terminated c string.
  * \param The severity level.
  *
  * \return True if the logger is enabled for the severity; false otherwise.
@@ -264,7 +265,7 @@ bool rcutils_logging_is_enabled_for(const char * name, int severity);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param The name of the logger.
+ * \param The name of the logger, must be null terminated c string.
  *
  * \return The severity threshold.
  */
@@ -319,7 +320,7 @@ void rcutils_log(
  *
  * \param location The pointer to the location struct
  * \param severity The severity level
- * \param name The name of the logger
+ * \param name The name of the logger, must be null terminated c string
  * \param format The format string for the message contents
  * \param args The variable argument list for the message format string
  */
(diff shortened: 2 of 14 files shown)


dhood:
OK, this is ready for some input.

This PR does the following:
1. Changes the notion of "global severity threshold" to "default severity threshold".
1. Adds the ability to specify the severity threshold of named loggers, with a hierarchy similar to log4cxx: `x.y.z` has severity of `x.y.z` if set, otherwise `x.y`, otherwise `x`, otherwise default.
    1. This differs from log4cxx in that we fallback to a default instead of attaching all loggers to a root logger. I did this to avoid having the root logger name, e.g. `rcutils`/empty string, prepended to all logger names. There is no prepending of any names in this PR.
1. Unnamed loggers will always use the default threshold.
    1. This includes logging macros that don't specify a name e.g. `RCUTILS_LOG_INFO`, which now pass `NULL` instead of `""` so we can distinguish them from having an empty name. This distinction probably isn't necessary, I just wanted to avoid the surprise when someone changes the severity of name `""` and it affects logging calls without names.
    1. I envision that rcl/rclcpp's `ROS_INFO` (or whatever) will call through to `RCUTILS_LOG_INFO_NAMED` with a name that prepends something like `ros.<package_name>`, so only people using `rcutils` directly will use the unnamed logging macros anyway.
1. Uses a string-to-string map for the severity levels instead of a string-to-int map. That's what was available; we can swap it out later if the performance is of concern.
1. Currently even if `rcutils_logging_initialize` hasn't been explicitly called, or it has been called but it had issues, the system can still continue to log with default settings. I've taken this approach because most of the logging will be called through macros that don't check the return code of `rcutils_logging_init`. If appropriate, we can remove the `RCUTILS_LOGGING_AUTOINIT` calls in the configuration functions and make them return errors/dummy values if logging is unititialised/there were errors during initialisation.

There were a couple of additions of other rcutils common functions that can be pulled out of this PR if they clutter the diff.

Out of the scope of this PR:
1. Caching logger levels to avoid [checking if they're enabled each call](https://github.com/ros2/rcutils/pull/57/files#diff-16a8c3749d32285b558137f65c5c60d2R56).
1. Sending notifications when logger severity thresholds change.
1. Intelligently generating default names for loggers (e.g. `ros.<package_name>` or similar).
1. Storing severities in a string-int map instead of string-string map.


CI including rclpy minimally updated to API changes (no functional changes to rclpy logging):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/599/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2670/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3396/)

include/rcutils/logging.h
```diff
@@ -56,14 +117,15 @@ typedef struct rcutils_log_location_t
...
-/// The severity levels of log message.
+/// The severity levels of log messages / logger thresholds.
 enum RCUTILS_LOG_SEVERITY
 {
   RCUTILS_LOG_SEVERITY_DEBUG = 0,  ///< The debug log level
   RCUTILS_LOG_SEVERITY_INFO = 1,  ///< The info log level
   RCUTILS_LOG_SEVERITY_WARN = 2,  ///< The warn log level
   RCUTILS_LOG_SEVERITY_ERROR = 3,  ///< The error log level
   RCUTILS_LOG_SEVERITY_FATAL = 4,  ///< The fatal log level
+  RCUTILS_LOG_SEVERITY_UNSET = 100,  ///< The unset log level
```
dirk-thomas: API not specifically checking for this would consider it to be higher than fatal which would hide all further log messages. Therefore I would prefer the unset value to be *lower* than all real severity values.

include/rcutils/logging.h
```diff
@@ -116,45 +178,153 @@ rcutils_logging_output_handler_t rcutils_logging_get_output_handler();
...
 
-/// The global severity threshold before calling the output handler.
+/// The default severity threshold for log calls.
 /**
- * The global severity threshold is being checked after the conditions when
- * using the various logging macros.
- *
- * \param severity The global severity threshold to be used.
+ * This severity threshold is used for nameless log calls and loggers with
+ * effective severity thresholds that are otherwise unspecified.
```
dirk-thomas: Doesn't this also serve as the fallback when no specific level for a given name is specified?

resource/logging_macros.h.em
```diff
@@ -46,13 +46,14 @@ extern "C"
  * \param condition_after The condition macro(s) inserted after the log call
  * \param name The name of the logger
  * \param ... The format string, followed by the variable arguments for the format string
+ * // TODO(dhood): optimise severity check via notifyLoggerLevelsChanged concept or similar.
```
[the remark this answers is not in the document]
dhood: yeah, that would be weird for people reading docs, hey: f12655b3afdee6e0b4b88e05e367ed2d49d029fc

src/logging.c
```diff
@@ -24,81 +24,260 @@ extern "C"
...
+{
+  if (!g_rcutils_logging_initialized) {
+    return RCUTILS_RET_OK;
+  }
+  rcutils_ret_t ret = RCUTILS_RET_OK;
+  if (g_rcutils_logging_severities_map_valid) {
+    rcutils_ret_t ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
+    if (ret != RCUTILS_RET_OK) {
+      fprintf(stderr, "Failed to finalize logging severities map: return code %d", ret);
+      g_rcutils_logging_severities_map_valid = false;
```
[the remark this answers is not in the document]
dhood: absolutely it should, thank you: 60b108f22f8bbd408e90f8acee21a1c04dfb9219

test/test_logging.cpp
```diff
@@ -28,9 +28,9 @@
 
 TEST(CLASSNAME(TestLogging, RMW_IMPLEMENTATION), test_logging_initialization) {
   EXPECT_FALSE(g_rcutils_logging_initialized);
-  rcutils_logging_initialize();
+  ASSERT_EQ(rcutils_logging_initialize(), RCUTILS_RET_OK);
   EXPECT_TRUE(g_rcutils_logging_initialized);
-  rcutils_logging_initialize();
+  ASSERT_EQ(rcutils_logging_initialize(), RCUTILS_RET_OK);
```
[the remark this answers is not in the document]
dhood: 43966e3005f2376904f6ee25806418eb200d0bc5

include/rcutils/logging.h
```diff
@@ -116,45 +178,153 @@ rcutils_logging_output_handler_t rcutils_logging_get_output_handler();
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
  * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
+ * \param name The name of the logger, must be null terminated c string.
  * \param severity The severity threshold to be used.
+ * \return `RCL_RET_OK` if successful, or
```
mikaelarguedas: RCUTILS_RET_OK ?

test/test_string_map.cpp
```diff
@@ -869,6 +869,95 @@ TEST(test_string_map, set) {
   }
 }
 
+TEST(test_string_map, key_exists) {
```
dhood: to the reviewer: `key_exists(n)` was added in order to distinguish between failed key retrieval and a key not existing: https://github.com/ros2/rcutils/pull/57/commits/70e7a2c9076f22a16c0d94f4bebe0de5cc1ab27f#diff-f909d0b1a298531f76132943dc55f86aL167

include/rcutils/types/string_map.h
```diff
@@ -273,6 +273,47 @@ RCUTILS_WARN_UNUSED
...
+ * or the string_map is NULL or invalid, or if the key is NULL.
+ * In all cases no error message is set.
+ *
+ * \param[in] string_map rcutils_string_map_t to be searched
+ * \param[in] key map key, must be null terminated c string
+ * \return `true` if key is in the map, or
+ * \return `false` if key is not in the map, or
+ * \return `false` for invalid arguments, or
+ * \return `false` if the string map is invalid, or
+ * \return `false` if an unknown error occurs
```
mikaelarguedas: if there are many different cases for this to return false, should we provide a way to know the reason of the failure ? (same below)

include/rcutils/logging.h
dhood: correct, that's what I meant by "loggers with effective severity thresholds that are otherwise unspecified". I'll re-word it.

include/rcutils/logging.h
dhood: OK, I will move it to be 0 and bump the other levels higher. I will copy python's logger levels https://docs.python.org/3/library/logging.html#logging-levels and set them with gaps of 10, and update rclpy to match.

src/logging.c
```diff
@@ -24,81 +24,260 @@ extern "C"
...
+  if (RCUTILS_LOG_SEVERITY_DEBUG == severity) {
+    severity_string = "DEBUG";
+  } else if (RCUTILS_LOG_SEVERITY_INFO == severity) {
+    severity_string = "INFO";
+  } else if (RCUTILS_LOG_SEVERITY_WARN == severity) {
+    severity_string = "WARN";
+  } else if (RCUTILS_LOG_SEVERITY_ERROR == severity) {
+    severity_string = "ERROR";
+  } else if (RCUTILS_LOG_SEVERITY_FATAL == severity) {
+    severity_string = "FATAL";
```
mikaelarguedas: This mapping seems useful and is already defined later in this file, should we put it somewhere for reuse?

test/test_logging.cpp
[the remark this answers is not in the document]
mikaelarguedas: The referenced commit didnt modify these lines (maybe its the github UI not updating though)

src/find.c
```diff
@@ -30,14 +30,22 @@ rcutils_find(const char * str, char delimiter)
...
     return 0;
   }
+  return rcutils_findn(str, delimiter, strlen(str));
+}
+
+size_t
+rcutils_findn(const char * str, char delimiter, size_t string_length)
+{
+  if (!str) {
+    return 0;
```
clalancette: I know you didn't come up with this to start with, but I think this is a bug.  Consider the case where the delimiter is the zero'th character.  In that case, the caller can't tell if they got a zero back because the string was NULL, or because the delimiter was in the zeroth position.  I think this should probably return an `ssize_t`, and return a negative value in the `!str` case.

test/test_logging.cpp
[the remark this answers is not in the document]
dhood: nah that was me :) b6f2cdaf661d761b06bcc683a0dad731d6e63e1e

include/rcutils/logging.h
dhood: bf54dd61ef30dc11caaeaceeb8008c3f6de29148

include/rcutils/logging.h
dhood: hopefully https://github.com/ros2/rcutils/pull/57/commits/6719f5da54f0e40e4c0fcd4ebf7a68044b2b5dec#diff-f1d39d4e841937fde043a8e49b7e6c2aR183 makes this clearer

include/rcutils/logging.h
dhood: oops! https://github.com/ros2/rcutils/pull/57/commits/6719f5da54f0e40e4c0fcd4ebf7a68044b2b5dec#diff-f1d39d4e841937fde043a8e49b7e6c2aR278

include/rcutils/types/string_map.h
dhood: The unknown error case doesn't actually apply here, so really they're all just from users passing invalid arguments, and I suppose (base off of https://github.com/ros2/rcutils/blob/79b93c20462b0e5ec766cd1807b0643bf25b9824/include/rcutils/types/string_map.h#L314) we prefer the cleaner interface when there aren't many internal failure modes. I removed the "unknown error" return case in 27bde82868c060878390e835d0754f8504d33d13 and think that should be sufficient (until the function grows in complexity underneath)

src/find.c
dhood: We consider that an invalid argument (or at least we do in the string_map https://github.com/ros2/rcutils/blob/42e7840defd2e1150df2f08eb802a4708283376c/src/string_map.c#L238), so if they specify invalid arguments then I think it's ok if they have ambiguous results (if it were an error that happened on our part then I agree), do you think that's fair? I clarified that it'll return 0 on `!str` in the docs though because it was missing: 8aa2d6c1f3bd214a0f0e10475b0c3912fb938298

src/logging.c
dhood: OK, 07f733c167f6aebfb0dc9794184b6dc0542d3d6c adds a mapping and uses it for the threshold settings and 939a4352cde7d475f40a978af5476ac745f43c7f uses it for the console outputting

dhood:
Latest CI including https://github.com/ros2/rclpy/pull/130:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3358/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/616/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2694/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3418/)

src/find.c
sloretz: It would be a very tiny performance improvement if this returned `SIZE_MAX` on invalid argument or the delimiter not being found. Calling code could have just one comparison against `SIZE_MAX` if it didn't need the string for anything else. Currently calling code needs to check if the string is valid, then get the length of the string, then after calling this function it must check if the return matches the length of the string it calculated.

I think it makes sense semantically for invalid arguments to return the same thing as a string not being found. It's like saying it didn't find the delimiter because there was no place to look.

clalancette: @sloretz That's a good idea, I like it.  It doesn't change the return type, and it does make semantic sense.  It would require a survey of the existing callers to make sure that they do the correct thing on the changed return value, but I think that would be better.  @dhood Do you agree?  If so, I'd say that that is a big enough change that we should do it in a different PR, and I'll open up an issue about it so that we can move forward on this one (and not forget about it in the future).

dhood: seems reasonable to me, but as you noted I didn't write this originally, so it might have been returning `size` for a particular reason. I'm sure we can adapt regardless

dhood:
this is ready for another round of review @ros2/team. everything in my previous summary is still valid (scope, points potentially up for discussion)

src/logging.c
```diff
@@ -24,81 +24,271 @@ extern "C"
...
 #define RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN 2048
 
+const char * g_rcutils_log_severity_names[] = {
+  [RCUTILS_LOG_SEVERITY_UNSET] = "UNSET",
+  [RCUTILS_LOG_SEVERITY_DEBUG] = "DEBUG",
+  [RCUTILS_LOG_SEVERITY_INFO] = "INFO",
+  [RCUTILS_LOG_SEVERITY_WARN] = "WARN",
+  [RCUTILS_LOG_SEVERITY_ERROR] = "ERROR",
+  [RCUTILS_LOG_SEVERITY_FATAL] = "FATAL",
+};
```
dirk-thomas: How are the other entries initialized?

src/logging.c
```diff
@@ -24,81 +24,271 @@ extern "C"
...
 {
+  RCUTILS_LOGGING_AUTOINIT
   return g_rcutils_logging_output_handler;
 }
 
 void rcutils_logging_set_output_handler(rcutils_logging_output_handler_t function)
 {
-  g_rcutils_logging_output_handler = function;
+  RCUTILS_LOGGING_AUTOINIT
+    g_rcutils_logging_output_handler = function;
```
dirk-thomas: Indentation seems to be off?

Same below.

src/logging.c
dhood: automatically as null, there's a check for null after retrieving the values at: https://github.com/ros2/rcutils/pull/57/files/ab249daaef9556c9aa2a2b79bc0ccb4a98829b61#diff-f909d0b1a298531f76132943dc55f86aR190

src/logging.c
dhood: yeah this is what uncrustify wanted: I'll disable what's making it complain

dhood: a78c8862ff06e7d75a5eac9c4cb6f92eb92d89cb

dhood:
Ready for review. What's changed since the last summary is that now the empty-named logger acts as if it's the root logger: changing the severity threshold for the empty-named logger changes the default severity. This was done in order to facilitate an empty root logger in https://github.com/ros2/rclpy/pull/132#discussion_r146419550. The default logger's severity is still stored as a global variable, instead of in the severity map, to facilitate global configuration even when the severity map can't be used for some reason.

include/rcutils/logging.h
```diff
@@ -116,45 +181,163 @@ rcutils_logging_output_handler_t rcutils_logging_get_output_handler();
...
+ * ------------------ | -------------
+ * Allocates Memory   | No, provided logging system is already initialized
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param name The name of the logger, must be null terminated c string or NULL.
+ * \param severity The severity level.
+ *
+ * \return True if the logger is enabled for the severity; false otherwise.
```
wjwwood: nitpick: `True` is uppercase and `false` is lower, lower is correct in C++

src/logging.c
```diff
@@ -24,81 +24,294 @@ extern "C"
...
+        ret = RCUTILS_RET_INVALID_ARGUMENT;
       }
       memcpy(g_rcutils_logging_output_format_string, g_rcutils_logging_default_output_format,
         strlen(g_rcutils_logging_default_output_format) + 1);
     }
+
+    if (!rcutils_allocator_is_valid(&allocator)) {
+      fprintf(
+        stderr,
+        "Provided allocator is invalid. Using the default allocator.\n");
```
wjwwood: Since you're returning an error code should you use `RCUTILS_SET_ERROR_MSG`?

wjwwood: Same below.

include/rcutils/find.h
```diff
@@ -23,14 +23,66 @@ extern "C"
...
 
+/// Return the first index of a character in a string.
+/**
+ * Search in a string for the first occurence of a delimiter.
+ *
+ * \param[in] str null terminated c string to search
+ * \param[in] delimiter the character to search for
+ * \returns the index of the first occurence of the delimiter if found, or
+ * \returns `string_length` if the delimiter is not found, or
+ * \returns `0` for invalid arguments
```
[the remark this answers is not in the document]
dhood: @clalancette raised the question in https://github.com/ros2/rcutils/pull/57#discussion_r145206340; it will be addressed as followup https://github.com/ros2/rcutils/issues/58

include/rcutils/logging.h
dhood: thanks, I was capitalising the sentence. fixed in 302466762b2c7b3419536f2196100a502a1e44c9

dhood:
@wjwwood could you please review the recent changes regarding [your suggestion](https://github.com/ros2/rcutils/pull/57#discussion_r148917234) to use `RCUTILS_SET_ERROR_MSG`:

- https://github.com/ros2/rcutils/pull/57/commits/159456221cb242a17e1f65cd851f2ab2ae28a966 adds setting of error messages where error codes are being returned (newlines removed in https://github.com/ros2/rcutils/pull/57/commits/c8e5725bc009db637d8fc3152bc80274214204fb)
- I saw a quote of your comment in https://github.com/ros2/rmw_connext/issues/265, so https://github.com/ros2/rcutils/pull/57/commits/371c8bf9a9ee0b468a402500798ea9094fe936ba logs when the error message is being overwritten (when it is detected) and 808a64b569025f4f1a2fea2d3f6fbb7610f73af9 returns an `RCUTILS_RET_MULTIPLE_ERRORS` rc in the event multiple issues occur.
    - In the case that the string_map has issues at https://github.com/ros2/rcutils/pull/57/commits/371c8bf9a9ee0b468a402500798ea9094fe936ba#diff-f909d0b1a298531f76132943dc55f86aL102, the [overwriting will be logged by rcutils](https://github.com/ros2/rcutils/blob/f9a0537e4008d9c12e32dbf51857e7a30a8f64e0/src/error_handling.c#L149) underneath us; the return code will still reflect multiple errors.

dhood:
I've moved the commits related to `RCUTILS_SET_ERROR_MSG` into a separate PR https://github.com/ros2/rcutils/pull/65 so that I can merge what has been approved here, and then we can more easily discuss the new commits.

This is CI for the approved changes here and the approved https://github.com/ros2/rclpy/pull/132. I'll merge.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3470/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/711/) (known flaky test)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2801/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3518/)

PR #59, opened 2017-10-24
Make key to feature_combinations a tuple instead of suffix

dhood:
connects to ros2/rclcpp#389

The suffix is quite specific to logging_macros.h and other consumers might want to iterate through combinations more generically

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/632/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2716/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3434/)

### DIFF
--- a/rcutils/logging.py
+++ b/rcutils/logging.py
@@ -97,47 +97,47 @@ class Feature:
 
 
 feature_combinations = OrderedDict((
-    (get_suffix_from_features([]), Feature()),
-    (get_suffix_from_features(['named']), Feature(
+    ((), Feature()),
+    (('named'), Feature(
         params=name_params,
         args=name_args)),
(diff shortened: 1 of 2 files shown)


dhood:
For a concrete example, if rclcpp wants to use the rcutils.logging module to create macros wrapping rcutils macros, without this PR it has to index this dict with the suffix that rcutils uses for rcutils' macros, which is too tightly coupled IMO. Now when iterating through the dict, rather than infer the feature combinations from their rcutils suffix, you just get the raw features as the key.

PR #61, opened 2017-10-26
pointer illegal access to out-of-scope stack pointer

gaoethan:
the following line_number_expansion is stack pointer:

`char line_number_expansion[10];
`

and it has already been out-of-scope in the following
access, which may cause unintended program behaviors:

```
size_t n = strlen(token_expansion);
...
memcpy(output_buffer + old_output_buffer_len, token_expansion, n + 1);

```


dhood:
Thanks for the fix! I updated the comments so they have more context in b4b6591b7df96fcf16abb49b755d24fe0b2891cf

PR #63, opened 2017-10-31
memory leak issues

gaoethan:
memory allocated from `rcutils_format_string()` is relatively easy
easy to forget to free and this leaks system resource. A wrapper
API added to make it shared and centralized, then fix those mem
leak issues using that API.

### DIFF
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -117,14 +117,16 @@ rcutils_set_error_state(
  * \param[in] ... Arguments for the format string.
  */
 #define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, format_string, ...) \
+do {  
   char * output_msg = NULL; \
   output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
   if (output_msg) { \
     RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
     allocator.deallocate(output_msg, allocator.state); \
   } else { \
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message"); \
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message\n"); \
   }
+} while (0)
 
 /// Return `true` if the error is set, otherwise `false`.
 RCUTILS_PUBLIC
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -107,7 +107,7 @@ rcutils_set_error_state(
 #define RCUTILS_SET_ERROR_MSG(msg, allocator) \
   rcutils_set_error_state(msg, __FILE__, __LINE__, allocator);
 
-/// Set the error message using a format stirng and format arguments.
+/// Set the error message using a format string and format arguments.
 /**
  * This function sets the error message using the given format string and
  * then frees the memory allocated during the string formatting.
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -126,7 +126,7 @@ rcutils_set_error_state(
     } else { \
       RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message\n"); \
     } \
-  } while (0)
+  } while (false)
 
 /// Return `true` if the error is set, otherwise `false`.
 RCUTILS_PUBLIC
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -107,18 +107,18 @@ rcutils_set_error_state(
 #define RCUTILS_SET_ERROR_MSG(msg, allocator) \
   rcutils_set_error_state(msg, __FILE__, __LINE__, allocator);
 
-/// Set the error msg with the format specified
+/// Set the error message using a format stirng and format arguments.
 /**
- * This function sets the error msg with the format specified and free the
- * corresponding memory allocated from rcutils_format_string_limit() or its
- * macro rcutils_format_string to avoid memory leak unconsciously.
+ * This function sets the error message using the given format string and
+ * then frees the memory allocated during the string formatting.
  *
- * \param[in] fmt_str format string or msg with a certain format string inline
- * \param[in] allocator the allocator to use for allocation
+ * \param[in] allocator The allocator to be used when allocating space for the error state.
+ * \param[in] format_string The string to be used as the format of the error message.
+ * \param[in] ... Arguments for the format string.
  */
-#define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, fmt_str, ...) \
+#define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, format_string, ...) \
   char * output_msg = NULL; \
-  output_msg = rcutils_format_string(allocator, fmt_str, __VA_ARGS__); \
+  output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
   if (output_msg) { \
     RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
     allocator.deallocate(output_msg, allocator.state); \
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -23,9 +23,9 @@ extern "C"
 #endif
 
 #include <stdbool.h>
+#include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <stddef.h>
 
 #include "rcutils/allocator.h"
 #include "rcutils/macros.h"
@@ -41,7 +41,7 @@ typedef struct rcutils_error_state_t
   rcutils_allocator_t allocator;
 } rcutils_error_state_t;
 
-#define SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
+#define RCUTILS_SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
 
 /// Copy an error state into a destination error state.
 /**
@@ -123,7 +123,7 @@ rcutils_set_error_state(
     RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
     allocator.deallocate(output_msg, allocator.state); \
   } else { \
-    SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message"); \
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message"); \
   }
 
 /// Return `true` if the error is set, otherwise `false`.
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -125,7 +125,7 @@ do {
     allocator.deallocate(output_msg, allocator.state); \
   } else { \
     RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message\n"); \
-  }
+  } \
 } while (0)
 
 /// Return `true` if the error is set, otherwise `false`.
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -117,16 +117,16 @@ rcutils_set_error_state(
  * \param[in] ... Arguments for the format string.
  */
 #define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, format_string, ...) \
-do { \
-  char * output_msg = NULL; \
-  output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
-  if (output_msg) { \
-    RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
-    allocator.deallocate(output_msg, allocator.state); \
-  } else { \
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message\n"); \
-  } \
-} while (0)
+  do { \
+    char * output_msg = NULL; \
+    output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
+    if (output_msg) { \
+      RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
+      allocator.deallocate(output_msg, allocator.state); \
+    } else { \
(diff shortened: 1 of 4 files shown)


gaoethan:
**[memory leak issues](https://github.com/ros2/rcl/pull/171)** https://github.com/ros2/rcl/pull/171

include/rcutils/error_handling.h
```diff
@@ -103,6 +103,22 @@ rcutils_set_error_state(
 #define RCUTILS_SET_ERROR_MSG(msg, allocator) \
   rcutils_set_error_state(msg, __FILE__, __LINE__, allocator);
 
+/// Set the formatted error msg and free its memory allocated
```
wjwwood: nitpick: no punctuation

include/rcutils/error_handling.h
```diff
@@ -103,6 +103,22 @@ rcutils_set_error_state(
...
+/**
+ * This function sets the error msg with the format specified and free the
```
wjwwood: This implies, at least to me, that a format string is given and extra arguments are expected to generate a string, a la `printf`, but that's no the case here. I think you mean something like "allocated string" or "non-literal" string or "string that should be deallocated after use".

include/rcutils/error_handling.h
```diff
@@ -103,6 +103,22 @@ rcutils_set_error_state(
 #define RCUTILS_SET_ERROR_MSG(msg, allocator) \
   rcutils_set_error_state(msg, __FILE__, __LINE__, allocator);
 
+/// Set the formatted error msg and free its memory allocated
+/**
+ * This function sets the error msg with the format specified and free the 
+ * corresponding memory allocated from function rcutils_format_string_limit() 
+ * or macro rcutils_format_string drived from it to avoid memory leak.
```
wjwwood: I think that should be the responsibility of the person who calls `rcutils_format_string()` or `rcutils_format_string_limit()`.

Why do we need a function here to do that. What if someone uses `snprintf` or just did `malloc` and `memcpy` to create a message string? Should those get functions too? (I would say no)

src/error_handling.c
```diff
@@ -167,6 +167,24 @@ rcutils_set_error_state(
...
+rcutils_set_formatted_error(
+  char * formatted_msg,
+  rcutils_allocator_t allocator)
+{
+  if (!formatted_msg) {
+    RCUTILS_SET_ERROR_MSG("No invalid memory allocated to format string", allocator);
+    return false;
+  }
+
+  RCUTILS_SET_ERROR_MSG(formatted_msg, allocator);
```
wjwwood: Because you implemented this as a function rather than a maco, (and then call the macro here) the error messages that use this will always get a file and line number here, rather than where it was called from. This isn't correct imo, so this would need to be implemented as a macro instead.

src/error_handling.c
```diff
@@ -167,6 +167,24 @@ rcutils_set_error_state(
...
+
+  allocator.deallocate(formatted_msg, allocator.state);
+  formatted_msg = nullptr;
```
wjwwood: Cannot use `nullptr` in C.

src/split.c
```diff
@@ -115,7 +115,7 @@ rcutils_split(
   if (rcutils_string_array_fini(string_array) != RCUTILS_RET_OK) {
     error_msg = rcutils_format_string(allocator, "FATAL: %s. Leaking memory", error_msg);
   }
-  RCUTILS_SET_ERROR_MSG(error_msg, allocator);
+  rcutils_set_formatted_error(error_msg, allocator);
```
wjwwood: This is a memory leak and should be fixed, but again, I don't think a special function is required here. Also this change will change the file and line number associated with this error state.

src/string_map.c
```diff
@@ -354,8 +354,7 @@ rcutils_string_map_unset(rcutils_string_map_t * string_map, const char * key)
   size_t key_index;
   if (!__get_index_of_key_if_exists(string_map->impl, key, strlen(key), &key_index)) {
     char * msg = rcutils_format_string(allocator, "key '%s' not found", key);
-    RCUTILS_SET_ERROR_MSG(msg, allocator)
-    allocator.deallocate(msg, allocator.state);
+    rcutils_set_formatted_error(msg, allocator);
```
wjwwood: This was ok before right?

src/error_handling.c
gaoethan: exactly, it's not in C++ and I'll revise it, thanks.

src/string_map.c
gaoethan: it doesn't check whether the memory allocated to `msg` is successful or not, just image that `msg` is **NULL** or a certain invalid memory address, and then to operate at that address.

gaoethan:
@wjwwood agree with you to implement it as a macro. and let me explain why I add that new function: actually I want to directly modify those issues, but I found that these issues are not unique and similar; moreover, it's really prone to forget to free the memory from rcutils_format_string() and I hope a helper API may benefit this.

I've ever considered to add one API to address this in format_string.h to implement current API content:
check whether the memory allocated from rcutils_format_string() is available or not
do the original action RCUTILS_SET_ERROR_MSG
free the memory allocated after using
in this way it will involve cross reference btw format_string.h and error_handling.h, I think it doesn't make sense for format_string.h, but adding it in error_handling.h, it will add a helper API for the case which needs to set error requiring string in a specific format with rcutils_format_string and doesn't impact the format_string.h yet.

anyway, I think this helper API (now should be helper macro) is not mandatory, please weight it and give your comments. thanks.

src/string_map.c
wjwwood: Makes sense.

wjwwood:
I would be fine with fixing the memory leaks (or potential null access) without a new function.

Or I would be fine with a new macro to help with this more common case, if it:

- is a macro and therefore preserves the file and line number of the calling location
- it does the formatting as well as error setting

What I mean by the second point is something like this:

```c++
RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
  allocator,
  "key '%s' not found", key);
return SOME_ERROR_CODE;
```

So that it creates and uses the format string together. I would prefer that to having the user create a string, give it to the macro and let it deallocate it. I would prefer to have the same actor allocate and deallocate if possible, so either the user does both or the function does both, but not a mixture.

Instead of code that used to do something like this:

```
char * msg = rcutils_format_string(allocator, "key '%s' not found", key);
if (!msg) {
  SAFE_WRITE_STDERR("could not allocate memory for error message\n");
} else {
  RCUTILS_SET_ERROR_MSG(msg, allocator);
}
allocator.deallocate(msg, allocator.state);
return SOME_ERROR_CODE;
```

gaoethan:
@wjwwood I propose not to add `return SOME_ERROR_CODE` to the macro definition, because it changes the flow of the code using it in a sense, moreover, the return code is also different in some components. no return will make it more flexible and common to use. thanks.

wjwwood:
> @wjwwood I propose not to add return SOME_ERROR_CODE to the macro definition, because it changes the flow of the code using it in a sense, moreover, the return code is also different in some components.

Yeah, that's what I proposed too:

https://github.com/ros2/rcutils/pull/63#issuecomment-341276044
> What I mean by the second point is something like this:

```c++
RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
  allocator,
  "key '%s' not found", key);
return SOME_ERROR_CODE;
```

The return statement is outside of the macro.

include/rcutils/error_handling.h
```diff
@@ -23,6 +23,8 @@ extern "C"
 #endif
 
 #include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
 #include <stddef.h>
```
wjwwood: nitpick: please alphabetically order the includes

src/split.c
```diff
@@ -113,9 +113,8 @@ rcutils_split(
 fail:
   error_msg = "unable to allocate memory for string array data";
   if (rcutils_string_array_fini(string_array) != RCUTILS_RET_OK) {
-    error_msg = rcutils_format_string(allocator, "FATAL: %s. Leaking memory", error_msg);
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, "FATAL: %s. Leaking memory", error_msg);
   }
-  RCUTILS_SET_ERROR_MSG(error_msg, allocator);
```
wjwwood: This doesn't look right, when will line 114's message be printed? (the case where finalizing the string array succeeds)

include/rcutils/error_handling.h
```diff
@@ -39,6 +41,8 @@ typedef struct rcutils_error_state_t
   rcutils_allocator_t allocator;
 } rcutils_error_state_t;
 
+#define SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
```
wjwwood: Since this is a public header, this needs to be namespaced, e.g. `RCUTILS_SAFE_FWRITE_TO_STDERR` or something like that.

src/split.c
gaoethan: I actually keep the original without further checking :), yes, it's not reasonable and gets it tweaked, thanks.

include/rcutils/error_handling.h
```diff
@@ -103,6 +107,25 @@ rcutils_set_error_state(
...
+ */
+#define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, format_string, ...) \
+  char * output_msg = NULL; \
+  output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
+  if (output_msg) { \
+    RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
+    allocator.deallocate(output_msg, allocator.state); \
+  } else { \
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message"); \
+  }
```
wjwwood: This should probably be put into a scope, i.e. within something like `{}` or `do {} while (0)`, so that you can use this macro more than once in a function without colliding symbol definitions, e.g. `char * output_msg`.

wjwwood: Sorry I didn't see this before.

src/split.c
```diff
@@ -111,11 +112,8 @@ rcutils_split(
   return RCUTILS_RET_OK;
 
 fail:
-  error_msg = "unable to allocate memory for string array data";
-  if (rcutils_string_array_fini(string_array) != RCUTILS_RET_OK) {
-    error_msg = rcutils_format_string(allocator, "FATAL: %s. Leaking memory", error_msg);
-  }
-  RCUTILS_SET_ERROR_MSG(error_msg, allocator);
+  RCUTILS_SET_ERROR_MSG("unable to allocate memory for string array data", allocator);
+  rcutils_string_array_fini(string_array);
```
wjwwood: I don't think this is a good idea to not check the return value. If it is not `RCUTILS_RET_OK` then it will have set the error message internally, overwriting the error message set one line above.

I'd do something like this:

```c++
if (rcutils_string_array_fini(string_array) != RCUTILS_RET_OK) {
  RCUTILS_SAFE_FWRITE_TO_STDERR("failed to finalize string array during error handling: ");
  RCUTILS_SAFE_FWRITE_TO_STDERR(rcutils_get_error_string_safe());
  RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
  rcutils_reset_error();
}
RCUTILS_SET_ERROR_MSG("unable to allocate memory for string array data", allocator);
```

include/rcutils/error_handling.h
```diff
@@ -103,6 +107,25 @@ rcutils_set_error_state(
...
+ * \param[in] ... Arguments for the format string.
+ */
+#define RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, format_string, ...) \
+  char * output_msg = NULL; \
+  output_msg = rcutils_format_string(allocator, format_string, __VA_ARGS__); \
+  if (output_msg) { \
+    RCUTILS_SET_ERROR_MSG(output_msg, allocator); \
+    allocator.deallocate(output_msg, allocator.state); \
+  } else { \
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory for error message"); \
```
wjwwood: I think this string needs a terminating `\n`.

include/rcutils/error_handling.h
gaoethan: @wjwwood exactly, good catch :+1:

wjwwood:
I started some CI for this pr:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/728/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2820/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3536/)

dhood:
I added 536ae0ed355ea0f600e66db84ceb9ad9865b2741 since I've seen https://github.com/ros/ros_comm/pull/1179

wjwwood:
Thanks @dhood, needs to rebased anyways:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3500/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/729/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2828/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3549/)

mikaelarguedas:
I reapplied https://github.com/ros2/rcutils/commit/536ae0ed355ea0f600e66db84ceb9ad9865b2741 that disappeared during rebase

gaoethan:
@wjwwood @dhood @mikaelarguedas thanks for your contribution.

BTW, @wjwwood I'll run ament_uncrustify to format code, but just wanna double confirm with you for 

> at least running the test suite for the package before pushing for review.

Do you mean "ament build --build-tests" or something else ?

wjwwood:
> > at least running the test suite for the package before pushing for review.

> Do you mean "ament build --build-tests" or something else ?

I mean with `ament test` (`ament build --build-tests` only builds the tests). I usually do this:

```
ament test --skip-build --skip-install --only <package I want to test>
```

This prevents `ament test` from trying to rebuild or reinstall any of the packages being tested and therefore avoids any issues with mismatched options.

This command basically just runs CTest, so you could do that too in the build folder of the package you're working on (assuming it is a cmake package).

wjwwood:
I should have just linked to our docs: https://github.com/ros2/ros2/wiki/Ament-Tutorial#run-the-tests

wjwwood:
Ok, finally got a Windows job to finish 😄.

PR #64, opened 2017-11-07
Have rcutils_join_path return a char *.

clalancette:
It is returning malloc'ed memory, so const char * doesn't
make sense for it.

connects to ros2/ros2#423

### DIFF
--- a/include/rcutils/concat.h
+++ b/include/rcutils/concat.h
@@ -22,6 +22,15 @@ extern "C"
 
 #include "rcutils/visibility_control.h"
 
+/// Return a newly allocated string that contains lhs, followed by delimiter, followed by rhs.
+/**
+ * This function allocates memory and returns it to the caller.  It is up to the
+ * caller to release the memory once it is done with it by calling `free`.
+ *
+ * \return char * concatenated string on success
+ *         NULL on invalid arguments
+ *         NULL on failure
+ */
 RCUTILS_PUBLIC
 char *
 rcutils_concat(const char * lhs, const char * rhs, const char * delimiter);
--- a/include/rcutils/concat.h
+++ b/include/rcutils/concat.h
@@ -24,8 +24,8 @@ extern "C"
 
 /// Return a newly allocated string that contains lhs, followed by delimiter, followed by rhs.
 /**
- * This function allocates memory and returns it to the caller.  It is up to the
- * caller to release the memory once it is done with it by calling `free`.
+ * This function allocates memory and returns it to the caller.
+ * It is up to the caller to release the memory once it is done with it by calling `free`.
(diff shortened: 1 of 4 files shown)


include/rcutils/filesystem.h
```diff
@@ -114,7 +114,7 @@ rcutils_is_readable_and_writable(const char * abs_path);
  *         NULL on failure
  */
 RCUTILS_PUBLIC
-const char *
+char *
```
dirk-thomas: The docblock should mention that the caller gets ownership of the allocated memory and how to free it.

Same for the internally used function `rcutils_concat`.

clalancette: Good call, will change that.

wjwwood:
Thanks for fixing this @clalancette!

include/rcutils/filesystem.h
clalancette: I've updated the documentation for `rcutils_concat` and `rcutils_join_path` as requested.  Let me know if the style I did it in is consistent with what we are doing elsewhere in the codebase (I tried to make it do that, but there are some different styles around).

include/rcutils/concat.h
```diff
@@ -22,6 +22,15 @@ extern "C"
 
 #include "rcutils/visibility_control.h"
 
+/// Return a newly allocated string that contains lhs, followed by delimiter, followed by rhs.
+/**
+ * This function allocates memory and returns it to the caller.  It is up to the
```
dhood: each sentence should start on its own line to reduce coupling

include/rcutils/filesystem.h
```diff
@@ -105,16 +105,20 @@ RCUTILS_PUBLIC
 bool
 rcutils_is_readable_and_writable(const char * abs_path);
 
-/// Concatenate path adding the right delimiter according to the platform.
+/// Return a newly allocated string that contains left_hand_path, followed by
+/// the correct delimiter for the platform, followed by right_hand_path.
```
mikaelarguedas: Brief description should fit on a single line. The extended description should go in the docblock below

include/rcutils/concat.h
clalancette: Thanks, I've fixed this here (and in `filesystem.h`).

include/rcutils/filesystem.h
clalancette: Thanks, I've changed this to be a single line now.

clalancette:
Since this passed CI, and @wjwwood approved it, I'm going to merge.  If anybody wants me to do some additional changes to the documentation, just let me know and I'll open a separate PR for that.  Thanks!

PR #65, opened 2017-11-09
Use RCUTILS_SET_ERROR_MSG in logging config

dhood:
requires ros2/rcutils#63

currently contains commits from https://github.com/ros2/rcutils/pull/57 but they'll go away once it's merged

### DIFF
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -34,9 +34,11 @@ extern bool g_rcutils_logging_initialized;
 
 /// Initialize the logging system using the specified allocator.
 /**
- * This function will always set the internal state to initialized even if an
- * error occurs, to avoid repeated failing initialization attempts since this
- * function is called automatically from logging macros.
+ * Initialize the logging system only if it was not in an initialized state.
+ * If an invalid allocator is passed, the initialization will fail.
+ * Otherwise, this function will still set the internal state to initialized
+ * even if an error occurs, to avoid repeated failing initialization attempts
+ * since this function is called automatically from logging macros.
  * To re-attempt initialization after failure, call rcutils_logging_shutdown()
  * before re-calling this function.
  *
@@ -51,7 +53,7 @@ extern bool g_rcutils_logging_initialized;
  * \param allocator rcutils_allocator_t to be used.
  * \return `RCUTILS_RET_OK` if successful.
  * \retrun `RCUTILS_RET_INVALID_ARGUMENT` if the allocator is invalid, in which
- *   case the default allocator will be used.
+ *   case initialization will fail.
  * \return `RCUTILS_RET_INVALID_ARGUMENT` if an error occurs reading the output
  *   format from the `RCUTILS_CONSOLE_OUTPUT_FORMAT` environment variable, in
  *   which case the default format will be used.
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -58,8 +58,6 @@ extern bool g_rcutils_logging_initialized;
  * \return `RCUTILS_RET_LOGGING_SEVERITY_MAP_INVALID` if the internal logger
  *   severity map cannot be initialized, in which case logger severity
  *   thresholds will not be configurable.
- * \return `RCUTILS_RET_MULTIPLE_ERRORS` if multiple errors occur, in which
- *   case the error message will contain that of the last error to occur.
  */
 RCUTILS_PUBLIC
 rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allocator);
@@ -84,8 +82,6 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
  * \return `RCUTILS_RET_LOGGING_SEVERITY_MAP_INVALID` if the internal logger
  *   severity map cannot be initialized, in which case logger severity
  *   thresholds will not be configurable.
- * \return `RCUTILS_RET_MULTIPLE_ERRORS` if multiple errors occur, in which
- *   case the error message will contain that of the last error to occur.
  */
 RCUTILS_PUBLIC
 rcutils_ret_t rcutils_logging_initialize();
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -35,12 +35,15 @@ extern bool g_rcutils_logging_initialized;
 /// Initialize the logging system using the specified allocator.
 /**
  * Initialize the logging system only if it was not in an initialized state.
+ *
  * If an invalid allocator is passed, the initialization will fail.
  * Otherwise, this function will still set the internal state to initialized
  * even if an error occurs, to avoid repeated failing initialization attempts
  * since this function is called automatically from logging macros.
- * To re-attempt initialization after failure, call rcutils_logging_shutdown()
- * before re-calling this function.
+ * To re-attempt initialization, call rcutils_logging_shutdown() before
+ * re-calling this function.
+ *
+ * If multiple errors occur, the error code of the last error will be returned.
  *
  * <hr>
  * Attribute          | Adherence
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -58,6 +58,8 @@ extern bool g_rcutils_logging_initialized;
  * \return `RCUTILS_RET_LOGGING_SEVERITY_MAP_INVALID` if the internal logger
  *   severity map cannot be initialized, in which case logger severity
  *   thresholds will not be configurable.
+ * \return `RCUTILS_RET_MULTIPLE_ERRORS` if multiple errors occur, in which
+ *   case the error message will contain that of the last error to occur.
  */
 RCUTILS_PUBLIC
 rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allocator);
@@ -82,6 +84,8 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
  * \return `RCUTILS_RET_LOGGING_SEVERITY_MAP_INVALID` if the internal logger
  *   severity map cannot be initialized, in which case logger severity
  *   thresholds will not be configurable.
+ * \return `RCUTILS_RET_MULTIPLE_ERRORS` if multiple errors occur, in which
+ *   case the error message will contain that of the last error to occur.
  */
 RCUTILS_PUBLIC
 rcutils_ret_t rcutils_logging_initialize();
--- a/src/logging.c
+++ b/src/logging.c
@@ -137,8 +137,8 @@ rcutils_ret_t rcutils_logging_shutdown()
   }
   rcutils_ret_t ret = RCUTILS_RET_OK;
   if (g_rcutils_logging_severities_map_valid) {
-    rcutils_ret_t ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
-    if (ret != RCUTILS_RET_OK) {
+    rcutils_ret_t string_map_ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
+    if (string_map_ret != RCUTILS_RET_OK) {
       fprintf(
         stderr,
         "Failed to finalize logging severities map: %s\n", rcutils_get_error_string_safe());
@@ -302,9 +302,9 @@ rcutils_ret_t rcutils_logging_set_logger_severity_threshold(const char * name, i
       "Unable to determine severity_string for severity", g_rcutils_logging_allocator);
     return RCUTILS_RET_INVALID_ARGUMENT;
   }
-  rcutils_ret_t ret = rcutils_string_map_set(
+  rcutils_ret_t string_map_ret = rcutils_string_map_set(
     &g_rcutils_logging_severities_map, name, severity_string);
-  if (ret != RCUTILS_RET_OK) {
+  if (string_map_ret != RCUTILS_RET_OK) {
     fprintf(
       stderr,
       "Error setting severity for logger named '%s': %s\n", name, rcutils_get_error_string_safe());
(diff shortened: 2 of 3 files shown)


dhood:
Ready for review.

Follow up from [the suggestion in https://github.com/ros2/rcutils/pull/57](https://github.com/ros2/rcutils/pull/57#discussion_r148917234) to use `RCUTILS_SET_ERROR_MSG`:

- 008816659698aaeba6de68725d282f5df0cd7736 adds setting of error messages where error codes are being returned
- I saw a quote of @wjwwood's comment in https://github.com/ros2/rmw_connext/issues/265, so c996eabfe23a21cdbc0bc623ac943a6cca6d0db0 logs when the error message is being overwritten (when it is detected) and a81295a89cc0e433dc753e583440f23bcc08ff06 returns an `RCUTILS_RET_MULTIPLE_ERRORS` rc in the event multiple issues occur.
    - In the case that the string_map has issues at https://github.com/ros2/rcutils/commit/a81295a89cc0e433dc753e583440f23bcc08ff06#diff-f909d0b1a298531f76132943dc55f86aR110, the [overwriting will be logged by rcutils](https://github.com/ros2/rcutils/blob/f9a0537e4008d9c12e32dbf51857e7a30a8f64e0/src/error_handling.c#L149) underneath us; the return code will still reflect multiple errors.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3472/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/713/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3520/)

include/rcutils/logging.h
```diff
@@ -58,6 +58,8 @@ extern bool g_rcutils_logging_initialized;
  * \return `RCUTILS_RET_LOGGING_SEVERITY_MAP_INVALID` if the internal logger
  *   severity map cannot be initialized, in which case logger severity
  *   thresholds will not be configurable.
+ * \return `RCUTILS_RET_MULTIPLE_ERRORS` if multiple errors occur, in which
+ *   case the error message will contain that of the last error to occur.
```
wjwwood: Mmm, I don't think this makes sense. Consider a case where an invalid argument error also produces a bad alloc while formatting the error message for instance... It would be better to receive either ret code than this one imo. With this error code you only know that something went wrong, but not what went wrong without parsing the error string (which is slow and might change over time). Part of the point of the ret codes is so the calling code can react to the issue within the called function, but this return code doesn't give you any useful information in that respect. I'd rather have one of the errors printed to stderr, or added to the error message, and the other one returned normally (with error message set and the proper return code).

src/logging.c
```diff
@@ -63,6 +64,14 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
 {
   rcutils_ret_t ret = RCUTILS_RET_OK;
   if (!g_rcutils_logging_initialized) {
+    if (!rcutils_allocator_is_valid(&allocator)) {
+      allocator = rcutils_get_default_allocator();
+      RCUTILS_SET_ERROR_MSG(
+        "Provided allocator is invalid. Using the default allocator.", allocator);
```
wjwwood: I actually feel like this should just case the function to fail, not fallback to the default allocator. If the allocator object itself is invalid, then that is a logic error in the program.

src/logging.c
```diff
@@ -78,33 +87,40 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
       g_rcutils_logging_output_format_string[chars_to_copy] = '\0';
     } else {
       if (NULL != ret_str) {
-        fprintf(
-          stderr,
-          "Failed to get output format from env. variable: %s. Using default output format.\n",
-          ret_str);
-        ret = RCUTILS_RET_INVALID_ARGUMENT;
+        if (rcutils_error_is_set()) {
+          fprintf(stderr, "Overwriting error message: %s\n", rcutils_get_error_string_safe());
```
wjwwood: The `RCUTILS_SET_ERROR_MSG` does this already when you override an error state without clearing it.

src/logging.c
```diff
@@ -78,33 +87,40 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
-    g_rcutils_logging_allocator = allocator;
     g_rcutils_logging_severities_map = rcutils_get_zero_initialized_string_map();
     rcutils_ret_t string_map_ret = rcutils_string_map_init(
       &g_rcutils_logging_severities_map, 0, g_rcutils_logging_allocator);
     if (string_map_ret != RCUTILS_RET_OK) {
       fprintf(
         stderr,
-        "Failed to initialize map for logger severities. Severities will not be configurable.\n");
+        "Failed to initialize logging severities map: %s\n", rcutils_get_error_string_safe());
+      rcutils_reset_error();
```
wjwwood: I would incorporate this into your new error state but not print anything here. Consider it being something like this:

```python
try:
  init_map(...)
except SomeMapError as exc:
  raise MyError("could not setup logging, because: %s" % str(exc))
```

Basically you're catching the string map error, and "re throwing" you're own more applicable error.

src/logging.c
```diff
@@ -121,9 +137,14 @@ rcutils_ret_t rcutils_logging_shutdown()
...
   if (g_rcutils_logging_severities_map_valid) {
-    rcutils_ret_t ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
-    if (ret != RCUTILS_RET_OK) {
-      fprintf(stderr, "Failed to finalize logging severities map: return code %d", ret);
+    rcutils_ret_t string_map_ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
+    if (string_map_ret != RCUTILS_RET_OK) {
+      fprintf(
+        stderr,
+        "Failed to finalize logging severities map: %s\n", rcutils_get_error_string_safe());
+      rcutils_reset_error();
```
wjwwood: Same thing here, I'd prefer to integrate this error into the new error message and not print anything to stderr.

src/logging.c
```diff
@@ -267,19 +292,26 @@ rcutils_ret_t rcutils_logging_set_logger_severity_threshold(const char * name, i
...
-  if (ret != RCUTILS_RET_OK) {
-    fprintf(stderr, "Error setting severity for logger named '%s'", name);
-    return ret;
+  if (string_map_ret != RCUTILS_RET_OK) {
+    fprintf(
+      stderr,
+      "Error setting severity for logger named '%s': %s\n", name, rcutils_get_error_string_safe());
+    rcutils_reset_error();
+    RCUTILS_SET_ERROR_MSG(
+      "Error setting severity for logger", g_rcutils_logging_allocator);
```
wjwwood: Same here.

include/rcutils/logging.h
dhood: ok, I see now that "return an error return code if there are any accumulated" in https://github.com/ros2/rmw_connext/issues/265 doesn't necessarily mean it is a "special" error code. I follow your reasoning and will change it to be one of the existing codes

src/logging.c
dhood: Yeah, the "fallback" behaviour in general was put there because I was worried that the logging macros repeatedly calling `RCUTILS_LOGGING_AUTOINIT` and ignoring the return code would cause hard failures to go unnoticed. However, that argument doesn't apply here since the autoinit macro won't pass a custom allocator anyway, so I will remove this fallback to the default allocator and it is the caller's responsibility to check the return code.

src/logging.c
dhood: Yeah but it gives a help-like output suggesting you should reset the error message, so to me it seems more user friendly to do it manually where possible.

mikaelarguedas: I'm not sure I understood this bit, do you mean that places invoking `RCUTILS_SET_ERROR_MSG` should add this print when they need to reset the error message because the one printed by `RCUTILS_SET_ERROR_MSG` is not clear ?
Would making `RCUTILS_SET_ERROR_MSG` overwrite message clearer solve this without having to add printfs manually wherever we reset the error message?

dhood: > Would making RCUTILS_SET_ERROR_MSG overwrite message clearer solve this without having to add printfs manually wherever we reset the error message?

we could change it, yeah, but to me the issue is that it's a little _too_ helpful at the moment, so it would actually have to be made what can be considered _less_ clear: https://github.com/ros2/rcutils/blob/f9a0537e4008d9c12e32dbf51857e7a30a8f64e0/src/error_handling.c#L157 To me it looks like a bug if that gets output, so I avoided it since in this case overwriting the error message is intentional.

include/rcutils/logging.h
dhood: 13cebe7634c0f99f893e497385cae0572b451253 + b9d749b6b4e5bdce38ef9f376026b8d42b54c050

src/logging.c
dhood: ac66eced2f3c381a887987ddb65af2d0b71017a4 removes the fallback

src/logging.c
dhood: since the previous error now causes a return (ac66eced2f3c381a887987ddb65af2d0b71017a4), this code has gone away naturally :smile:

src/logging.c
dhood: 10373ef429e41951f09ad594b495c750a2ec1b76 does this :+1:

include/rcutils/logging.h
wjwwood: I was thinking, for that other comment, that it would return multiple return codes, maybe a list of them or something.

dhood: ah ok, thanks for clarifying (I can't find the full context of that quote of yours so I'm inventing a lot of context! )

src/logging.c
```diff
@@ -78,31 +87,27 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
       g_rcutils_logging_output_format_string[chars_to_copy] = '\0';
     } else {
       if (NULL != ret_str) {
-        fprintf(
-          stderr,
-          "Failed to get output format from env. variable: %s. Using default output format.\n",
-          ret_str);
+        RCUTILS_SET_ERROR_MSG(
+          "Failed to get output format from env. variable. Using default output format.",
```
wjwwood: It would be good if this error message could include the `ret_str` in some way.

src/logging.c
```diff
@@ -78,31 +87,27 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
-      fprintf(
-        stderr,
-        "Failed to initialize map for logger severities. Severities will not be configurable.\n");
+      // If an error message was set it will have been overwritten by rcutils_string_map_init.
+      char * msg = rcutils_format_string(
+        g_rcutils_logging_allocator,
+        "Failed to initialize map for logger severities [%s]. Severities will not be configurable.",
+        rcutils_get_error_string_safe());
+      rcutils_reset_error();
+      RCUTILS_SET_ERROR_MSG(msg, g_rcutils_logging_allocator)
```
wjwwood: What if `msg` is `NULL`?

wjwwood: Looks like this line could also set the error state: https://github.com/ros2/rcutils/blob/master/src/format_string.c#L41

src/logging.c
```diff
@@ -121,9 +126,15 @@ rcutils_ret_t rcutils_logging_shutdown()
...
-    if (ret != RCUTILS_RET_OK) {
-      fprintf(stderr, "Failed to finalize logging severities map: return code %d", ret);
+    rcutils_ret_t string_map_ret = rcutils_string_map_fini(&g_rcutils_logging_severities_map);
+    if (string_map_ret != RCUTILS_RET_OK) {
+      char * msg = rcutils_format_string(
+        g_rcutils_logging_allocator,
+        "Failed to finalize map for logger severities: %s",
+        rcutils_get_error_string_safe());
+      rcutils_reset_error();
+      RCUTILS_SET_ERROR_MSG(msg, g_rcutils_logging_allocator)
```
wjwwood: Again, what if `msg` is `NULL`?

src/logging.c
```diff
@@ -267,19 +282,26 @@ rcutils_ret_t rcutils_logging_set_logger_severity_threshold(const char * name, i
...
     &g_rcutils_logging_severities_map, name, severity_string);
-  if (ret != RCUTILS_RET_OK) {
-    fprintf(stderr, "Error setting severity for logger named '%s'", name);
-    return ret;
+  if (string_map_ret != RCUTILS_RET_OK) {
+    char * msg = rcutils_format_string(
+      g_rcutils_logging_allocator,
+      "Error setting severity for logger named '%s': %s", name, rcutils_get_error_string_safe());
+    rcutils_reset_error();
+    RCUTILS_SET_ERROR_MSG(msg, g_rcutils_logging_allocator)
```
wjwwood: Same here.

src/logging.c
dhood: I've marked this PR as requiring https://github.com/ros2/rcutils/pull/63/ and swapped to use `RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING` in https://github.com/ros2/rcutils/pull/65/commits/a088188e0ea0eb4d9bc0a57cb51342a5128fc20e which will take care of the NULL case.

I think I've missed your point regarding the format_string allocator check: I don't think that `RCUTILS_CHECK_ALLOCATOR` sets the error state or error message (`RCUTILS_CHECK_ALLOCATOR_WITH_MSG` [would though](https://github.com/ros2/rcutils/blob/00e2b9db9f4b71374f1388570be2de3edabd2e37/include/rcutils/allocator.h#L112))

src/logging.c
dhood: done in https://github.com/ros2/rcutils/pull/65/commits/2d5f39b90bd210975a626f53fae295431da97f3e

src/logging.c
wjwwood: You're right, w.r.t. the check allocator call 👍

dhood:
no worries, moving back into "in progress" in the meantime (but no work planned)

dhood:
CI now that https://github.com/ros2/rcutils/pull/63's been merged:
 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3544/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/751/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2886/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3628/)

PR #66, opened 2017-11-10
Migrate CHECK_ARGUMENT_FOR_NULL and CHECK_FOR_NULL_WITH_MSG from rcl into rcutils

tfoote:
Connects to ros2/rcl#176

### DIFF
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -89,6 +89,36 @@ void
 rcutils_set_error_state(
   const char * error_msg, const char * file, size_t line_number, rcutils_allocator_t allocator);
 
+/// Check an argument for a null value.
+/**
+ * If the argument's value is null, set the error message saying so and
+ * return the error_return_type after setting the error message.
+ *
+ * \param[in] argument The argument to test.
+ * \param[in] error_return_type The type to return in case of an error.
+ * \param[in] allocator The allocator to use if an error message needs to be allocated.
+ */
+#define RCUTILS_CHECK_ARGUMENT_FOR_NULL(argument, error_return_type, allocator) \
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(argument, #argument " argument is null", \
+    return error_return_type, allocator)
+
+/// Check an value for a null value with a user defined message and error statement.
+/**
+ * If the argument's value is null, set the error statement will be evaluated
+ * after setting the error message.
+ *
+ * \param[in] value The value to test.
+ * \param[in] msg The error message if value is false.
+ * \param[in] error_statement The statement to evaluate if value is false.
+ * \param[in] allocator The allocator to use if an error message needs to be allocated.
+ */
+#define RCUTILS_CHECK_FOR_NULL_WITH_MSG(value, msg, error_statement, allocator) \
+  if (!(value)) { \
+    RCUTILS_SET_ERROR_MSG(msg, allocator); \
+    error_statement; \
+  }
+
+
 /// Set the error message, as well as append the current file and line number.
 /**
  * If an error message was previously set, and rcutils_reset_error() was not called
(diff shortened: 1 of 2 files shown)


dhood:
hope I didn't mislead you on this in https://github.com/ros2/rcl/pull/176#issuecomment-343328932, but rcutils already has the implementation of these macros: https://github.com/ros2/rcutils/blob/3a8957008a5e05f30fe4060d8cb605cb6aceab82/src/common.h#L25
I thought your question was about if we should de-duplicate the implementation, not about if it was missing or not

tfoote:
Ahh, I didn't find those. Luckily they're the same. However those are in the source folder so aren't exposed/can't be reused. I guess I'll just move them from the old location into the headers to reuse them? Or should I keep the current location and define an implementation. Is there a specific reason that these macros are suppressed while others are in the headers?

dhood:
I think moving them into the include dir would be fine since that's how we have it for, e.g. [RCUTILS_SET_ERROR_MSG](https://github.com/ros2/rcutils/blob/master/include/rcutils/error_handling.h#L103) that [RCL_SET_ERROR_MSG forwards to](https://github.com/ros2/rcl/blob/e1bfa26cea18161693484ce2ea3d7715157fc942/rcl/include/rcl/error_handling.h#L40), but I'll let @wjwwood comment if there was a particular reason they chose to put the macros in src when adding them in https://github.com/ros2/rcutils/commit/f573bba6fa391941e10f40514690bd39e30cbfb0

mikaelarguedas:
> Is there a specific reason that these macros are suppressed while others are in the headers?

Not that I know of. My guess is that we expected this file to grow with internal common macros for rcutils that we may not want to expose to other packages. But if this one becomes useful outside I think it's fine to expose them

wjwwood:
It’s fine to move them to a public header. Just make sure they are namespaced and documented.

include/rcutils/error_handling.h
```diff
@@ -89,6 +89,35 @@ void
...
+ *
+ * \param[in] argument The argument to test.
+ * \param[in] error_return_type The type to return in case of an error.
+ * \param[in] allocator The allocator to use if an error message needs to be allocated.
+ */
+#define RCUTILS_CHECK_ARGUMENT_FOR_NULL(argument, error_return_type, allocator) \
+  RCUTILS_CHECK_FOR_NULL_WITH_MSG(argument, #argument " argument is null", \
+    return error_return_type, allocator)
+
+/// Check an value for a null value with a user defined message and error statement.
```
mikaelarguedas: I think keeping "argument" in the description here is fine, it'll just refer to the macro argument instead of the function's argument.
It'll also stay consistent with the description below that refers to it as an "argument"

include/rcutils/error_handling.h
```diff
@@ -89,6 +89,35 @@ void
...
+ * after setting the error message.
+ *
+ * \param[in] value The value to test.
+ * \param[in] msg The error message if value is false.
+ * \param[in] error_statement The statement to evaluate if value is false.
+ * \param[in] allocator The allocator to use if an error message needs to be allocated.
+ */
+#define RCUTILS_CHECK_FOR_NULL_WITH_MSG(value, msg, error_statement, allocator) \
+  if (!(value)) { \
+    RCUTILS_SET_ERROR_MSG(msg, allocator); \
```
mikaelarguedas: should the allocator be checked before calling SET_ERROR_MSG? An alternative would be to check it in `rcutils_set_error_state` itself

tfoote: There's no need to check as it's a pass by value at the moment.

At a high level do we want to be passing `allocator_t` or `allocator_t *`s The current implementation is not using the pointers. And I missed that in [your pr](https://github.com/ros2/rcl/pull/178) which assumes pointers. 

The current implementations almost always do not use the pointer. I'd originally implemented things using the pointer but switched to pass by value to homogenize with the current implementation.

tfoote: Sorry rereading this I was distracted by the other commit. We should definitely check in one of the two places. Since rcutils_set_error_state is public I think we have to check at that level. And since this function won't error with it not set it might be ok to pass through w/o checking.

mikaelarguedas: Sounds good to me I'll add a check in the rcutils function `rcutils_set_error_state`. This should not block this PR as this one just keeps the current behavior and the rcutils one will be an improvement.

PR #69, opened 2017-11-14
Undo mixing of once's and named's doc_lines

dhood:
A couple of places have been accidentally referencing once's doc line (which was actually called `name_doc_line`) e.g. http://docs.ros2.org/beta3/api/rcutils/logging__macros_8h.html#a4fb4a0978ca7725899f98d4c0b064e44. 

https://github.com/ros2/rcutils/commit/26d7748305fa1dfb134a46a37a84a0ef2c7cf304 also adds a watch file for the python module so it triggers template rebuilding (not necessary, just came across it at the same time)

linux ci [Build Status](http://ci.ros2.org/job/ci_linux/3516/)

### DIFF
--- a/rcutils/logging.py
+++ b/rcutils/logging.py
@@ -26,10 +26,11 @@ name_params = OrderedDict((
     ('name', 'The name of the logger'),
 ))
 name_args = {'name': 'name'}
+name_doc_lines = []
 once_args = {
     'condition_before': 'RCUTILS_LOG_CONDITION_ONCE_BEFORE',
     'condition_after': 'RCUTILS_LOG_CONDITION_ONCE_AFTER'}
-name_doc_lines = [
+once_doc_lines = [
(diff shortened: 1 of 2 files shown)


rcutils/logging.py
```diff
@@ -26,10 +26,11 @@
     ('name', 'The name of the logger'),
 ))
 name_args = {'name': 'name'}
+name_doc_lines = []
```
mikaelarguedas: Is the assumption here that `name` is explicit enough and doesnt need a doc string? Or will this get a message added in a follow-up PR ?

dhood: the "doc lines" are not for documenting the parameters (that's line 26), it's for what comes after "Log a message with severity DEBUG with the following conditions". name isn't a condition in that sense so it doesn't (and probably won't ever) have a doc line in that sense

PR #74, opened 2017-11-21
Update the rcutils_find* APIs to return SIZE_MAX on invalid arguments.

### LINKED ISSUE #58: find() and find_last() functions return ambiguous value
If a NULL or empty string is passed into `rcutils_find()` or `rcutils_find_last()`, they return 0.  However, it may also be the case that the delimiter being searched for is in the zero'th position, and the caller doesn't have any way to disambiguate this.  As discussed in https://github.com/ros2/rcutils/pull/57#pullrequestreview-69972479, we should probably change this so that `rcutils` and `rcutils_find_last` return something different; either a negative value (which would require us to change the return type), or the size of the string (indicating it wasn't found).  The latter will obviously only work on empty strings, not NULL arguments.  In either case, we need to audit all of the callers and make sure they do the correct thing.

clalancette:

fixes #58 

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3582/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/778/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2922/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3675/)

### DIFF
--- a/src/find.c
+++ b/src/find.c
@@ -17,6 +17,7 @@ extern "C"
 {
 #endif
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -28,7 +29,7 @@ size_t
 rcutils_find(const char * str, char delimiter)
 {
   if (!str || strlen(str) == 0) {
-    return 0;
+    return SIZE_MAX;
   }
   return rcutils_findn(str, delimiter, strlen(str));
 }
@@ -36,8 +37,8 @@ rcutils_find(const char * str, char delimiter)
 size_t
 rcutils_findn(const char * str, char delimiter, size_t string_length)
 {
-  if (!str) {
-    return 0;
+  if (!str || strlen(str) == 0) {
+    return SIZE_MAX;
   }
 
   for (size_t i = 0; i < string_length; ++i) {
@@ -52,7 +53,7 @@ size_t
 rcutils_find_last(const char * str, char delimiter)
 {
   if (!str || strlen(str) == 0) {
-    return 0;
+    return SIZE_MAX;
   }
   return rcutils_find_lastn(str, delimiter, strlen(str));
 }
@@ -60,8 +61,8 @@ rcutils_find_last(const char * str, char delimiter)
 size_t
 rcutils_find_lastn(const char * str, char delimiter, size_t string_length)
 {
-  if (!str) {
-    return 0;
+  if (!str || strlen(str) == 0) {
+    return SIZE_MAX;
   }
   size_t last_found = string_length;
   for (size_t i = 0; i < string_length; ++i) {
--- a/src/find.c
+++ b/src/find.c
@@ -37,7 +37,7 @@ rcutils_find(const char * str, char delimiter)
 size_t
 rcutils_findn(const char * str, char delimiter, size_t string_length)
 {
-  if (!str || strlen(str) == 0) {
+  if (!str || string_length == 0) {
     return SIZE_MAX;
   }
 
@@ -61,7 +61,7 @@ rcutils_find_last(const char * str, char delimiter)
 size_t
 rcutils_find_lastn(const char * str, char delimiter, size_t string_length)
 {
-  if (!str || strlen(str) == 0) {
+  if (!str || string_length == 0) {
     return SIZE_MAX;
   }
   size_t last_found = string_length;
--- a/src/find.c
+++ b/src/find.c
@@ -28,7 +28,7 @@ extern "C"
 size_t
 rcutils_find(const char * str, char delimiter)
 {
-  if (!str || strlen(str) == 0) {
+  if (!str || 0 == strlen(str)) {
     return SIZE_MAX;
   }
   return rcutils_findn(str, delimiter, strlen(str));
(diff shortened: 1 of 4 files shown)


src/find.c
```diff
@@ -28,16 +29,16 @@ size_t
...
   }
   return rcutils_findn(str, delimiter, strlen(str));
 }
 
 size_t
 rcutils_findn(const char * str, char delimiter, size_t string_length)
 {
-  if (!str) {
-    return 0;
+  if (!str || strlen(str) == 0) {
```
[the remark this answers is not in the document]
clalancette: Oh yeah, you are definitely right about that.  That should be a check if `string_length == 0`.  I'll fix that.

mikaelarguedas:
Dismissing my review, and +1ing both @dhood's review about non-null terminated strings.

> btw if we update to return SIZE_MAX whenever the delimiter is not found, usage such as

It looks to me that this returns SIZE_MAX only if the provided string is invalid. It will return the length of the string if it's not found so this code doesn't need to be updated.

But it raises the question of "Should it return SIZE_MAX in both the case where the input string is invalid and the one where the deimiter is not found?"

clalancette:
> there's inconsistency when the string is valid but empty. In most cases string_length is returned if the delimiter is not found, but in the case of "" as the input string, SIZE_MAX is being returned when the delimiter is not found. from #58 it sounds like SIZE_MAX should be returned in all cases

That was sort of intentional; I was trying to make it so that you could determine whether the inputs were valid (SIZE_MAX if not), and whether you found the delimiter (offset of the delimiter, or length of entire string if not).  Now that you mention it, though, an "empty" string could be considered a special case of "can't find delimiter".  I'll take another look and see what happens if we change to returning SIZE_MAX for can't find delimiter, in all cases.

> btw if we update to return SIZE_MAX whenever the delimiter is not found, usage such as
> 
> rcutils/src/logging.c
> 
> Line 473 in 77a7e9e
>  size_t chars_to_start_delim = rcutils_find(str + i, token_start_delimiter); 
> will have to be updated

So I actually did audit these callers.  I opted not to make any changes in both of those cases in `src/logging.c`, mostly because the case where it returns SIZE_MAX is handled already for both of them.  In particular, in both cases there is an if statement dealing with where the size is larger than what is left, and they both do the "correct" thing in that case.  Maybe I should add a comment in those places; what do you think?

clalancette:
It looks like it is reasonable to use SIZE_MAX for all cases of "not found", so I've changed to that.  Let me know what you think.

test/test_find.cpp
```diff
@@ -80,14 +82,14 @@ TEST(test_find, find) {
 }
 
 TEST(test_find, findn) {
-  size_t ret0 = test_findn("", '/', 0, 0);
-  LOG((size_t)0, ret0);
+  size_t ret0 = test_findn("", '/', 0, SIZE_MAX);
+  LOG((size_t)SIZE_MAX, ret0);
```
dhood: the size_t casts are unnecessary here and below now yeah?

dhood:
> In particular, in both cases there is an if statement dealing with where the size is larger than what is left, and they both do the "correct" thing in that case.

Yep you're right (I think I'd remembered they'd have issues if we returned -1 when it was originally suggested, but they are indeed good as is)

test/test_find.cpp
clalancette: Yeah, for the SIZE_MAX checks, that's probably right.  I'll make that fix and submit a CI just to be sure.

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/782/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2929/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3680/)

test/test_find.cpp
dhood: I wasn't aware of this but judging from [the warnings on OS X](http://ci.ros2.org/job/ci_osx/2929/warnings10Result/new/) `SIZE_MAX` may not fit into `size_t`.. this probably means we'll need to reconsider the return type of the find functions if we want to take this approach

clalancette: > I wasn't aware of this but judging from the warnings on OS X SIZE_MAX may not fit into size_t.. this probably means we'll need to reconsider the return type of the find functions if we want to take this approach

Arg, that is annoying.  I'll do a little more investigating to see what I can find here.

clalancette: I see what is going on.  The warning that clang is throwing is very misleading.  On x86_64 macOS, `size_t` is a long unsigned int, which is 64-bit unsigned.  SIZE_MAX is an unsigned long long, which is *also* 64-bit unsigned.  The problem is that the `LOG` statement is using `%zu` to print out the values, and an `unsigned long long` is not technically a `size_t`, even though they are compatible.  The solution is to make sure we still cast SIZE_MAX to a `size_t` before passing it to `LOG` (like what was there before).  I'll fix that, and put a comment in there for the next hapless soul who tries to remove it.

clalancette:
Another try on CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3592/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/785/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2932/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3683/)

clalancette:
Thanks everyone for the reviews.  I'm going to merge this now.

PR #75, opened 2017-11-25
Only evaluate logging 'function'/'expression' if severity is enabled

dhood:
While working on https://github.com/ros2/demos/pull/194 I noticed that  `RCUTILS_LOG_DEBUG_FUNCTION` evaluates the function even when the debug severity is not enabled, which is inefficient. I looked for a specific reason it might have been done this way but couldn't find one (it's been like this since it was added). This behaviour is [not the case in ROS 1](https://github.com/ros/ros_comm/blob/4383f8fad9550836137077ed1a7120e5d3e745de/tools/rosconsole/include/ros/console.h#L366): the conditions are only evaluated if the severity is enabled.

An example of a side effect of this change is that the "hit" flag for the 'once' filter will only get evaluated after the severity is enabled now. I see this as an improvement because otherwise, without this PR,  `RCUTILS_LOG_DEBUG_ONCE` would never be printed if the first time it gets hit is when the severity threshold is INFO instead of DEBUG. Similar side effects apply for `SKIPFIRST`, `THROTTLE` but I don't consider any of them points for concern. The behaviour after this PR matches what I expect if/when we add a "notify logger levels changed" concept for resetting filter states whenever severity thresholds are changed.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3598/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/789/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2936/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3688/)

### DIFF
--- a/test/test_logging_macros.cpp
+++ b/test/test_logging_macros.cpp
@@ -108,18 +108,31 @@ TEST_F(TestLoggingMacros, test_logging_expression) {
 }
 
 int g_counter = 0;
+bool g_function_called = false;
 
 bool not_divisible_by_three()
 {
+  g_function_called = true;
   return (g_counter % 3) != 0;
 }
 
 TEST_F(TestLoggingMacros, test_logging_function) {
+  // check that evaluation of a specified function does not occur if the severity is not enabled
+  g_rcutils_logging_default_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
+  for (int i : {1, 2, 3, 4, 5, 6}) {
+    g_counter = i;
+    RCUTILS_LOG_DEBUG_FUNCTION(&not_divisible_by_three, "message %d", i);
+  }
+  EXPECT_EQ(0u, g_log_calls);
+  EXPECT_FALSE(g_function_called);
+  g_rcutils_logging_default_severity_threshold = RCUTILS_LOG_SEVERITY_DEBUG;
+
   for (int i : {1, 2, 3, 4, 5, 6}) {
     g_counter = i;
     RCUTILS_LOG_INFO_FUNCTION(&not_divisible_by_three, "message %d", i);
   }
   EXPECT_EQ(4u, g_log_calls);
+  EXPECT_TRUE(g_function_called);
   EXPECT_EQ("message 5", g_last_log_event.message);
 }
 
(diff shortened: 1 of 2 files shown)


test/test_logging_macros.cpp
```diff
@@ -108,18 +108,31 @@ TEST_F(TestLoggingMacros, test_logging_expression) {
...
+bool not_divisible_by_three()
 {
+  g_function_called = true;
   return (g_counter % 3) != 0;
 }
 
 TEST_F(TestLoggingMacros, test_logging_function) {
+  // check that evaluation of a specified function does not occur if the severity is not enabled
+  g_rcutils_logging_default_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
   for (int i : {1, 2, 3, 4, 5, 6}) {
```
mikaelarguedas: do we need to make 6 calls here to confirm the function is not called a single time ? We could just do:
```c
g_counter = 1;
RCUTILS_LOG_DEBUG_FUNCTION(&not_divisible_by_three, "should not be logged");
EXPECT_EQ(0u, g_log_calls);
EXPECT_FALSE(g_function_called);
```

dhood: yep you're right: 3dcd3c2864dad70396d576415813461c760e4c0f

dhood:
This is what the doxygen looks like for a macro with a condition after https://github.com/ros2/rcutils/pull/75/commits/ad46cef69f802990069b880290614baf34600b43:

screen shot 2017-11-24 at 1 47 45 pm

Without a condition it just says e.g.: `Log a message with severity INFO.`

test/test_logging_macros.cpp
```diff
@@ -108,18 +108,28 @@ TEST_F(TestLoggingMacros, test_logging_expression) {
...
+bool not_divisible_by_three()
 {
+  g_function_called = true;
   return (g_counter % 3) != 0;
 }
 
 TEST_F(TestLoggingMacros, test_logging_function) {
+  // check that evaluation of a specified function does not occur if the severity is not enabled
+  g_rcutils_logging_default_severity_threshold = RCUTILS_LOG_SEVERITY_INFO;
+  RCUTILS_LOG_DEBUG_FUNCTION(&not_divisible_by_three, "should not be logged");
```
mikaelarguedas: We still need to set `g_counter` to 1 otherwise `not_divisible_by_three` returns `false`. The check of `g_function_called` is still valid but it makes the check for `g_log_calls` not relevant

dhood: Really just checking `g_function_called` is `false` _should_ be sufficient here, but more coverage can't hurt: 8c369f940f1ecd72212f8c943092033c13394f2c

PR #76, opened 2017-11-25
Make LOG macro behave like genuine code

yenWu:
Macro is really a code replacement. In this implementation, you may worry about
some bugs
1.
```
        // Declaration
        int func(int, int)
        LOG(expected actural);
```
2.
```
        while (lst & lst = lst->next) // Miss ';'
        LOG(expected, actural)
```

After this change, you could make LOG macro act like a genuine code and only use
once.

### DIFF
--- a/test/test_find.cpp
+++ b/test/test_find.cpp
@@ -24,7 +24,8 @@
 } while (0)
 #else
 #define LOG(X, arg) do { \
-    (void)arg; \
+    (void)(X); \
+    (void)(arg); \
 } while (0)
 #endif
 
--- a/test/test_find.cpp
+++ b/test/test_find.cpp
@@ -21,7 +21,7 @@
 #if ENABLE_LOGGING
 #define LOG(expected, actual) do { \
     printf("Expected: %zu Actual: %zu\n", expected, actual); \
-  } while (0)
+} while (0)
 #else
 #define LOG(X, arg) {}
 #endif
--- a/test/test_find.cpp
+++ b/test/test_find.cpp
@@ -19,8 +19,9 @@
 #define ENABLE_LOGGING 1
 
 #if ENABLE_LOGGING
-#define LOG(expected, actual) { \
-    printf("Expected: %zu Actual: %zu\n", expected, actual);}
+#define LOG(expected, actual) do { \
+    printf("Expected: %zu Actual: %zu\n", expected, actual); \
+  } while (0)
 #else
 #define LOG(X, arg) {}
 #endif
--- a/test/test_find.cpp
+++ b/test/test_find.cpp
@@ -23,7 +23,9 @@
     printf("Expected: %zu Actual: %zu\n", expected, actual); \
 } while (0)
 #else
-#define LOG(X, arg) {}
+#define LOG(X, arg) do { \
+    (void)arg; \
+} while (0)
 #endif
 
 size_t test_find(const char * str, char delimiter, size_t expected_pos)


clalancette:
While this is a totally fine change to make, it's not super important because this is just in a test, and that `LOG` macro isn't used anywhere outside of `test_find.c`.  I'm happy to merge it, though I'd like to wait until #74 goes in, since it may have some merge conflicts.

clalancette:
Now that #74 is merged, we can merge this.  Here's CI, just up until rcutils, for this:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3608/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/795/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2942/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3698/)

dhood:
@clalancette that CI doesn't seem to use this change

test/test_find.cpp
```diff
@@ -19,8 +19,9 @@
 #define ENABLE_LOGGING 1
 
 #if ENABLE_LOGGING
-#define LOG(expected, actual) { \
-    printf("Expected: %zu Actual: %zu\n", expected, actual);}
+#define LOG(expected, actual) do { \
+    printf("Expected: %zu Actual: %zu\n", expected, actual); \
+  } while (0)
 #else
 #define LOG(X, arg) {}
```
dhood: @yenWu the same reasoning would apply here when `ENABLE_LOGGING` is defined as 0, right?

clalancette:
D'oh, you are right.  Sorry about that.  Will kick off a new one.

test/test_find.cpp
```diff
@@ -23,7 +23,9 @@
     printf("Expected: %zu Actual: %zu\n", expected, actual); \
 } while (0)
 #else
-#define LOG(X, arg) {}
+#define LOG(X, arg) do { \
+    (void)arg; \
```
yenWu: @clalancette , I don't figure it out why `(void) arg` appear there. In my opinion, it makes sense like that:
```
#define LOG(X, arg) do { \
        ;
    } while (0)
```

test/test_find.cpp
yenWu: @dhood ,Yes. And thank @clalancette for fixing it up.

test/test_find.cpp
yenWu: or save some space :-P
```
#define LOG(X, arg) do {;} while (0)
```

mikaelarguedas: TL;DR I think it is to avoid "unused variable warnings".

As a lot of this testing code is initializing variables without using them for anything else than the LOG call, if the LOG macro expands to nothing, the resulting code would have many unused variables. I can imagine cases where casting `X` as well would be relevant. This code just happens to always pass litterals as the first argument, but it should compile without warnings even if we pass a variable as the first argument. So I'd recommend to update this patch to void-cast X as well.

Though I didnt try to compile that branch with or without the cast so their may be another reason for this change.

clalancette: Yep, @mikaelarguedas is exactly right.  If you don't have the `(void)arg;`, then when you compile with `ENABLE_LOGGING` set to 0, you get a bunch of warnings about unused variables.  While this isn't fatal, I figured I may as well clean it up while I was in there.

clalancette:
For whatever reason, the 4 space indent is what uncrustify wants for the macros.  Thanks for the review, merging now.

test/test_find.cpp
yenWu: Thank for explanations.

dirk-thomas:
I might be missing the point. Can one you elaborate why the `do {} while (0)` is being used rather then the curly braces before? The code before should have worked for the two mentioned cases in the original description.

sloretz:
@dirk-thomas I think the two cases are examples of bugs that can happen with just braces. The do/while loop prevents the log macro from accidentally being interpreted as a function or loop body if the user forgets a semi-colon above.

dirk-thomas:
I see how the first case would be weird.

The second case seems to be "half reasonable" assuming the user want's to log a message N times using the loop. Why should we prevent this from being used like that?

sloretz:
I think the second case is an example where the user wanted to iterate to the end of the list, but accidentally logged a message for each item in the list because they forgot a semicolon on the while loop. I think it is still possible to log a message N times in this way as long as the do/while is terminated with a semicolon.

PR #80, opened 2017-12-02
Use strlen / strnlen_s in RCUTILS_SAFE_FWRITE_TO_STDERR

### LINKED ISSUE #78: RCUTILS_SAFE_FWRITE_TO_STDERR with variable length messages
If you pass a message of type `const char *` to `RCUTILS_SAFE_FWRITE_TO_STDERR` e.g. https://github.com/ros2/rcutils/blob/954ae9bb071436ad281f31c388deaf39afe0e465/src/split.c#L116

then the `sizeof(msg)` calculation in the macro gives the size of the pointer:
https://github.com/ros2/rcutils/blob/954ae9bb071436ad281f31c388deaf39afe0e465/include/rcutils/error_handling.h#L44

From what I can tell the `count` passed to `fwrite` is not a _max_ size, but _the_ size, so I think this line can cause buffer overrun if the msg is less than length 8. That doesn't happen with our existing code because they're all longer (instead our messages get truncated), but it could happen.

What's the appropriate fix here?
- continue using `sizeof(msg)` and only use this macro with static buffers
- switch to `strlen(msg)` and only use this macro will null-terminated buffers (also a risk)

dhood:
fixes #78 
Otherwise it does not work correctly for dynamic buffers

CI forthcoming, at which point I'll put this into review once I'm sure it works cross-platform

### DIFF
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -22,6 +22,8 @@ extern "C"
 {
 #endif
 
+#define __STDC_WANT_LIB_EXT1__ 1  // needed for `strnlen_s`
+#include <string.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdio.h>
@@ -41,7 +43,13 @@ typedef struct rcutils_error_state_t
   rcutils_allocator_t allocator;
 } rcutils_error_state_t;
 
-#define RCUTILS_SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
+#ifdef __STDC_LIB_EXT1__ || WIN32
+// Limit the buffer size in the `fwrite` call to give an upper bound to buffer overrun in the case
+// of non-null terminated `msg`.
+#define RCUTILS_SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), strnlen_s(msg, 4096), stderr)
+#else
+#define RCUTILS_SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), strlen(msg), stderr)
+#endif
 
 /// Copy an error state into a destination error state.
 /**
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -22,12 +22,14 @@ extern "C"
 {
 #endif
 
-#define __STDC_WANT_LIB_EXT1__ 1  // needed for `strnlen_s`
-#include <string.h>
+// Needed for `strnlen_s` from `<string.h>` for implementations that define `__STDC_LIB_EXT1__`
+#define __STDC_WANT_LIB_EXT1__ 1
+
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


include/rcutils/error_handling.h
```diff
@@ -41,7 +43,13 @@ typedef struct rcutils_error_state_t
   rcutils_allocator_t allocator;
 } rcutils_error_state_t;
 
-#define RCUTILS_SAFE_FWRITE_TO_STDERR(msg) fwrite(msg, sizeof(char), sizeof(msg), stderr)
+#ifdef __STDC_LIB_EXT1__ || WIN32
```
mikaelarguedas: please use `_WIN32` instead of `WIN32` like the rest of the codebase

mikaelarguedas: Does this comparison ever evaluate to false ?

As it stands `__STDC_LIB_EXT1__` is always defined isn't it ? or is `string.h` undefining it ?

include/rcutils/error_handling.h
```diff
@@ -22,6 +22,8 @@ extern "C"
 {
 #endif
 
+#define __STDC_WANT_LIB_EXT1__ 1  // needed for `strnlen_s`
+#include <string.h>
```
wjwwood: nitpick: alphabetical ordering

include/rcutils/error_handling.h
wjwwood: Note: @dhood is defining `__STDC_WANT_LIB_EXT1__` above not `__STDC_LIB_EXT1__`, so my guess is `string.h` might set it.

mikaelarguedas: I see, it may deserve a comment to explain what's going on here

mikaelarguedas: I didnt know that modern GCC now accept boolean conditions on `ifdef` I remember always having to use `#if defined` for this kind of operation

wjwwood: Yeah, that is odd, I wouldn't be surprised if that didn't do what your expecting.

dhood:
Regarding `strnlen_s` support:
- Visual Studio 2015 [has `strnlen_s` support](https://msdn.microsoft.com/en-us/library/z50ty2zh.aspx) by default
- Some implementations in theory support it with the `#define __STDC_WANT_LIB_EXT1__ 1` before `#include <string.h>`
- Experimentally I've checked that the versions we're using of gcc and clang don't appear to support it even with `#define __STDC_WANT_LIB_EXT1__ 1`
- Even for implementations that do support it we need to have the `__STDC_WANT_LIB_EXT1__` _before_ the `#include <string.h>` so it's complicated to reliably use the extension anyway, so I just removed it in the end

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3691/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/858/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3024/) (flaky tests)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3786/) (flaky tests)

PR #83, opened 2017-12-13
out-of-bounds read issue

gaoethan:
`g_rcutils_log_severity_names` of 51 8-byte elements at element index 51 (byte offset 408) is out of bound using index level (which evaluates to 51).

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -274,7 +274,7 @@ rcutils_ret_t rcutils_logging_set_logger_level(const char * name, int level)
   // Convert the severity value into a string for storage.
   // TODO(dhood): replace string map with int map.
   if (level < 0 ||
-    level >
+    level >=
     (int)(sizeof(g_rcutils_log_severity_names) / sizeof(g_rcutils_log_severity_names[0])))
   {
     RCUTILS_SET_ERROR_MSG(
--- a/test/test_logging.cpp
+++ b/test/test_logging.cpp
@@ -164,6 +164,9 @@ TEST(CLASSNAME(TestLogging, RMW_IMPLEMENTATION), test_logger_severities) {
     rcutils_logging_set_logger_level("rcutils_test_loggers", -1));
   ASSERT_EQ(
     RCUTILS_RET_INVALID_ARGUMENT,
+    rcutils_logging_set_logger_level("rcutils_test_loggers", 51));
+  ASSERT_EQ(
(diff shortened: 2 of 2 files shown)


gaoethan:
Actually it's more a question for me than this minor fix for the following change.
```
 enum RCUTILS_LOG_SEVERITY
  {
 -  RCUTILS_LOG_SEVERITY_DEBUG = 0,  ///< The debug log level
 -  RCUTILS_LOG_SEVERITY_INFO = 1,  ///< The info log level
 -  RCUTILS_LOG_SEVERITY_WARN = 2,  ///< The warn log level
 -  RCUTILS_LOG_SEVERITY_ERROR = 3,  ///< The error log level
 -  RCUTILS_LOG_SEVERITY_FATAL = 4,  ///< The fatal log level
 +  RCUTILS_LOG_SEVERITY_UNSET = 0,  ///< The unset log level
 +  RCUTILS_LOG_SEVERITY_DEBUG = 10,  ///< The debug log level
 +  RCUTILS_LOG_SEVERITY_INFO = 20,  ///< The info log level
 +  RCUTILS_LOG_SEVERITY_WARN = 30,  ///< The warn log level
 +  RCUTILS_LOG_SEVERITY_ERROR = 40,  ///< The error log level
 +  RCUTILS_LOG_SEVERITY_FATAL = 50,  ///< The fatal log level
  };
```
Why is it left so many **NULL pointers** instead of the original way with sequential 0,1,2,3... ? thanks.

mikaelarguedas:
> Why is it left so many NULL pointers instead of the original way with sequential 0,1,2,3... ? thanks.

Regarding the choice of the logging levels values, they have been chosen to allow users to define intermediate levels. The current values are matching other logging system like [Python logging module](https://docs.python.org/3.5/library/logging.html#levels).

AFAICT The fact that the array is using the same indices was to follow that logic and easily access custom levels in the array. Apparently there was some discussion regarding the other values initialization https://github.com/ros2/rcutils/pull/57#discussion_r146086049.

gaoethan:
@mikaelarguedas okay, thanks for your sharing :)

dhood:
thanks @gaoethan 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3788/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3124/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3874/)

PR #85, opened 2018-01-24
take a timestamp when a console logging message happens

wjwwood:
Also make it available for console formatting.


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3894/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1013/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3220/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3990/)

wjwwood:
Missed a change in `rclcpp`'s tests, new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3897/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1014/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3221/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3991/)

dhood:
For posterity I'd like to mention an alternative that I wondered about but discarded. Rather than modifying the output handler signature to always include an rcutils timestamp (some handlers might want a different type of time to rcutils system time, or not care at all), the signature could instead be left as is and any output handlers interested in time would have to call now() themselves. The issue is that if multiple handlers are attached they might get called sequentially, in which case you want them to all have access to the time of the original call, not the time that the handler got called.

Similarly, handlers that use sim time, for example, will want the sim time of the original call as well, not when the handler was called. Since rcutils doesn't "know" about sim time, I imagine the sim timestamp would be passed to the log call in the 'context' dictionary that we have previously spoken about log calls including one day.

src/logging.c
```diff
@@ -517,6 +528,24 @@ void rcutils_logging_console_output_handler(
       token_expansion = name;
     } else if (strcmp("message", token) == 0) {
       token_expansion = message_buffer;
+    } else if (strcmp("seconds", token) == 0) {
+      if (snprintf(
+          numeric_storage, sizeof(numeric_storage),
+          "%f", timestamp / 1e9) == 0)
```
dhood: Think this should check for `< 0`

wjwwood: You're right, but maybe `<= 0`? If it returns zero then it means it wrote no bytes, which also seems like an issue.

wjwwood:
> Rather than modifying the output handler signature to always include an rcutils timestamp (some handlers might want a different type of time to rcutils system time, or not care at all), the signature could instead be left as is and any output handlers interested in time would have to call now() themselves.

That's a good point, I sort of thought about that when I was deciding whether or not have a steady time stamp too. On the one hand you really want all timestamps to agree whether you log to file, console or rosout, so that would indicate take the timestamp at the time of the log call and pass it to the output handlers. On the other, someone might want, system time, steady time, or sim time. So this pr was sort of an incremental step between where we have system time and in the future we could add more timestamps or require the output handlers to get their own various timestamps themselves.

wjwwood:
I'm happy to change this pr how ever you think would be best. I just opened it because I wanted to see how much time passed between logs.

I'll fix the failing tests (sort of fragile since they will break anytime the file has new lines added above a certain point, but unavoidable I think).

src/logging.c
dhood: agreed!

dhood:
I agree that this is in the right direction. To me the contents of the PR are fine now, it's just the missing test coverage that I think should be addressed before merging

src/logging.c
wjwwood: I don't think I can test this case, since I get no feedback (other than output to stderr) when this occurs (there's no return code from the function).

wjwwood: I fixed the comparisons here: 7f1a17b

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3921/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1038/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3244/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4014/)

dhood:
The new output format tokens aren't currently tested: could you please add/extend a test to cover this? Here are the other ones that check formatting: https://github.com/ros2/rcutils/pull/53/files#diff-4783eea3d9e98e3d3b03afdb62839106R36

src/logging.c
```diff
@@ -518,6 +529,24 @@ void rcutils_logging_console_output_handler(
...
+    } else if (strcmp("seconds", token) == 0) {
+      if (snprintf(
+          numeric_storage, sizeof(numeric_storage),
+          "%f", timestamp / 1e9) <= 0)
+      {
+        RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to convert float to string.\n");
+        goto cleanup;
+      }
+      token_expansion = numeric_storage;
+    } else if (strcmp("nanoseconds", token) == 0) {
```
dirk-thomas: I guess users might use `{seconds}.{nanoseconds}` but then get confused by the variable nature of digits of the nanoseconds part. I think an additional placeholder is desired / necessary which pads the output with leasing zeros to a fixed length of 9 digits.

wjwwood: I think you're confusing the purpose of `{nanoseconds}`, it's the total timestamp in nanoseconds, not the nanoseconds part of time. `{seconds}` is floating point time. So, `{seconds}.{nanoseconds}` doesn't really make any sense. See the documentation:

https://github.com/ros2/rcutils/pull/85/files#diff-f1d39d4e841937fde043a8e49b7e6c2aR62

dirk-thomas: Oh, got it now. Thanks for clarifying.

dhood: that was how I also first interpreted it (as sec.nanosec), so if it's happened to a few people maybe there's a way we can rename it to avoid confusion. We could just do `time` like in ROS 1 and additionally `time_ns`? Then you'd be less likely to think `{time}.{time_ns}`, but 'time' is not as descriptive as we'd like moving forward..

wjwwood: It's a fair point, however, that the padding is never considered. So on the floating point on the seconds may not be consistent from machine to machine. I'll see what I can do about that, as ideally it would be similar.

I'm actually writing the missing test right now, I so I'll think if a fixed padding makes sense in either case.

wjwwood: It doesn't need to be concise I don't think, we could use `timestamp` and `timestamp_as_nanoseconds`.

wjwwood: @dhood and I talked about it offline and we decided on `time` and `time_as_nanoseconds`, using `time` rather than `timestamp` to be closer to what ROS 1 does, and using `_as_nanoseconds` to be maximumly specific without worrying about verbosity too much because we expect it to not be used that much in favor of the floating point seconds version.

wjwwood:
I factored out the conversion of the timepoints to public functions with tests. They now print fixed width, so for the nanoseconds always 20 characters (19 digits and either a `-` if negative or a leading ` ` if positive), and the seconds is always 21 characters (same as nanoseconds, but with a decimal point `.`).

I also fixed the logging tests, only the last test case was being checked due to a logical bug in the test. I also fixed a cmake style (though lint_cmake made me indent it weird) and prevented some "error being overwritten" messages in some of the tests. Those aren't really an issue, but it's nice to search for them in the CI output periodically and these were false positives.

include/rcutils/time.h
```diff
@@ -100,6 +100,84 @@ RCUTILS_WARN_UNUSED
 rcutils_ret_t
 rcutils_steady_time_now(rcutils_time_point_value_t * now);
 
+/// Return a timepoint as nanoseconds in a string.
+/**
+ * The number is always fixed width, with left padding zeros up to the maximum
+ * number of digits the timepoint can represent.
+ * Right now that is 19 digits (so 19 characters) for a signed 64-bit integer.
+ * Positive timepoint values will have a leading space, whereas negative values
```
dhood: I understand that the motivation for this is so that positive and negative outputs are aligned, but isn't that only going to be necessary in like 0.1% of use cases? I think the small use case of swapping between negative and positive timestamps and wanting them to be aligned doesn't warrant the extraneous space for 99.9% of use cases (I might be missing something though). Having the default output be `[ 123.456]` is surprising to me.

include/rcutils/time.h
```diff
@@ -100,6 +100,84 @@ RCUTILS_WARN_UNUSED
...
+ * ------------------ | -------------
+ * Allocates Memory   | No [1]
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ * <i>[1] if `snprintf()` does not allocate additional memory internally</i>
+ *
+ * \param[in] timepoint the time to be made into a string
+ * \param[out] str the output string in which it is stored
+ * \param[in] str_size the output string in which it is stored
```
dhood: this doc line is copypasta (and below)

src/time.c
```diff
@@ -0,0 +1,91 @@
...
+  const rcutils_time_point_value_t * timepoint,
+  char * str,
+  size_t str_size)
+{
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(timepoint, RCUTILS_RET_INVALID_ARGUMENT, allocator)
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(str, RCUTILS_RET_INVALID_ARGUMENT, allocator)
+  if (*timepoint >= 0) {
+    // have an explicit leading space to align with negative values which start with `-`
+    if (rcutils_snprintf(str, str_size, " %.19" PRId64, *timepoint) <= 0) {
```
dhood: If `str_size` was specified as 1, the `snprintf`'s return value will be 0 from my understanding ("The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character."). I don't think we should error in that case anymore. However, there's [a test below](https://github.com/ros2/rcutils/pull/85/files#diff-885f7a53192e3dffac3511e600e2afbdR159) for `str_size=1` that expects `RCUTILS_RET_OK`, is that test passing?

test/test_logging_output_timestamps.regex
```diff
@@ -0,0 +1 @@
+'[\s,-][0-9]{10}[.][0-9]{9}' '[\s,-][0-9]{19}'\n'[\s,-][0-9]{10}[.][0-9]{9}' '[\s,-][0-9]{19}'
```
dhood: `\s` instead of ` ` will also match other whitespace characters which seems unnecessarily loose. Was there a reason for not just using a space?

include/rcutils/time.h
wjwwood: > I understand that the motivation for this is so that positive and negative outputs are aligned, but isn't that only going to be necessary in like 0.1% of use cases?

That's right, I had reservations about it too, but given I made these changes in the first place in order to try and keep a fixed width I was placing extreme value on never having a different string length. However, the case pointed out for floats (variable mantissa) is far more likely and the case I'm optimizing for here should honestly never happen (that's why I originally made the timepoint unsigned 😛).

I can change this.

src/time.c
wjwwood: All of the tests are passing (and running 😄) from what I can tell. As well as on Windows where the `snprintf` like functions some times behave slightly differently.

I think that it will return `21` or whatever it _would_ have been, regardless of the value of `str_size`, rather than returning how many it _actually_ wrote based on truncating due to `str_size` being smaller than the result.

test/test_logging_output_timestamps.regex
wjwwood: Nope, I can change that and rerun the tests.

include/rcutils/time.h
dhood: +1 for having the default misaligned for positive/negative numbers, if the misalignment bothers someone they can add an option for it :smile: I know we pad in some cases in ros1 e.g. `[ERROR]` vs [ INFO]` but those are much more likely.

src/time.c
dhood: ohhh, I see. it will successfully put (0 chars + NUL) but won't _return_ 0. good good, this clears it up, thanks

test/test_logging_output_timestamps.regex
dhood: no need now I was mostly just curious if there was regex trickery going on

wjwwood:
Ok, I rebased and addressed the comments, waiting for re-review before CI.

wjwwood:
Also have to make a new rclcpp pr because I messed that up, and fix rclpy warnings, before CI.

src/time.c
```diff
@@ -32,57 +32,44 @@ RCUTILS_PUBLIC
...
+  uint64_t nanoseconds = abs_time_point % (1000 * 1000 * 1000);
+  if (
+    rcutils_snprintf(
+      str, str_size, "%s%.10" PRId64 ".%.9" PRId64,
+      (*time_point >= 0) ? "" : "-", seconds, nanoseconds) <= 0)
+  {
+    RCUTILS_SET_ERROR_MSG("failed to format time point into string as float seconds", allocator)
+    return RCUTILS_RET_ERROR;
   }
+  // if (*time_point >= 0) {
```
dhood: leftover :smile:

include/rcutils/time.h
wjwwood: Oops, 94dcf14 😄

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4090/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1186/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4150/)

I didn't see the `rclpy` stuff locally, and I think it might be resolved by the rebase.

include/rcutils/time.h
```diff
@@ -100,6 +100,84 @@ RCUTILS_WARN_UNUSED
...
 
+/// Return a time point as nanoseconds in a string.
+/**
+ * The number is always fixed width, with left padding zeros up to the maximum
+ * number of digits the time point can represent.
+ * Right now that is 19 digits (so 19 characters) for a signed 64-bit integer.
+ * Negative values will have a leading `-`, so they will be one character
+ * longer than the positive values.
+ *
+ * The recommended minimum size of the input string is 32 characters, but
```
dirk-thomas: How is this "recommended minimum size" relevant in the docblock? Isn't is sufficient to document the exact max size necessary? Also why 32?

Same below.

wjwwood: Because the usable block size on my machine is `16` bytes, and `32` is the smallest multiple of that in which the exact size of this string fits. Also, it gives us a little flexibility for the future, e.g. if we wanted to add `+` to the front of the time to distinguish it from `-`. Not saying we will, but if when you ask for 21 characters malloc gives you 32 anyways...

This is the program I used to check:

```c++
#include <cassert>
#include <cstdio>

#if __linux__
#include <malloc.h>
#else
#include <malloc/malloc.h>
#endif

int main()
{
  uint8_t * bytes = (uint8_t *)malloc(1);
  assert(bytes);
#if __linux__
  printf("%zu\n", malloc_usable_size(bytes));
#else
  printf("%zu\n", malloc_size(bytes));
#endif
  return 0;
}
```

I didn't try it on Windows.

dhood:
there are failing windows tests in case you didn't notice @wjwwood

wjwwood:
Ok... I had to rewrite the `rcutils_snprintf()` functions, but I think I got it all fixed up.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1386/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3635/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4519/)

src/snprintf.c
```diff
@@ -0,0 +1,63 @@
...
+  int ret = rcutils_vsnprintf(buffer, buffer_size, format, args);
+  va_end(args);
+  return ret;
+}
+
+int
+rcutils_vsnprintf(char * buffer, size_t buffer_size, const char * format, va_list args)
+{
+  if (
+    !(NULL == buffer && 0 == buffer_size) &&
```
dhood: I imagine the purpose of this is to allow the use case of passing (null,0,format,..) in order to get the number of characters that would-be formatted (as is done [here](https://github.com/ros2/rcutils/pull/85/files#diff-d20fbd64fcff93291e6ef221061e976aR51)). However, I still think we should ensure that `format` is not null in that case.

On windows, https://msdn.microsoft.com/en-us/library/d3xd30zz.aspx suggests to me this use case of calculating would-be characters is not supported by `_vsnprintf_s` (it will set `errno` if any are null, and we will end up returning -1 below). It would be good for portability to instead use `_vscprintf` in that case, so that `rcutils_vsnprintf` can always be used with a null buffer, at which point we can replace checks [such as this](https://github.com/ros2/rcutils/pull/85/files#diff-d20fbd64fcff93291e6ef221061e976aR48) with a single call to `rcutils_vsnprintf`. Could be left for another day though since we already have this disparity.

include/rcutils/snprintf.h
```diff
@@ -20,30 +20,41 @@ extern "C"
...
+ *
+ * > Number of characters written if successful or negative value if an error
+ * > occurred.
+ * > If the resulting string gets truncated due to buf_size limit, function
+ * > returns the total number of characters (not including the terminating
+ * > null-byte) which would have been written, if the limit was not imposed.
  *
  * \see snprintf()
+ * \see _snprintf_s()
+ * \returns -1 if there is an error, but not if there is truncation
```
dhood: `vsnprintf` can return a "negative value if an error occurred", so we should either map that to -1 or use the same wording here

src/snprintf.c
wjwwood: That's a good point (or a couple). I'll fix it up some and get you to look again.

wjwwood:
@dhood I think I addressed your feedback, can you give it another look?

More CI in the meantime (always assume success 😬):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4455/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1390/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3640/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4525/)

test/test_time.cpp
```diff
@@ -117,3 +118,109 @@ TEST_F(TestTimeFixture, test_rcutils_steady_time_now) {
...
+
+  // test truncations
+  timepoint = 100;
+  ret = rcutils_time_point_value_as_nanoseconds_string(&timepoint, buffer, 18);
+  EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string_safe();
+  EXPECT_STREQ("00000000000000001", buffer);
+
+  const char * test_str = "should not be touched";
+  timepoint = 100;
+  (void)memmove(buffer, test_str, std::min(sizeof(buffer), strlen(test_str)));
```
dhood: Correct me if I'm wrong, but if we're gonna go the `std::min` route, then:
If `std::min` returns `sizeof(buffer)`, shouldn't we manually null terminate `buffer`?
In the other case, I think you'll want to copy `strlen(test_str)+1` (or manually null terminate `buffer`).

test/test_snprintf.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <gtest/gtest.h>
+
+#include "rcutils/snprintf.h"
+
+// Tests the rcutils_snprintf() function.
+TEST(TestSnprintf, test_snprintf) {
+  const char * test_str = "0123456789";
+  char buffer[256];
+  int ret;
+
```
dhood: be good to have a test for the `buffer=null, size=0` case, which we know has different behaviour in the underlying function on windows

include/rcutils/snprintf.h
```diff
@@ -20,30 +20,47 @@ extern "C"
...
+ * > null-byte) which would have been written, if the limit was not imposed.
+ *
+ * If `NULL` and `0` are given for buffer and buffer_size respectively, the
+ * size of the string that would be generated is returned.
+ * Either snprintf() or _vscprintf() is used to calculate this value.
  *
  * \see snprintf()
+ * \see _snprintf_s()
+ * \returns the number of bytes that would have been written given enough space,
+ *   or a negative number (usually -1) if there is an error,
```
dhood: is it useful to provide "usually -1"? It's not obvious to me how a user can make use of that info

test/test_time.cpp
wjwwood: You're right, I'll refactor this and add an expect case which checks for properly null terminated strings.

test/test_snprintf.cpp
wjwwood: I'll add one.

dhood:
@wjwwood I rebased and addressed the two comments if you want to take a look at those commits at some point

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4692/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1579/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3893/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4770/)

wjwwood:
Changes lgtm, thanks @dhood!

wjwwood:
Thanks for pushing this through @dhood!

dhood:
nooo problem!

dhood:
Wiki updates:
added to [logger config demo](https://github.com/ros2/ros2/wiki/Logging-and-logger-configuration#console-output-formatting)
[logger page](https://github.com/ros2/ros2/wiki/Logging#console-output-configuration) references rcutils docs which will include this change in the bouncy release

PR #90, opened 2018-02-06
time sanity checks

### LINKED ISSUE #88: sanity check for rcutils_steady_time_now
`rcutils_steady_time_now` currently trusts the OS specific implementations to provide truly monotonic time values. Issue #86 and [Internet](https://stackoverflow.com/questions/3657289/linux-clock-gettimeclock-monotonic-strange-non-monotonic-behavior) show the monotonicity claim should be verified.

serge-nikulin:
- monotonicity check
- overflow checks

closes #88

### DIFF
--- a/src/time_unix.c
+++ b/src/time_unix.c
@@ -68,9 +68,29 @@ rcutils_system_time_now(rcutils_time_point_value_t * now)
     RCUTILS_SET_ERROR_MSG("unexpected negative time", rcutils_get_default_allocator());
     return RCUTILS_RET_ERROR;
   }
-  *now = RCUTILS_S_TO_NS((uint64_t)timespec_now.tv_sec) + timespec_now.tv_nsec;
-  return RCUTILS_RET_OK;
+
+  rcutils_time_point_value_t total_seconds_in_ns =
+    RCUTILS_S_TO_NS(timespec_now.tv_sec);
+  bool overflow_happened = timespec_now.tv_sec > (INT64_MAX / 1000000000LL);
+
+  rcutils_time_point_value_t total_ns =
+    total_seconds_in_ns + timespec_now.tv_nsec;
+  overflow_happened = overflow_happened ||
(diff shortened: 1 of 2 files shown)


wjwwood:
Thanks for the pr, I will need to make some changes to allow us to accept it, but I'm working on a pr for that now. I just wanted to make sure others weren't also doing the same thing.

serge-nikulin:
@wjwwood, do you have ETA for this one? I would like to fork *after* this one. Thanks!
CC: @dejanpan

wjwwood:
@serge-nikulin I've been working on this, but there are a few things I'm trying to address in this pr before I'd consider it for merge:

- provide a way to force memory allocation in the initialization of the thread-local storage so it is avoided during runtime
  - allocation can occur on first access of thread-local storage in new threads, even when using the new `thread_local` keyword, but especially when you implement using pthreads
- provide a pthread based implementation, since we've had users which build on iOS which doesn't implement `thread_local` or `__thread`
- provide a way for users to disable the checks

So I don't have an eta yet, but I'm working on it.

serge-nikulin:
@wjwwood, how about using shared, non-TLS `last sample` and serializing access to it with interlocked update, critical section, mutiex, etc?

wjwwood:
@serge-nikulin can do that too, but then this function is not lock free anymore. And it could potentially be a high contention lock. I'll think about it though.

wjwwood:
I'm going to close this. I don't think it's unreasonable to rely on the OS to give us monotonic time. If someone is concerned about this they may cherry-pick this patch or a similar one into their local version for debugging.

I know the idea was to have this catch this nonobvious issue early, but the cost of having either a thread-local variable (which may allocate on the first access in each new thread), a lock which may be highly contentious, or something else doesn't seem worth it.

If someone wants to continue pushing on this issue let me know and I'll reopen it.

PR #91, opened 2018-02-15
Misra c round1

mikaelarguedas:
This addresses a subset of misra C violations in this package

/cc @dejanpan

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1079/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3302/) (unrelated osrf_pycommon failures)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4064/) (unrelated osrf_pycommon failures)

### DIFF
--- a/src/string_map.c
+++ b/src/string_map.c
@@ -38,7 +38,7 @@ typedef struct rcutils_string_map_impl_t
 } rcutils_string_map_impl_t;
 
 rcutils_string_map_t
-rcutils_get_zero_initialized_string_map()
+rcutils_get_zero_initialized_string_map(void)
 {
   static rcutils_string_map_t zero_initialized_string_map;
   zero_initialized_string_map.impl = NULL;
--- a/src/string_map.c
+++ b/src/string_map.c
@@ -59,7 +59,7 @@ rcutils_string_map_init(
   RCUTILS_CHECK_ALLOCATOR_WITH_MSG(
     &allocator, "invalid allocator", return RCUTILS_RET_INVALID_ARGUMENT)
   string_map->impl = allocator.allocate(sizeof(rcutils_string_map_impl_t), allocator.state);
-  if (!string_map->impl) {
+  if (NULL == string_map->impl) {
     RCUTILS_SET_ERROR_MSG(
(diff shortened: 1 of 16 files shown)


mikaelarguedas:
Placing this in review and will run CI on OSX ASAP

mikaelarguedas:
MacOS CI passed. ready for review

src/string_map.c
```diff
@@ -219,7 +219,7 @@ rcutils_string_map_clear(rcutils_string_map_t * string_map)
     return RCUTILS_RET_STRING_MAP_INVALID, rcutils_get_default_allocator())
   size_t i = 0;
   for (; i < string_map->impl->capacity; ++i) {
-    if (string_map->impl->keys[i]) {
+    if (string_map->impl->keys[i] != NULL) {
```
sloretz: Just a question, I notice it's consistent about `NULL == something` for equality and `something != NULL` for inequality. Is the order of these a rule in misra C?

mikaelarguedas: No that's not part of MISRA, but it's part of our [design guildelines](https://github.com/ros2/ros2/wiki/Developer-Guide#c)

> Always place literals on the left hand side of comparison operators, e.g. 0 == ret instead of ret == 0
> - rationale: ret == 0 too easily turns into ret = 0 by accident

sloretz: Is our rule for all comparison operators, or just `==`? If the former then should this line be `NULL != string_map->impl->keys[i])`?

mikaelarguedas: Yeah I guess that would be better.
My reading was that it's a lot less likely to write `=` instead of `!=` and more likely to write `=` instead of `==` and that's what we apply everywhere else in the codebase. But I'm not opposed to apply the rule to all comparison operators and not only `==`

sloretz: `!=` becoming `=` seems a lot less likely. I'm not advocating changing anything, my question is answered by you saying it isn't part of Misra C

PR #95, opened 2018-03-09
honor BUILD_SHARED_LIBS in test libraries as well

mikaelarguedas:
suggestion for https://github.com/ros2/ros2/issues/468

connects to ros2/ros2#468

### DIFF
--- a/test/memory_tools/CMakeLists.txt
+++ b/test/memory_tools/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Create the memory_tools library which is used by the tests.
-add_library(${PROJECT_NAME}_memory_tools SHARED memory_tools.cpp)
+add_library(${PROJECT_NAME}_memory_tools memory_tools.cpp)
 if(NOT WIN32)
   # otherwise the compiler will error since the allocator functions have
   # different exception specifiers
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
merging this into #94

dirk-thomas:
One question after the fact: in which case is it actually relevant that these test libraries which are not being installed are being built statically?

mikaelarguedas:
In none that I have in mind, this is just a matter of consistency throughout the package and avoid the potential confusion of "I requested my workspace to be built statically but in my build output N% of the libraries are linked dynamically".

dirk-thomas:
While that might be nice for consistency I don't see a strong need for that for the tests. The motivating goal is to support use cases where the artifacts need to be static on some platforms. The tests are irrelevant in that case. It would be absolutely fine to have a hard coded `STATIC` or `SHARED` on those test targets even when the other libraries are using the type indicated by `BUILD_SHARED_LIBS`.

I just want to clarify this here so that we find a consensus that not all tests *must* obey `BUILD_SHARED_LIBS` in all packages.

dirk-thomas:
E.g. we also build `gtest` as a static library even when `BUILD_SHARED_LIBS=1` is explicitly requested.

mikaelarguedas:
> While that might be nice for consistency I don't see a strong need for that for the tests. The motivating goal is to support use cases where the artifacts need to be static on some platforms. The tests are irrelevant in that case. It would be absolutely fine to have a hard coded STATIC or SHARED on those test targets even when the other libraries are using the type indicated by BUILD_SHARED_LIBS.
I just want to clarify this here so that we find a consensus that not all tests must obey BUILD_SHARED_LIBS in all packages.

That's a fair point, I agree that we shouldn't require that all tests must obey `BUILD_SHARED_LIBS`.

In this particular case (that was the original motivation that I forgot along the way :smile: ) we've been discussing in several occasions making "memory tools" a reusable piece of code by other packages and other projects so it will likely not be an "rcutils test library never installed" for long. With that in mind it makes sense IMO to make it respect `BUILD_SHARED_LIBS` like any other library we expect other packages to reuse.

PR #97, opened 2018-03-15
make log location parameter const

wjwwood:
It doesn't appear to need to be mutable, and it can be copied by the output handler if needed.


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1177/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3380/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4145/)

PR #101, opened 2018-05-01
update to use new memory_tools from osrf_testing_tools_cpp

wjwwood:
This is just a pr to use the new memory tools that I refactored out of this package and `rcl` too. After this will come a pr to make use of the memory tools in this repository more and fix memory allocation issues.


CMakeLists.txt
```diff
@@ -63,12 +63,21 @@ configure_file(
...
   "import em"
-  "em.invoke(['-o', 'include/rcutils/logging_macros.h', '-D', 'rcutils_module_path=\"${rcutils_module_path}\"', '${CMAKE_CURRENT_SOURCE_DIR}/resource/logging_macros.h.em'])")
+  "\
+em.invoke( \
+  [ \
+    '-o', 'include/rcutils/logging_macros.h', \
+    '-D', 'rcutils_module_path=\"${rcutils_module_path}\"', \
+    '${CMAKE_CURRENT_SOURCE_DIR}/resource/logging_macros.h.em' \
+  ] \
+)")
```
wjwwood: @dhood I broke this line up because it was pretty long. It appears to work correctly, but if you'd prefer it not to be this way I can revert it.

wjwwood:
First round of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4317/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1309/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3546/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4390/)

ubercool2:
@serge-nikulin fyi

dejanpan:
@serge-nikulin fyi

serge-nikulin:
yes, I'm a subscriber.

CMakeLists.txt
dhood: fine by me thanks

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4449/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1384/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3632/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4517/)

test/test_logging_macros.c
```diff
@@ -72,7 +72,7 @@ int main(int argc, char ** argv)
   if (strcmp(g_last_log_event.location->function_name, "main")) {
     return 5;
   }
-  if (g_last_log_event.location->line_number != 65u) {
+  if (g_last_log_event.location->line_number != line_number) {
```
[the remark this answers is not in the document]
dhood: LFTM

wjwwood:
New CI with fixes for previous failures:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1442/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3715/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4600/)

wjwwood:
Another round of CI (fixed the macOS issue, but aarch64 is still failing in a strange way):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4526/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1444/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4603/)

wjwwood:
I'm working on the issue on aarch64.

dhood:
these PRs are back in review right @wjwwood ?

wjwwood:
Feel free to review but I’m still debugging the aarch64 issue.

wjwwood:
I've spent enough time on the `aarch64` issue that I'm not convinced I can sort it out in time for the release. Also, I'm convinced that it's only on shutdown (memory tools is working otherwise). So I've made an issue, collecting my debug information:

https://github.com/osrf/osrf_testing_tools_cpp/issues/3

And I'm going to disable memory tools on ARM for now. I'll put this in review again after I update our cmake to avoid doing memory tools on ARM, in addition to already avoiding it on Windows.

serge-nikulin:
@dejanpan, FYI

test/test_allocator.cpp
```diff
@@ -25,65 +25,73 @@
...
+    disable_monitoring_in_all_threads();
+    osrf_testing_tools_cpp::memory_tools::uninitialize();
   }
 };
 
 /* Tests the default allocator.
  */
 TEST_F(CLASSNAME(TestAllocatorFixture, RMW_IMPLEMENTATION), test_default_allocator_normal) {
-  ASSERT_NO_MALLOC(
+  EXPECT_NO_MALLOC(
```
dhood: is there a default callback? if not, I think that since the callbacks have been removed [above](https://github.com/ros2/rcutils/pull/101/files#diff-f0bc7d22597f044081f4f76be91940d9L37), this won't do anything if an unexpected malloc occurs? Maybe it should be wrapped in [something like this?](https://github.com/osrf/osrf_testing_tools_cpp/blob/88223568ce50cdbe30ab5780154716d9438e83d9/test_osrf_testing_tools_cpp/test/test_example_memory_tools.cpp#L66)

test/test_time.cpp
```diff
@@ -14,56 +14,64 @@
...
-    set_on_unexpected_realloc_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED REALLOC";});
-    set_on_unexpected_free_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED FREE";});
-    start_memory_checking();
+    osrf_testing_tools_cpp::memory_tools::initialize();
+    on_unexpected_malloc([]() {FAIL() << "UNEXPECTED MALLOC";});
+    on_unexpected_realloc([]() {FAIL() << "UNEXPECTED REALLOC";});
+    on_unexpected_calloc([]() {FAIL() << "UNEXPECTED CALLOC";});
+    on_unexpected_free([]() {FAIL() << "UNEXPECTED FREE";});
+    enable_monitoring_in_all_threads();
+    EXPECT_NO_MEMORY_OPERATIONS(;);
```
dhood: could you add a comment for what this is trying to ensure?

test/test_time.cpp
```diff
@@ -14,56 +14,64 @@
...
+using osrf_testing_tools_cpp::memory_tools::expect_no_free_end;
+using osrf_testing_tools_cpp::memory_tools::expect_no_malloc_begin;
+using osrf_testing_tools_cpp::memory_tools::expect_no_malloc_end;
+using osrf_testing_tools_cpp::memory_tools::expect_no_realloc_begin;
+using osrf_testing_tools_cpp::memory_tools::expect_no_realloc_end;
+using osrf_testing_tools_cpp::memory_tools::MemoryToolsService;
+using osrf_testing_tools_cpp::memory_tools::on_unexpected_calloc;
+using osrf_testing_tools_cpp::memory_tools::on_unexpected_free;
+using osrf_testing_tools_cpp::memory_tools::on_unexpected_malloc;
+using osrf_testing_tools_cpp::memory_tools::on_unexpected_realloc;
```
dhood: it's a bit confusing to not see any usage of these in this file (I know they come from the `EXPECT_NO_MEMORY_OPERATIONS` macro).. I imagine there's a reason they either can't or shouldn't be embedded in the macro itself, but wanted to point out my concern (from a usability perspective) nonetheless

test/test_time.cpp
```diff
@@ -14,56 +14,64 @@
...
 {
 public:
   void SetUp()
   {
-    set_on_unexpected_malloc_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED MALLOC";});
-    set_on_unexpected_realloc_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED REALLOC";});
-    set_on_unexpected_free_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED FREE";});
-    start_memory_checking();
+    osrf_testing_tools_cpp::memory_tools::initialize();
+    on_unexpected_malloc([]() {FAIL() << "UNEXPECTED MALLOC";});
```
dhood: I know you probably had a reason for favouring the new name, but I did want to point out from a user perspective that `set_*_callback` seemed to better capture what this function was doing. Given the other macros e.g. `EXPECT_NO_MEMORY_OPERATIONS` taking in a set of statements to check, now when I look at something called `on_unexpected_malloc` I expect it to similarly be checking for unexpected malloc on a particular set of statements (which is not what it's doing). We can survive with the naming either way though, for sure.

wjwwood: I was following the `on_<event>(callback)` style of things like JQuery, e.g. https://api.jquery.com/on/. Personally I find the new syntax easier to read, but that's subjective. I don't really follow:

> I expect it to similarly be checking for unexpected malloc on a particular set of statements (which is not what it's doing).

Reading that, it's like `on_unexpected_malloc(<statements>)`  becomes "on an unexpected malloc for statements `<statements>` do...?". It's an incomplete thought.

So I agree that `set_*_callback` is very unambiguous, but I don't exactly see the issue with `on_*`, and it's 9 characters shorter.

My personal preference would be to keep the new name, but I'm open to other arguments.

test/test_time.cpp
wjwwood: Not sure, looks like a copy-paste typo. I'll remove it.

test/test_time.cpp
wjwwood: I'll see if I can clean that up.

test/test_allocator.cpp
wjwwood: Looks like a bug. I should setup the callbacks before that.

test/test_time.cpp
dhood: Yeah, I don't mean that the function as it's named is exceptionally misleading, just that on first glance I expected that line to be doing _checking_ for unexpected mallocs as opposed to setting a callback, whereas with the other naming it was clearer. If that's a naming pattern used by other languages e.g. jquery as you mentioned, then that's fine now that it's been changed

wjwwood:
I addressed the comments (I think, please have another look @dhood), here's new CI with `--packages-up-to test_osrf_testing_tools_cpp rcutils rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1495/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3773/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4671/)

test/test_time.cpp
wjwwood: > on first glance I expected that line to be doing checking for unexpected mallocs as opposed to setting a callback

That's what I was talking about too, but the keyword `on` doesn't make sense to me without a callback, as-in `on <something> do <this (callback)>`, so I could imagine `on` taking statements _and_ a callback, but not just statements. So maybe it's important to understand why `on` indicates to you (and potentially users) that something is being checked right there.

At any rate, it's documented:

https://github.com/osrf/osrf_testing_tools_cpp/blob/67b321750a863cd0e2a49a8ddd5b1227031513de/osrf_testing_tools_cpp/include/osrf_testing_tools_cpp/memory_tools/testing_helpers.hpp#L26

dhood:
as a note: I haven't reviewed the new code in the new repo

wjwwood:
@dhood that's fine. I think reviewing its usage here and having the tests pass will be good enough for now. Future changes can be reviewed in more detail.

PR #106, opened 2018-06-07
Prevent default logger's level from being unset

dhood:
Otherwise getting the _effective_ level of a logger can return unset, where it should always return one of info, warn, etc (falling back to the default if a logger's level hasn't been explicitly set).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4564/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1474/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3751/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4649/)

### DIFF
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -32,6 +32,12 @@ extern "C"
 #define RCUTILS_LOGGING_SEPARATOR_CHAR '.'
 #define RCUTILS_LOGGING_SEPARATOR_STRING "."
 
+/**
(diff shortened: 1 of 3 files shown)


dhood:
For context, this was discovered while testing https://github.com/ros2/rcl/pull/256 (setting unset would cause debug messages to be printed because unset < debug).

PR #110, opened 2018-07-31
move serialized_message from rmw

Karsten1987:
Connects to ros2/rmw#144

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ set(rcutils_sources
   src/get_env.c
   src/logging.c
   src/repl_str.c
+  src/serialized_message.c
   src/snprintf.c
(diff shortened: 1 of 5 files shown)


mikaelarguedas:
(sorry if this is not ready for review)

I think we mentioned renaming this to be more agnostic as this really is just a binary blob.

serialized_message implies a serialization format that from my understanding will be added to the rmw structure and not the rcutils one. Does this make sense?

Maybe the name change is planned for later when the structures will diverge and not be just forwarded?

Karsten1987:
agreed. I was planning on renaming, but as you mentioned, in a second run.
I thought we have one commit, moving things over and then followed by a renaming to keep the history somewhat understandable.

PR #114, opened 2018-08-09
Compile flag fixes

jpsamper2009:
## Description
- Courtesy of @serge-nikulin 
- This PR fixes warning generated when building with:

```
-Wall
-Wextra
-pedantic
-Wcast-align
-Wunused
-Wconversion
-Wsign-conversion
-Wdouble-promotion
-fvisibility=hidden
```

- Related to ros2/rclcpp#530 and ros2/rcl#276

### DIFF
--- a/src/cmdline_parser.c
+++ b/src/cmdline_parser.c
@@ -30,7 +30,7 @@ bool rcutils_cli_option_exist(char ** begin, char ** end, const char * option)
 char * rcutils_cli_get_option(char ** begin, char ** end, const char * option)
 {
   size_t idx = 0;
-  size_t end_idx = end - begin;
+  size_t end_idx = (size_t)(end - begin);
   for (; idx < end_idx; ++idx) {
     if (strncmp(begin[idx], option, strlen(option)) == 0) {
       break;
--- a/src/error_handling.c
+++ b/src/error_handling.c
@@ -195,7 +195,7 @@ format_error_string(void)
   if (!__rcutils_error_is_set(__rcutils_error_state)) {
     return;
   }
-  size_t bytes_it_would_have_written = snprintf(
+  size_t bytes_it_would_have_written = (size_t)snprintf(
     NULL, 0,
     __error_format_string,
     __rcutils_error_state->message,
--- a/src/format_string.c
+++ b/src/format_string.c
@@ -45,7 +45,7 @@ rcutils_format_string_limit(
   va_list args2;
   va_copy(args2, args1);
   // first calculate the output string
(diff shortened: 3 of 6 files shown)


src/cmdline_parser.c
```diff
@@ -30,7 +30,7 @@ bool rcutils_cli_option_exist(char ** begin, char ** end, const char * option)
 char * rcutils_cli_get_option(char ** begin, char ** end, const char * option)
 {
   size_t idx = 0;
-  size_t end_idx = end - begin;
+  size_t end_idx = (size_t)(end - begin);
```
clalancette: Looking around a bit, I think the basic problem here is that taking the difference between two pointers should be represented by `ptrdiff_t`, not `size_t` (in theory, the difference can be negative: https://bytes.com/topic/c/answers/460624-subtracting-pointers-type-use-store-result).  While changing to this, you'll probably also have to change the type of `idx` to match.

src/error_handling.c
```diff
@@ -195,7 +195,7 @@ format_error_string(void)
   if (!__rcutils_error_is_set(__rcutils_error_state)) {
     return;
   }
-  size_t bytes_it_would_have_written = snprintf(
+  size_t bytes_it_would_have_written = (size_t)snprintf(
```
clalancette: There's actually a few problems here.  `snprintf` is defined to return an integer, because the return value can be negative on failure.  So the return value should be stored in an integer.  After that, `bytes_it_would_have_written` should be checked for negative; if it failed, we definitely don't want to continue on and pass bogus sizes around as unsigned.  Finally, we would need a cast from `int` to `size_t` while passing `bytes_it_would_have_written` into `allocator.allocate()` and the lower-down `snprintf` to avoid the warnings.

src/format_string.c
```diff
@@ -45,7 +45,7 @@ rcutils_format_string_limit(
   va_list args2;
   va_copy(args2, args1);
   // first calculate the output string
-  size_t bytes_to_be_written = rcutils_vsnprintf(NULL, 0, format_string, args1);
+  size_t bytes_to_be_written = (size_t)rcutils_vsnprintf(NULL, 0, format_string, args1);
```
clalancette: Like in error_handling.c, the return value of `rcutils_vsnprintf` is actually `int`, not `size_t`, since it can fail.  The same comments apply about storing in an int, checking for negative, and then casting where it is used.

jpsamper2009:
This PR was a for a much older version of the code. I'll close it and readdress with a newer version (if the issues still exist).

PR #118, opened 2018-09-12
Add LL suffix to avoid c4307

sloretz:
MSVC will warn with code C4307 and calculate an unintended value if the result of `RCUTILS_S_TO_NS` is greater than 32bits. As far as I can tell this is because `1000` is small enough to fit in an `int` so that becomes its type according to https://en.cppreference.com/w/cpp/language/integer_literal . MSVC says the size of `int` is 4 bytes even on 64bit platforms https://msdn.microsoft.com/en-us/library/s3f49ktz.aspx. In the `RCUTILS_S_TO_NS` case this means that `seconds * 1000 * 1000 * 1000` is done using 32bit math and will overflow when seconds is greater than 2.

### DIFF
--- a/include/rcutils/time.h
+++ b/include/rcutils/time.h
@@ -27,18 +27,18 @@ extern "C"
 #include "rcutils/visibility_control.h"
 
 /// Convenience macro to convert seconds to nanoseconds.
-#define RCUTILS_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))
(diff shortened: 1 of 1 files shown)


sloretz:
CI windows testing everything above rcutils. [Build Status](https://ci.ros2.org/job/ci_windows/5149/). I'm assuming this change is obviously safe on other platforms, so I'm proposing to merge this if just this window CI passes.

Issue #113, opened 2018-08-07
Logging macros aren't safe in all contexts


clalancette:
Consider the following structure:

```
if (condition)
    RCUTILS_LOG_ERROR("error");
else
{
    RCUTILS_LOG_INFO("info");
}
```

With the current ROS2 rcutils, this code will fail to compile with something like:

```
/home/ubuntu/teleop_twist_ws/src/badmacro/src/badmacro.cpp: In member function ‘int BadMacro::main()’:
/home/ubuntu/teleop_twist_ws/src/badmacro/src/badmacro.cpp:21:7: error: expected ‘}’ before ‘else’
       else
       ^~~~
In file included from /opt/ros/bouncy/include/rclcpp/client.hpp:39:0,
                 from /opt/ros/bouncy/include/rclcpp/callback_group.hpp:23,
                 from /opt/ros/bouncy/include/rclcpp/node_interfaces/node_base_interface.hpp:25,
                 from /opt/ros/bouncy/include/rclcpp/executor.hpp:31,
                 from /opt/ros/bouncy/include/rclcpp/executors/multi_threaded_executor.hpp:24,
                 from /opt/ros/bouncy/include/rclcpp/executors.hpp:21,
                 from /opt/ros/bouncy/include/rclcpp/rclcpp.hpp:144,
                 from /home/ubuntu/teleop_twist_ws/src/badmacro/src/badmacro.cpp:3:
```

I have a minimal reproducer of this problem over at https://github.com/clalancette/badmacro .  We probably need to wrap all of these macros in the `do { ... } while(0)` construct to make them safe in these circumstances.

dirk-thomas:
We can certainly make the macros not robust but calling a macro like this (without the curly braces) is simply "wrong" / very bad practice. The logging macro could even be compiled out for performance in which case the calling code really needs to use curly braces (if you don't want to replace the macro with a no-op).

clalancette:
> We can certainly make the macros not robust but calling a macro like this (without the curly braces) is simply "wrong" / very bad practice. The logging macro could even be compiled out for performance in which case the calling code really needs to use curly braces (if you don't want to replace the macro with a no-op).

Ah, good point, I had forgotten about the fact that the macros could be compiled out.  I ended up opening a PR on the original problem code to add the braces: https://github.com/ros-drivers/joystick_drivers/pull/126 .  I still think it is a good idea to make the macros more robust, so I'll leave this open (though it is low priority).

[...]

PR #120, opened 2018-10-05
Add macro semicolons

clalancette:
Connects to https://github.com/ros2/rcutils/issues/113


mjcarroll:
Maybe just a quick note on why the `do...while(0)` is necessary?

PR #121, opened 2018-10-29
avoid dynamic memory allocation during error handling

wjwwood:
This pull request refactors the way the error handling API in rcutils works so that it does not use dynamic memory and there does not need an allocator in most cases and also should make it easier to use in real-time and high performance situations.

Overview of major changes:

- `rcutils_error_state_t` now uses fixed length strings
- the formatted error string (`rcutils_get_error_string()`) returns a fixed sized string
- error string formatting no longer uses snprintf, and instead uses custom functions which explicitly avoid memory allocations
- there's no longer an `rcutils_get_error_string_safe()` as you cannot get a nullptr for the error string anymore (so it's always "safe")
- suppression of the warning about overriding the current error now does a string compare
  - this is ok from a performance perspective since it is only checked when the error state is set (it was not cleared before setting it again)
  - this is slightly less optimal than a direct pointer comparison because it could incorrectly suppress the warning when setting the same error message in a loop (same string, not same pointer)
- a new method `rcutils_initialize_error_handling_thread_local_storage` is required to be called in new threads in order to avoid memory allocations in other functions
  - it is the only function that takes an allocator now, and it is only used in certain cases
- the pthread implementation was removed
  - it complicated the implementation significantly
  - I'm not sure it needed anymore, see: https://github.com/ros2/rmw/pull/80#issuecomment-432869993
  - I'd prefer an entirely separate implementation in a separate file if we need to resurrect pthread support
- I've documented the use of dynamic memory allocation for each function and tested it with the memory testing tools in `osrf_testing_tools_cpp`.

Main migration issues:

- code which previously did `rcutils_get_error_string_safe()` will need to change to `rcutils_get_error_string().str` (so long as the temporary is ok)
  - to avoid the temporary, they'll need to do something like `rcutils_error_string_t copy = rcutils_get_error_string();`)
- the error set macros which previously required that you pass an allocator no longer takes an allocator, e.g. change `RCUTILS_SET_ERROR_MSG("char array has no valid allocator", error_msg_allocator);` -> `RCUTILS_SET_ERROR_MSG("char array has no valid allocator");`

These changes are annoying, but I think ultimately they're for the better. I considered keeping existing signatures and doing tick-tock, and that's an option of the macro's which take allocators, but not so for the get error string signature, since it needs to return the copy to the fixed sized string. I decided against doing it for the macros because it would require using a separate name for the new macros and because deprecating them is not easy in pure C.

I have a lot more changes up and down the stack due to these API changes, and I'll open pull requests for those once I get feedback on the proposed changes here.

----

/cc @serge-nikulin fyi


wjwwood:
Here's the first "downstream" pull request due to these API changes: https://github.com/ros2/rmw/pull/153

include/rcutils/error_handling.h
```diff
@@ -22,58 +22,119 @@ extern "C"
...
+  RCUTILS_ERROR_MESSAGE_MAX_LENGTH - \
+  RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH - \
+  RCUTILS_ERROR_STATE_LINE_NUMBER_STR_MAX_LENGTH - \
+  RCUTILS_ERROR_FORMATTING_CHARACTERS - \
+  1
+
+/// Struct wrapping a fixed-size c string used for returning the formatted error string.
+typedef struct rcutils_error_string_t
+{
+  const char str[RCUTILS_ERROR_MESSAGE_MAX_LENGTH];
```
dirk-thomas: Why should the member be `const`? Not making it const would:
* avoid the need to cast it away in functions like `__rcutils_format_error_string` and
* also match user expectations better imo.
If the variable of type `rcutils_error_string_t` itself is `const` then the member is const implicitly.

Same for the members of `rcutils_error_state_t` below.

wjwwood: I can change them.

wjwwood:
CI to see what's missing/broken:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5410/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2120/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4487/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5354/)

src/error_handling.c
```diff
@@ -14,313 +14,150 @@
...
+      "<<<\n",
+      gtls_rcutils_error_string.str,
+      error_string,
+      file,
+      line_number);
+  }
+#endif
+  gtls_rcutils_error_state = error_state;
+  gtls_rcutils_error_string_is_formatted = false;
+  gtls_rcutils_error_string = (const rcutils_error_string_t) {0};
```
clalancette: I don't actually know this syntax in C; what does this do to a statically-allocated array embedded in a structure?

src/error_handling_helpers.h
```diff
@@ -0,0 +1,160 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+size_t
+__rcutils_copy_string(char * dst, size_t dst_size, const char * src)
+{
+  // doesn't matter how long src actually is if it is longer than dst, so limit to dst + 1
```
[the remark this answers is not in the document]
clalancette: I see.  And now I actually understand a bit more about why this feels "weird" to me; it's because these helpers are all in a header file, which is intrinsically "public" to me in C.  Further, I just realized that if more than one file did a `#include "./error_handling_helpers.h"`, the link would fail because of a multiple-symbol issue.  I think I'd feel a lot better about this if this header file just had the prototype for `__rcutils_format_error_string` in it, and then all of the implementation (including the helpers) was in a C file.  This would also solve the multi-symbol issue (though that is less important).  What do you think about that approach?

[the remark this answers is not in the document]
wjwwood: I could do that, but the reason it's in a header file is so that I can create a test for it but I was explicitly not including it in more than one place.

It might feel public, but I don't consider it public since I don't install the header (also why it's in `src` and not in `include`). I was also hoping that it would get inlined because it was in the header, and avoiding another translation unit. That being said I don't mind moving the implementation into a `.c` file.

But does that address the concerns about these functions being unsafe to use if you ignore the assumptions?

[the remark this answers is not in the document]
clalancette: > But does that address the concerns about these functions being unsafe to use if you ignore the assumptions?

Yeah, if they are hidden in a C file, then there is no way for any external users to call them, so I'm much more comfortable with eliding internal sanity checks.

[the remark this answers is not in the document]
wjwwood: But the prototypes are still in a header, so it doesn't change their visibility as it is right now, right? Or do you mean that I should just rename the file to be `.c` and include that (because that also works)?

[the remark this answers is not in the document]
wjwwood: @clalancette can you have a look at that commit and see if there's more you want me to do (I'm happy to continue iterating on it)

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5433/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2137/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4506/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5371/)

Also, OpenSplice:

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/5432/)

wjwwood:
I had to rebase a few repositories due to recent merges, so here's new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5434/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2138/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4507/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5372/)

wjwwood:
The only test failure was due to the launch problem. Also the job is marked as failing, but the job finished (no compiler error). @mjcarroll

wjwwood:
This is waiting on ros2/launch#153 to have clean CI, but otherwise I believe CI is good for this set of pull requests and ready for review.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5440/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2144/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4515/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5378/)

src/error_handling_helpers.h
```diff
@@ -0,0 +1,178 @@
...
+  dst[size_to_copy] = '\0';
+  return size_to_copy;
+}
+
+// do not use externally, internal function which is only to be used by error_handling.c
+static
+void
+__rcutils_reverse_str(char * string_in, size_t string_len)
+{
+  assert(string_in != NULL);
```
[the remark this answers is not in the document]
jacobperron: nit: `if (0 == string_len)`

src/error_handling.c
```diff
@@ -14,313 +14,161 @@
...
-  }
-  snprintf(
-    __rcutils_error_string, bytes_it_would_have_written + 1,
-    __error_format_string,
-    __rcutils_error_state->message,
-    __rcutils_error_state->file,
-    __rcutils_error_state->line_number);
-  // The Windows version of snprintf does not null terminate automatically in all cases.
-  __rcutils_error_string[bytes_it_would_have_written] = '\0';
+  return str1 == str2 || 0 == strncmp(str1, str2, count);
```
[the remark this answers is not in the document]
jacobperron: Curious, the following cope compiles and outputs "true" for me with gcc and g++:

```c
#include <stdio.h>

int mycompare(const char * foo, const char * bar)
{
  return foo == bar;
}

int main(void)
{
  const char * foo = NULL;
  const char * bar = NULL;
  if (mycompare(foo, bar)) {
    printf("true\n");
  }
  else {
    printf("false\n");
  }
  return 0;
}
```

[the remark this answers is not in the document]
jacobperron: But I see what you mean; if they are not initialized as `char *`.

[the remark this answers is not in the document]
wjwwood: You'd have to do this to be equivalent to the code in the repository:

```c++
typedef struct my_string_t { char str[10]; } my_string_t;

int mycompare(const char * foo, const char * bar)
{
  return foo == bar;
}

int main(void)
{
  my_string_t foo = {.str = "test"};
  my_string_t bar = {.str = "test"};
  if (mycompare(foo.str, bar.str)) {
    printf("true\n");
  } else {
    printf("false\n");
  }
  return 0;
}
```

In which case the comparison doesn't produce a warning, but they can also never be null. I could change the `__same_str` function to take the `rcutils_error_string_t` instead.

include/rcutils/error_handling.h
```diff
@@ -22,58 +22,123 @@ extern "C"
...
+#define RCUTILS_ERROR_FORMATTING_CHARACTERS 6  // ', at ' + ':'
+
+// max formatted string length
+#define RCUTILS_ERROR_MESSAGE_MAX_LENGTH 1024
+
+// adjustable max length for user defined error message
+// remember "chained" errors will include previously specified file paths
+// e.g. "some error, at /path/to/a.c:42, at /path/to/b.c:42"
+#define RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH 768
+// with RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH = 768, RCUTILS_ERROR_STATE_FILE_MAX_LENGTH == 228
```
sloretz: nit 229?
```
$ expr 1024 - 768 - 20 - 6 - 1
229
```

include/rcutils/error_handling.h
```diff
@@ -22,58 +22,123 @@ extern "C"
...
+#define RCUTILS_ERROR_STATE_FILE_MAX_LENGTH \
+  RCUTILS_ERROR_MESSAGE_MAX_LENGTH - \
+  RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH - \
+  RCUTILS_ERROR_STATE_LINE_NUMBER_STR_MAX_LENGTH - \
+  RCUTILS_ERROR_FORMATTING_CHARACTERS - \
+  1
```
sloretz: nit Recommend parenthesis to avoid order of operation issues like `2 * RCUTILS_ERROR_STATE_FILE_MAX_LENGTH`

src/error_handling.c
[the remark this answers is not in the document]
wjwwood: Sorry, I'm only half-right about this, the first argument is passed by the user, so it could be null.

I've found some things I'll refactor and test for here.

src/error_handling.c
```diff
@@ -14,313 +14,161 @@
...
-#endif
-    local_allocator = rcutils_get_default_allocator();
+  if (!gtls_rcutils_error_is_set) {
+    return (rcutils_error_string_t) {"error not set"};  // NOLINT(readability/braces)
   }
-  char * error_string = *error_string_ptr;
-  if (error_string != NULL) {
-    local_allocator.deallocate(error_string, local_allocator.state);
+  if (!gtls_rcutils_error_string_is_formatted) {
+    __rcutils_format_error_string(&gtls_rcutils_error_string, &gtls_rcutils_error_state);
```
sloretz: Is `gtls_rcutils_error_string_is_formatted` supposed to be set to `true` after this?

src/error_handling_helpers.h
```diff
@@ -0,0 +1,181 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
```
sloretz: 2018? Is this moved from somewhere?

src/error_handling_helpers.h
```diff
@@ -0,0 +1,181 @@
...
+// do not use externally, internal function which is only to be used by error_handling.c
+static
+size_t
+__rcutils_copy_string(char * dst, size_t dst_size, const char * src)
+{
+  assert(dst != NULL);
+  assert(dst_size > 0);
+  assert(src != NULL);
+  // doesn't matter how long src actually is if it is longer than dst, so limit to dst + 1
+  size_t src_length = strnlen(src, dst_size + 1);
```
sloretz: nit, since the check on line 60 includes `src_length` being equal to `dst_size`, I think the `+ 1` can be omitted here.

src/error_handling.c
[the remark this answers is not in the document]
jacobperron: > Sorry, I'm only half-right about this, the first argument is passed by the user, so it could be null.

As long as they're both not null.

test/test_error_handling_helpers.cpp
```diff
@@ -0,0 +1,150 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
sloretz: nit 2018?

src/error_handling.c
wjwwood: Yup! Good catch. I don't have a good way to test this atm.

include/rcutils/error_handling.h
wjwwood: You're right, I think I applied the -1 for the null terminator twice in my calculator.

src/error_handling_helpers.h
wjwwood: Yeah, that's true, but I wanted to allow it to exceed (I was looking at it in lldb) and not just match. I think it can be omitted now though.

wjwwood:
Ok I think I addressed all of the feedback, can you guys give it a another look.

src/error_handling_helpers.h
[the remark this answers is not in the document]
wjwwood: I'm going to push to get this merged. I'm happy to follow up with more changes in an additional PR if you have more feedback later @clalancette.

wjwwood:
Optimistic CI, pending feedback on the new changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2145/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4516/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5379/)

* Linux (OpenSplice) [Build Status](https://ci.ros2.org/job/ci_linux/5442/)

src/error_handling.c
```diff
@@ -78,9 +78,92 @@ static
...
+    if (0 >= bytes_left) {break;}
+
+    // write the last part of the state error message
+    written = __rcutils_copy_string(offset, bytes_left,
+        "'\n"
+        "\n"
+        "rcutils_reset_error() should be called after error handling to avoid this.\n"
+        "<<<\n");
+    bytes_left -= written;
+    if (bytes_left <= 0) {break;}
```
sloretz: Nit, this check is unnecessary

wjwwood: Yup, I think I can commit that one and rerun the tests locally only. f81a5c8

wjwwood:
Ok, I'm going to take my window to merge. I'm happy to follow up with changes in new pr's if needed.

PR #134, opened 2019-01-29
Drops legacy launch API usage.

hidmic:
Connected to ros2/launch#159. Drops legacy launch API usage in logging test cases.

### DIFF
--- a/test/test_logging_long_messages.py
+++ b/test/test_logging_long_messages.py
@@ -14,42 +14,36 @@
 
 import os
 
-from launch.legacy import LaunchDescriptor
-from launch.legacy.exit_handler import ignore_exit_handler
-from launch.legacy.launcher import DefaultLauncher
-from launch.legacy.output_handler import ConsoleOutput
-from launch_testing import create_handler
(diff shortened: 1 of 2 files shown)


test/test_logging_long_messages.py
```diff
@@ -14,42 +14,36 @@
...
-    rc = launcher.launch()
-
-    assert rc == 0, \
-        "The launch file failed with exit code '" + str(rc) + "'"
-
-    handler.check()
+    launch_service = LaunchService()
+    launch_service.include_launch_description(ld)
+    return_code = launch_test.run(launch_service)
+    assert return_code == 0, 'Launch failed with exit code %r' % (return_code,)
```
mjcarroll: This keeps the behavior consistent, but is it what we actually want to do?  Maybe I'm mistaken, but it seems like we would want to bubble up the actual return code here, rather than masking it.

mjcarroll: Looking elsewhere, it seems that we do this pretty consistently, so I'm guessing it's a non-issue.

PR #135, opened 2019-02-25
Add option to skip memory tool tests for address sanitizer

bhatsach:
Introduce a new cmake option -DSKIP_ASAN_INCOMPATIBLE_TESTS to turn off
memory tool tests. This is needed to prevent the override of LD_PRELOAD or library path
to interfere from the working of address sanitizer

Connects to ros2/ci#245

PR #139, opened 2019-03-21
Use CMake property to determine when to use memory_tools.

nuclearsandwich:
Uses the CMake property established in osrf/osrf_testing_tools_cpp#25 to avoid repeating the same checks performed by memory tools themselves in every dependent test.

Depends on osrf/osrf_testing_tools_cpp#25, and replaces #135, required for ros2/ci#245


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6493/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2848/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5323/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6297/)

PR #140, opened 2019-03-25
Migrate launch tests to new launch_testing features & API

hidmic:
Connected to ros2/launch#208.

### DIFF
--- a/test/test_logging_long_messages.py
+++ b/test/test_logging_long_messages.py
@@ -13,38 +13,41 @@
 # limitations under the License.
 
 import os
+import unittest
 
 from launch import LaunchDescription
-from launch import LaunchService
 from launch.actions import ExecuteProcess
-from launch_testing.legacy import LaunchTestService
-from launch_testing.legacy.output import create_output_test_from_file
+from launch.actions import OpaqueFunction
+
+import launch_testing
+import launch_testing.asserts
 
 
-def test_logging_long_messages():
+def generate_test_description(ready_fn):
+    launch_description = LaunchDescription()
     # Set the output format to a "verbose" format that is expected by the executable output
     os.environ['RCUTILS_CONSOLE_OUTPUT_FORMAT'] = \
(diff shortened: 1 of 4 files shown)


test/test_logging_output_format.py
```diff
@@ -35,55 +38,59 @@ def test_logging_output_format():
...
+    launch_description.add_action(
+        OpaqueFunction(function=lambda context: ready_fn())
     )
 
-    launch_service = LaunchService()
-    launch_service.include_launch_description(ld)
-    return_code = launch_test.run(launch_service)
-    assert return_code == 0, 'Launch failed with exit code %r' % (return_code,)
+    return launch_description, {'processes_to_test': processes_to_test}
+
```
pbaughman: I suspect the lack of 'active' tests here will cause a issue, but it might be a problem that needs to be fixed in launch_testing.  I'm suspicious that in the absence of tests, the test thread can send the 'shutdown' signal to the LaunchService before it's finished launching the processes, which may cause the test runner to hang waiting for the LaunchService to terminate.  For robustness, it might be worth waiting for the process_to_test to terminate.

test/test_logging_long_messages.py
```diff
@@ -13,38 +13,41 @@
...
+class TestLoggingLongMessages(unittest.TestCase):
 
-if __name__ == '__main__':
-    test_logging_long_messages()
+    def test_logging_output(self, proc_output, process_name):
+        """Test executable output against expectation."""
+        proc_output.assertWaitFor(
+            expected_output=launch_testing.tools.expected_output_from_file(
+                path=os.path.join(os.path.dirname(__file__), process_name)
+            ), process=process_name, timeout=10
```
ivanpauno: Maybe, it's clearer process and timeout in a new line.

hidmic: I could but it looks odd, right before the comma if you drop a line here

Issue #146, opened 2019-04-09
`test_string_map` is leaking memory


thomas-moulard:
ASAN detected, will send a PR to fix this.

```
Running main() from ../../../install-asan/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
[==========] Running 16 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 16 tests from test_string_map
[ RUN      ] test_string_map.lifecycle
[       OK ] test_string_map.lifecycle (0 ms)
[ RUN      ] test_string_map.getters
[       OK ] test_string_map.getters (0 ms)
[ RUN      ] test_string_map.reserve_and_clear
[       OK ] test_string_map.reserve_and_clear (0 ms)
[ RUN      ] test_string_map.check_size_and_capacity
[       OK ] test_string_map.check_size_and_capacity (0 ms)
[ RUN      ] test_string_map.set_no_resize_check_size_and_capacity
[       OK ] test_string_map.set_no_resize_check_size_and_capacity (0 ms)
[ RUN      ] test_string_map.set_no_resize_exceeding_capacity
[       OK ] test_string_map.set_no_resize_exceeding_capacity (0 ms)
[ RUN      ] test_string_map.set_no_resize
[       OK ] test_string_map.set_no_resize (1 ms)
[ RUN      ] test_string_map.set
[       OK ] test_string_map.set (0 ms)
[ RUN      ] test_string_map.key_exists
[       OK ] test_string_map.key_exists (0 ms)
[ RUN      ] test_string_map.key_existsn
[       OK ] test_string_map.key_existsn (0 ms)
[ RUN      ] test_string_map.unset
[       OK ] test_string_map.unset (0 ms)
[ RUN      ] test_string_map.get
[       OK ] test_string_map.get (0 ms)
[ RUN      ] test_string_map.getn
[       OK ] test_string_map.getn (0 ms)
[ RUN      ] test_string_map.get_next_key
[       OK ] test_string_map.get_next_key (0 ms)
[ RUN      ] test_string_map.copy
[       OK ] test_string_map.copy (0 ms)
[ RUN      ] test_string_map.strange_keys
[       OK ] test_string_map.strange_keys (0 ms)
[----------] 16 tests from test_string_map (1 ms total)

[----------] Global test environment tear-down
[==========] 16 tests from 1 test case ran. (1 ms total)
[  PASSED  ] 16 tests.

=================================================================
[...]

PR #147, opened 2019-04-09
Fix ASAN failure in test_string_map.cpp

thomas-moulard:
Partially addresses issue described in #146.

### DIFF
--- a/test/test_string_map.cpp
+++ b/test/test_string_map.cpp
@@ -130,54 +130,56 @@ TEST(test_string_map, lifecycle) {
   }
 }
 
-TEST_F(TestStringMap, getters) {
+TEST_F(TestStringMap, getters_capacity_null_list) {
   rcutils_ret_t ret;
+  size_t capacity;
+  ret = rcutils_string_map_get_capacity(NULL, &capacity);
+  EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
+}
 
-  // null for string_map
-  {
-    size_t capacity, size;
-    ret = rcutils_string_map_get_capacity(NULL, &capacity);
-    EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
-    rcutils_reset_error();
-    ret = rcutils_string_map_get_size(NULL, &size);
-    EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
-    rcutils_reset_error();
-  }
+TEST_F(TestStringMap, getters_size_null_list) {
+  rcutils_ret_t ret;
+  size_t size;
+  ret = rcutils_string_map_get_size(NULL, &size);
+  EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
+}
 
-  // null for capacity/size
-  {
-    rcutils_string_map_t string_map = rcutils_get_zero_initialized_string_map();
-    ret = rcutils_string_map_init(&string_map, 0, allocator);
-    ASSERT_EQ(RCUTILS_RET_OK, ret);
+TEST_F(TestStringMap, getters_capacity_null_capacity) {
+  rcutils_ret_t ret;
+  ret = rcutils_string_map_init(&string_map, 0, allocator);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
 
-    ret = rcutils_string_map_get_capacity(&string_map, NULL);
-    EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
-    rcutils_reset_error();
-    ret = rcutils_string_map_get_size(&string_map, NULL);
-    EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
-    rcutils_reset_error();
-  }
+  ret = rcutils_string_map_get_capacity(&string_map, NULL);
+  EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
+}
 
-  // initialize to 0
-  {
-    rcutils_string_map_t string_map = rcutils_get_zero_initialized_string_map();
-    ret = rcutils_string_map_init(&string_map, 0, allocator);
-    ASSERT_EQ(RCUTILS_RET_OK, ret);
+TEST_F(TestStringMap, getters_size_null_size) {
+  rcutils_ret_t ret;
+  ret = rcutils_string_map_init(&string_map, 0, allocator);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
 
-    size_t capacity = 42;
-    EXPECT_EQ(
-      RCUTILS_RET_OK,
-      rcutils_string_map_get_capacity(&string_map, &capacity));
-    EXPECT_EQ(0u, capacity);
+  ret = rcutils_string_map_get_size(&string_map, NULL);
+  EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
+}
 
-    size_t size = 42;
-    ret = rcutils_string_map_get_size(&string_map, &size);
-    EXPECT_EQ(RCUTILS_RET_OK, ret);
-    EXPECT_EQ(0u, size);
+TEST_F(TestStringMap, getters_initialize_to_zero) {
+  rcutils_ret_t ret;
+  ret = rcutils_string_map_init(&string_map, 0, allocator);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
 
-    ret = rcutils_string_map_fini(&string_map);
-    ASSERT_EQ(RCUTILS_RET_OK, ret);
-  }
+  size_t capacity = 42;
+  EXPECT_EQ(
+    RCUTILS_RET_OK,
+    rcutils_string_map_get_capacity(&string_map, &capacity));
+  EXPECT_EQ(0u, capacity);
+
+  size_t size = 42;
+  ret = rcutils_string_map_get_size(&string_map, &size);
+  EXPECT_EQ(RCUTILS_RET_OK, ret);
+  EXPECT_EQ(0u, size);
+
+  ret = rcutils_string_map_fini(&string_map);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
 }
 
 TEST(test_string_map, reserve_and_clear) {
--- a/test/test_string_map.cpp
+++ b/test/test_string_map.cpp
@@ -37,9 +37,15 @@ protected:
 
   void TearDown() final
   {
-    EXPECT_EQ(RCUTILS_RET_OK,
-      rcutils_string_map_fini(&string_map)) <<
-      rcutils_get_error_string().str;
+    // Ensure the next call will be successful even if the test
+    // failed.
+    rcutils_reset_error();
+
+    // In some cases, the test may exit before the list has been initialized.
+    // This is OK to ignore.
+    const rcutils_ret_t ret = rcutils_string_map_fini(&string_map);
+    EXPECT_TRUE(ret == RCUTILS_RET_OK || ret == RCUTILS_RET_NOT_INITIALIZED);
+    rcutils_reset_error();
   }
 
   rcutils_allocator_t allocator;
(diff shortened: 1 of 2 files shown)


thomas-moulard:
This fixes an issue with the `getters` test case. I introduced a fixture in this PR and will expand its usage in the following PRs.

@wjwwood PTAL

tfoote:
@thomas-moulard For best cross referencing please put the issue number into the body. Github doesn't auto link it in the title. 

And if you use `Fixes` like this, github will automatically close it when the PR is merged.

Fixes #146

thomas-moulard:
Thanks @tfoote - this does not completely solve the bug so I rephrased to prevent GitHub from closing the bug.

wjwwood:
Also, please separate (or avoid) style / structural changes into a separate commit. It's hard to tell what's part of using a test fixture and what's addressing an issue detected by ASAN.

test/test_string_map.cpp
```diff
@@ -21,6 +21,32 @@
...
+    allocator = rcutils_get_default_allocator();
+    failing_allocator = get_failing_allocator();
+    string_map = rcutils_get_zero_initialized_string_map();
+  }
+
+  void TearDown() final
+  {
+    EXPECT_EQ(RCUTILS_RET_OK,
+      rcutils_string_map_fini(&string_map)) <<
+      rcutils_get_error_string().str;
```
wjwwood: Doing a fini in the tear down when you don't know if an init occurred isn't a good idea imo.

test/test_string_map.cpp
```diff
@@ -115,10 +140,6 @@ TEST(test_string_map, getters) {
 
   // null for capacity/size
   {
-    rcutils_string_map_t string_map = rcutils_get_zero_initialized_string_map();
-    ret = rcutils_string_map_init(&string_map, 0, allocator);
```
wjwwood: How are you able to remove the init here? Presumably it was done for a reason.

tfoote:
> Maybe you can explain, or show the original error/warning from ASAN, or put a comment in the code, because I don't see what the root issue was here or what the fix was.

@wjwwood The asan output is in #146

thomas-moulard:
```
Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7f5be60feb50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7f5be5e10ba5 in __default_allocate (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/install/rcutils/lib/librcutils.so+0x3ba5)
    #2 0x7f5be5e189ff in rcutils_string_map_init (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/install/rcutils/lib/librcutils.so+0xb9ff)
    #3 0x55a76cca47d4 in test_string_map_getters_Test::TestBody() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x1e7d4)
    #4 0x55a76cd72efb in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xecefb)
    #5 0x55a76cd64ca3 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xdeca3)
    #6 0x55a76cd10a97 in testing::Test::Run() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x8aa97)
    #7 0x55a76cd11ec2 in testing::TestInfo::Run() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x8bec2)
    #8 0x55a76cd12a66 in testing::TestCase::Run() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x8ca66)
    #9 0x55a76cd2db77 in testing::internal::UnitTestImpl::RunAllTests() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xa7b77)
    #10 0x55a76cd759ae in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xef9ae)
    #11 0x55a76cd66f6c in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xe0f6c)
    #12 0x55a76cd2a90b in testing::UnitTest::Run() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0xa490b)
    #13 0x55a76ccfde5a in RUN_ALL_TESTS() (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x77e5a)
    #14 0x55a76ccfdda0 in main (/home/ANT.AMAZON.COM/tmoulard/ros2_ws/build-asan/rcutils/test_string_map+0x77da0)
    #15 0x7f5be527db96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
```

test/test_string_map.cpp
thomas-moulard: I modified this code to allow the `*_fini`to fail silently in this case as it won't leak memory.
I don't think it's worth splitting the fixture to handle those cases differently.

test/test_string_map.cpp
thomas-moulard: Oops - reverted

thomas-moulard:
My bad for the stylistic change, I will do better next time.

thomas-moulard:
Now that this test case is fixed, no ASAN output:

```
./test_string_map --gtest_filter=test_string_map.getters
Running main() from ../../../install-asan/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
Note: Google Test filter = test_string_map.getters
[==========] Running 0 tests from 0 test cases.
[==========] 0 tests from 0 test cases ran. (0 ms total)
[  PASSED  ] 0 tests.

```

test/test_string_map.cpp
thomas-moulard: Now, the initialization is done in the fixture so this is not necessary anymore here.

thomas-moulard:
I broke down the test a bit more to simplify the memory management.
Here is the output of the test cases impacted by this PR. Filtering out the other ones
as there are other issues in other test cases I'd rather tackle in a follow-up PR.

```
./test_string_map --gtest_filter='TestStringMap.*'
Running main() from ../../../install-asan/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
Note: Google Test filter = TestStringMap.*
[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from TestStringMap
[ RUN      ] TestStringMap.getters_capacity_null_list
[       OK ] TestStringMap.getters_capacity_null_list (0 ms)
[ RUN      ] TestStringMap.getters_size_null_list
[       OK ] TestStringMap.getters_size_null_list (0 ms)
[ RUN      ] TestStringMap.getters_capacity_null_capacity
[       OK ] TestStringMap.getters_capacity_null_capacity (0 ms)
[ RUN      ] TestStringMap.getters_size_null_size
[       OK ] TestStringMap.getters_size_null_size (0 ms)
[ RUN      ] TestStringMap.getters_initialize_to_zero
[       OK ] TestStringMap.getters_initialize_to_zero (0 ms)
[----------] 5 tests from TestStringMap (0 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 1 test case ran. (0 ms total)
[  PASSED  ] 5 tests.
```

test/test_string_map.cpp
thomas-moulard: oops - I reverted the change, `getters_capacity_null_capacity` and `getters_size_null_size` are operating on an initialized list again.

test/test_string_map.cpp
```diff
@@ -21,6 +21,38 @@
...
+  void TearDown() final
+  {
+    // Ensure the next call will be successful even if the test
+    // failed.
+    rcutils_reset_error();
+
+    // In some cases, the test may exit before the list has been initialized.
+    // This is OK to ignore.
+    const rcutils_ret_t ret = rcutils_string_map_fini(&string_map);
+    EXPECT_TRUE(ret == RCUTILS_RET_OK || ret == RCUTILS_RET_NOT_INITIALIZED);
```
wjwwood: I really don't like this. I think calls to `init` and `fini` should be matched. I see that they are matched in the `getters_initialize_to_zero` test case below, so is this even needed? If a test case is missing a `fini` and therefore this case is needed, then I think it's a bug in the test case.

To a lesser degree I dislike the defensive `rcutils_reset_error` calls, as they lead to cases which I'll point out below.

I think this would be better with the entire `TearDown` removed.

test/test_string_map.cpp
```diff
@@ -98,55 +130,56 @@ TEST(test_string_map, lifecycle) {
   }
 }
 
-TEST(test_string_map, getters) {
-  auto allocator = rcutils_get_default_allocator();
+TEST_F(TestStringMap, getters_capacity_null_list) {
   rcutils_ret_t ret;
+  size_t capacity;
+  ret = rcutils_string_map_get_capacity(NULL, &capacity);
+  EXPECT_EQ(RCUTILS_RET_INVALID_ARGUMENT, ret) << rcutils_get_error_string().str;
```
wjwwood: Not doing a `rcutils_reset_error` here is a bad practice in my opinion, because if this case were to be extended and another rcutils function that can fail were called after this line, it would be missing an `rcutils_reset_error` between them. Saving one call to it at the end of each case doesn't save that much, but can lead to forgetting to add it in follow up cases.

I'd prefer a helper function that gets the error and resets it, which could be the right hand side of the `<<` in the expect statement. But even that I feel is unnecessary and error prone, e.g. if the expect passes but the function still failed then the error would not be fetched and therefore not reset.

test/test_string_map.cpp
wjwwood: > so is this even needed?

Nevermind, I see that `getters_size_null_size` init's but does not fini, which is inconsistent with `getters_initialize_to_zero`. Is there a reason for that? I still think removing this from the tear down and doing the fini in the test case is the best solution.

thomas-moulard: Removed the TearDown.

test/test_string_map.cpp
thomas-moulard: This LGTM, I put back the `rcutils_reset_error` everywhere.

thomas-moulard:
Thanks William! I removed the `TearDown` and put back the `rcutils_reset_error` into the different test cases.

I let on purpose the one in `SetUp` on purpose. Rationale: if an exception is thrown (very unlikely) or if a test does not clean the error state properly, it can take time to understand that the failing test is actually the previous one. It also can have weird implications if gtest flags are passed like to randomize the test order.
I don't have a strong preference though. I can remove this one too if you prefer.

thomas-moulard:
Additionally, a naive question - what's the point of using `EXPECT_*` instead of `ASSERT_*` if we don't reset the error state for each EXPECT. Won't this cause problems when the next `rcutils` function is called?

wjwwood:
> Additionally, a naive question - what's the point of using EXPECT_* instead of ASSERT_* if we don't reset the error state for each EXPECT. Won't this cause problems when the next rcutils function is called?

Ideally we would reset the error after each EXPECT which may fail, but sadly we don't have a great track record on that, as it's easy to overlook and hard to automatically check.

As to why not use ASSERT, there may be cases where that makes more sense. In the other pr I commented about a case where the we EXPECT after an `rcutils_something_init` when that should almost certainly just be ASSERT.

thomas-moulard:
@wjwwood finally I added the scope exit macro in this PR to. Triggered a CI build, should be OK to merge it green.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6672/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2991/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5480/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6501/)

ros-discourse:

https://discourse.ros.org/t/introducing-ros2-sanitizer-report-and-analysis/9287/1

PR #154, opened 2019-04-27
Use GCC extension for printf-like functions

thomas-moulard:
This commit annotates rcutils_format_string_limit, rcutils_log
and rcutils_snprintf with the GCC extension for printf-like
statements.

Annotating this function allows GCC to emit warnings when dangerous
patterns are used through the whole codebase.

As those functions are wrappers around different flavors of printf,
surfacing those warnings prevents dangerous printf statements.

For instance:

```
const char* message = // some dynamically evaluated logging message
printf(message);
```

...is unsafe. Instead, we should be using:

```
printf("%s", message);
```

...to avoid crashes if the variable "message" contains a sequence
of characters printf may recognize (like "%d").

printf-related warnings are enabled by '-Wformat*' flags ('-Wformat',
'-Wformat-security', etc.).

The complete list of flags and detected patterns is available in
the GCC documentation:

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

Note: clang does also support this attribute but not MSVC. On
Windows, no additional safety is provided by this commit.

### DIFF
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -64,6 +64,32 @@ extern "C"
 #define RCUTILS_STRINGIFY(x) RCUTILS_STRINGIFY_IMPL(x)
 #define RCUTILS_UNUSED(x) (void)(x)
 
+#if defined _WIN32 || defined __CYGWIN__
+#define RCUTILS_ATTRIBUTE_PRINTF_FORMAT(format_string_index, first_to_check_index)
+#else
+// This macro can be used to annotate printf-like functions which are relying
+// on a format string and a variable number of arguments.
+//
+// This enables GCC to emit warnings if dangerous patterns are detected.
+// See GCC documentation for details:
+// https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html
+//
+// format_string_index is the index of the format string passed to the function
+// first_to_check_index is the index of the first "optional argument".
+//
+// For the following function:
+// int snprintf(char *str, size_t size, const char *format, ...);
+//              ^^^^^^^^^  ^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^  ^^^
+//              ARG 1      ARG 2        ARG 3               ARG 4
+//                                      format string       first optional argument
+//
+// format_string_index value would be 3, first_to_check_index value would be 4.
+//
+// IMPORTANT: the first argument has an index of ONE (not zero!).
+#define RCUTILS_ATTRIBUTE_PRINTF_FORMAT(format_string_index, first_to_check_index) \
+  __attribute__ ((format (printf, format_string_index, first_to_check_index)))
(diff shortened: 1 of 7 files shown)


test/test_logging.cpp
```diff
@@ -113,15 +113,15 @@ TEST(CLASSNAME(TestLogging, RMW_IMPLEMENTATION), test_logging) {
   EXPECT_EQ("name3", g_last_log_event.name);
   EXPECT_EQ("message 33", g_last_log_event.message);
 
-  rcutils_log(NULL, RCUTILS_LOG_SEVERITY_WARN, "", "");
+  rcutils_log(NULL, RCUTILS_LOG_SEVERITY_WARN, "", "%s", "");
```
thomas-moulard: This prevents GCC emitting warning (see `-Wformat-zero-length`)

include/rcutils/macros.h
```diff
@@ -64,6 +64,32 @@ extern "C"
...
+//
+// For the following function:
+// int snprintf(char *str, size_t size, const char *format, ...);
+//              ^^^^^^^^^  ^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^  ^^^
+//              ARG 1      ARG 2        ARG 3               ARG 4
+//                                      format string       first optional argument
+//
+// format_string_index value would be 3, first_to_check_index value would be 4.
+//
+// IMPORTANT: the first argument has an index of ONE (not zero!).
```
dirk-thomas: Please use a proper docblock instead of just comment lines to document the macro (in both cases).

thomas-moulard: Done. I changed the doc for the Windows case. Let me know if this is the right strategy.

dirk-thomas: LGTM

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6829/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3124/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6654/)

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6840/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3133/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6662/)

thomas-moulard:
@dirk-thomas Good to go on my side 🚢

dirk-thomas:
Thanks for the improvement.

dirk-thomas:
@thomas-moulard This patch broke all nightly builds. Please run more through tests in the future. ros/pluginlib#152 will hopefully fix the builds.

PR #157, opened 2019-05-06
Rename result variable for clarity.

nuclearsandwich:
This is a followup to #155. As mentioned by @clalancette `basename_copy` is also a candidate as it is what `basec` is intended to abbreviate.


nuclearsandwich:
CI for this change was run in https://github.com/ros2/rcl_logging/pull/12#issuecomment-489793091

Though there are failures in the run, they're all due to the fact that linters are now being run for the rcl_logging packages so this small internal change broke nothing (as one would hope).

PR #158, opened 2019-05-14
Add colorized output

ivanpauno:
It is just what the title says.

Working well in Linux. Colors were taken from ROS1:
https://github.com/ros/rosconsole/blob/ba4e1ae06b5bce3039ac5d20853248b51198e431/src/rosconsole/rosconsole.cpp#L75-L85
https://github.com/ros/rosconsole/blob/ba4e1ae06b5bce3039ac5d20853248b51198e431/src/rosconsole/rosconsole.cpp#L341-L363

I haven't tried it on MAC, but as far as I understand it should work as-is.

I could do something like this in windows if it is desired:
https://stackoverflow.com/questions/8765938/colorful-text-using-printf-in-c

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -22,10 +22,12 @@ extern "C"
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
-#include <unistd.h>
 
 #ifdef WIN32
+# include <io.h>
 # include <windows.h>
+#else
+# include <unistd.h>
 #endif
 
 #include "rcutils/allocator.h"
@@ -734,6 +736,7 @@ rcutils_ret_t rcutils_logging_format_message(
       } \
     } \
   }
+# define EXPAND(x) x
 # define APPLY(macro, ...) EXPAND(macro(__VA_ARGS__))
 # define SET_OUTPUT_COLOR_WITH_SEVERITY(status, severity, stream, output_array) \
   { \
--- a/src/logging.c
+++ b/src/logging.c
@@ -20,7 +20,9 @@ extern "C"
 #include <ctype.h>
 #include <inttypes.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "rcutils/allocator.h"
 #include "rcutils/error_handling.h"
@@ -44,17 +46,12 @@ const char * g_rcutils_log_severity_names[] = {
   [RCUTILS_LOG_SEVERITY_FATAL] = "FATAL",
 };
 
-#ifdef WIN32
-  #define COLOR_NORMAL ""
-  #define COLOR_RED ""
-  #define COLOR_GREEN ""
-  #define COLOR_YELLOW ""
-#else
-  #define COLOR_NORMAL "\033[0m"
-  #define COLOR_RED "\033[31m"
-  #define COLOR_GREEN "\033[32m"
-  #define COLOR_YELLOW "\033[33m"
-#endif
+enum rcutils_colorized_output
+{
+  RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE,
+  RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE,
+  RCUTILS_COLORIZED_OUTPUT_DEFAULT,
+};
 
 bool g_rcutils_logging_initialized = false;
 
@@ -75,6 +72,7 @@ int g_rcutils_logging_default_logger_level = 0;
 bool g_force_stdout_line_buffered = false;
 bool g_stdout_flush_failure_reported = false;
 
+enum rcutils_colorized_output g_colorized_output = RCUTILS_COLORIZED_OUTPUT_DEFAULT;
 
 rcutils_ret_t rcutils_logging_initialize(void)
 {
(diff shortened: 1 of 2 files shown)


dirk-thomas:
It would be good to have an env var to control the behavior (force enable, force disable) and the default could be based on if the shell is interactive (if I recall correctly that is what other cases use).

ivanpauno:
> It would be good to have an env var to control the behavior (force enable, force disable) and the default could be based on if the shell is interactive (if I recall correctly that is what other cases use).

I added an env `RCUTILS_COLORIZED_OUTPUT`, possible values: `FORCE_ENABLE`, `FORCE_DISABLE`, `DEFAULT`.
I used `isatty` for checking if it is a terminal or not as default behavior.

I also moved the "add colour logic " from `rcutils_logging_format_message` to `rcutils_logging_console_output_handler`, as the first function is also used for formatting a message in   other log functions (`rcl/src/logging.c: rcl_logging_ext_lib_output_handler`). I suppose we don't want to colorize that output.

ivanpauno:
An unrelated comment:
I think `rcutils_logging_initialize` is using `strcmp` in the wrong way with `line_buffered` (L83 and L86).
https://github.com/ros2/rcutils/blob/343e4065b8ac580ad0636a1d89e134659bf4c9ac/src/logging.c#L77-L98

Also, I think all the code related to `RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED` should be moved to `rcutils_logging_initialize_with_allocator`.

@dirk-thomas if you agree, I will open another PR correcting that.

dirk-thomas:
> possible values: `FORCE_ENABLE`, `FORCE_DISABLE`, `DEFAULT`.

Let's use standard boolean values for the new env var `RCUTILS_COLORIZED_OUTPUT`: `0` to disable the colorization, `1` to force using colorization and when no value is set use the automatic decision based on `isatty`.

ivanpauno:
> > possible values: `FORCE_ENABLE`, `FORCE_DISABLE`, `DEFAULT`.
> 
> Let's use standard boolean values for the new env var `RCUTILS_COLORIZED_OUTPUT`: `0` to disable the colorization, `1` to force using colorization and when no value is set use the automatic decision based on `isatty`.

Done!
Internally, I'm still using an `enum` with the three possibilities, as I have to defer the decision depending if the output stream is `stdout` or `stderr`.

src/logging.c
```diff
@@ -44,6 +46,13 @@ const char * g_rcutils_log_severity_names[] = {
   [RCUTILS_LOG_SEVERITY_FATAL] = "FATAL",
 };
 
+enum rcutils_colorized_output
+{
+  RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE,
+  RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE,
+  RCUTILS_COLORIZED_OUTPUT_DEFAULT,
```
dirk-thomas: May I suggest to use:

* `RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE` implicit value: `0` (matching env var value)
* `RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE` implicit value: `1` (matching env var value)
* `RCUTILS_COLORIZED_OUTPUT_AUTO` implicit value: `2`

src/logging.c
```diff
@@ -628,12 +667,131 @@ rcutils_ret_t rcutils_logging_format_message(
...
+  }
+#ifdef WIN32
+# define IS_OUTPUT_COLORIZED(is_colorized, stream) \
+  { \
+    if (g_colorized_output == RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE) { \
+      is_colorized = true; \
+    } else if (g_colorized_output == RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE) { \
+      is_colorized = false; \
+    } else { \
+      is_colorized = _isatty(_fileno(stream)) != 0; \
```
ivanpauno: I don't know if `_isatty` is the right thing to use here, because it returns non-zero when the filedescriptor is a character device (and not a terminal like in linux/mac).
I tried some examples without seeing a `SetConsoleTextAttribute` error, so I'm happy with it.
- In the following example, I see a `SetConsoleTextAttribute` error if I force colorizing the output with the environment variable. It doesn't appear with the auto behavior.
```bat
ros2 run PACKAGE NODE_THAT_LOGS __log_level:=debug > log.txt
```
- I haven't seen errors when redirecting `stdout` to other character devices. e.g.:
```bat
ros2 run PACKAGE NODE_THAT_LOGS __log_level:=debug > null
```

Option 2:
- Use `isatty`, and add a pragma for avoiding the deprecated warning.

@dirk-thomas What do you think?

dirk-thomas: I don't know enough about `_isatty` but sounds good to me if it works for you..

src/logging.c
```diff
@@ -628,12 +667,131 @@ rcutils_ret_t rcutils_logging_format_message(
...
+# define IS_OUTPUT_COLORIZED(is_colorized, stream) \
+  { \
+    if (g_colorized_output == RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE) { \
+      is_colorized = true; \
+    } else if (g_colorized_output == RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE) { \
+      is_colorized = false; \
+    } else { \
+      is_colorized = isatty(fileno(stream)) != 0; \
+    } \
+  }
```
dirk-thomas: This macro is almost identical to the Windows implementation above. Please avoid that duplication and only customize the `isatty` call using a platform specific macro.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7​11​4/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3332/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5​82​7/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6​93​3/)

ivanpauno:
I have tested locally that linter is passing now in both windows and linux after https://github.com/ros2/rcutils/pull/158/commits/e407efd2a7d6a574a5ec80b0b49e46708eab5e0b. I'm going to skip running ci jobs again only because of linter failures.

peci1:
Just to keep information in one place, I'll amend to this years old PR.

When using `ros2 launch`, the nodes have `isatty == False` for both stdout and stderr. So to get colorized output of ros2 launch, you have to force it using the env variable.

I've added a documentation note about this caveat: https://github.com/ros2/ros2_documentation/pull/5726 .

PR #161, opened 2019-05-20
Correct bug when reading RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED

ivanpauno:
I previously mentioned this in https://github.com/ros2/rcutils/pull/158#issuecomment-492367741.

Two changes:
- Logic for reading `RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED` was wrong. `g_force_stdout_line_buffered` was always silently set to 1, except when you did:
 `export RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED=1`.
- Moved the code from `rcutils_logging_initialize` to `rcutils_logging_initialize_with_allocator`. I think we don't have any direct call to `rcutils_logging_initialize_with_allocator`, but it should be there.

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -114,6 +114,7 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
     } else {
       fprintf(stderr, "Error getting env. variable "
         "RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED: %s\n", ret_str);
+      ret = RCUTILS_RET_INVALID_ARGUMENT;
     }
 
     // Check the environment variable for colorized output
--- a/src/logging.c
+++ b/src/logging.c
@@ -82,24 +82,6 @@ enum rcutils_colorized_output g_colorized_output = RCUTILS_COLORIZED_OUTPUT_AUTO
 
 rcutils_ret_t rcutils_logging_initialize(void)
 {
-  const char * line_buffered;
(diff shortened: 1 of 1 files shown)


src/logging.c
```diff
@@ -116,9 +98,27 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
     g_rcutils_logging_output_handler = &rcutils_logging_console_output_handler;
     g_rcutils_logging_default_logger_level = RCUTILS_DEFAULT_LOGGER_DEFAULT_LEVEL;
 
+    // Check the environment variable for line buffered output
+    const char * line_buffered;
+    const char * ret_str = rcutils_get_env("RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED", &line_buffered);
+
+    if (NULL == ret_str) {
+      if (!strcmp(line_buffered, "1")) {
```
dirk-thomas: An explicit comparison of the return value with a numeric (`== 0`) would be more readable imo and likely avoid the logic error from before to be repeated.

Similar below (`!= 0`).

ivanpauno:
CI (only fastrtps)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7160/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3372/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5863/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6990/)

PR #163, opened 2019-06-03
Fix armhf warning

emersonknapp:
Related to https://github.com/ros2/ros2/issues/721

Fixes these warnings https://ci.ros2.org/view/All/job/test_ci_linux-armhf/8/warnings23Result/package.-1643903456/

On armhf, size_t is 32 bits, so gcc raises warning for this line "comparison is always true due to limited range of data type"

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/src/error_handling.c
+++ b/src/error_handling.c
@@ -22,6 +22,7 @@ extern "C"
 #include <rcutils/error_handling.h>
 
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/src/error_handling.c
+++ b/src/error_handling.c
@@ -33,6 +33,10 @@ extern "C"
 // RCUTILS_REPORT_ERROR_HANDLING_ERRORS and RCUTILS_WARN_ON_TRUNCATION are set in the header below
 #include "./error_handling_helpers.h"
 
+#ifndef SIZE_MAX
+#define SIZE_MAX (size_t)-1
+#endif
+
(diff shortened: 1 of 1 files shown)


src/error_handling.c
```diff
@@ -33,6 +33,10 @@ extern "C"
 // RCUTILS_REPORT_ERROR_HANDLING_ERRORS and RCUTILS_WARN_ON_TRUNCATION are set in the header below
 #include "./error_handling_helpers.h"
 
+#ifndef SIZE_MAX
```
sloretz: include `<stdint.h>` for `SIZE_MAX`? On what platforms is it undefined?

emersonknapp: > include <stdint.h> for SIZE_MAX
yes, good catch

>  On what platforms is it undefined?
It's a C99 definition, I think the only place we don't get C99 is on Windows.

sloretz: It looks like VS 2017 and 2019 has `SIZE_MAX` in `<limits.h>`. https://docs.microsoft.com/en-us/cpp/c-runtime-library/data-type-constants?view=vs-2017 . Do you know if the "C Run-Time Library" is available when building C++ code on windows? If so, maybe the `ifndef` can be replaced with another include.

emersonknapp: I don't know offhand if it's available, but I can try a CI run on Windows without the `ifndef` to see if it's okay that way

emersonknapp:
Windows check for SIZE_MAX
[Build Status](https://ci.ros2.org/job/ci_windows/7108/)

PR #166, opened 2019-06-28
Implement rcutils_mkdir.

clalancette:
Needed for the spdlog logging backend.

Connected to ros2/rcl_logging#17

### DIFF
--- a/include/rcutils/filesystem.h
+++ b/include/rcutils/filesystem.h
@@ -144,6 +144,25 @@ rcutils_to_native_path(
   const char * path,
   rcutils_allocator_t allocator);
 
+/// Create the specified directory.
+/**
+ * This function creates an absolutely-specified directory.  If any of the
+ * intermediate directories do not exist, this function will return False.
+ * If the abs_path already exists, and is a directory, this function will
+ * return True.
+ *
+ * \param[in] abs_path
+ * \param[in] allocator
+ * \return bool True if making the directory was successful, False otherwise
+ *              False if path is NULL
+ *              False if path is empty
+ *              False if path is not absolute
+ *              False if any intermediate directories don't exist
+ */
+RCUTILS_PUBLIC
+bool
+rcutils_mkdir(const char * abs_path);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/rcutils/filesystem.h
+++ b/include/rcutils/filesystem.h
@@ -152,6 +152,9 @@ rcutils_to_native_path(
  * If the abs_path already exists, and is a directory, this function will
  * return True.
  *
+ * This function is not thread-safe due to mkdir races as described in the
+ * openat(2) documentation.
+ *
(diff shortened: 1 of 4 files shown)


clairewangosrf:
(going through ci tutorial, test launch ci)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7764/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3860/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6347/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7611/)

clalancette:
> Can you add some tests for this?

Done in 8075e9d

clalancette:
New CI is in the meta-issue: https://github.com/ros2/ros2/issues/789#issuecomment-536019068

include/rcutils/filesystem.h
```diff
@@ -144,6 +144,25 @@ rcutils_to_native_path(
...
+ *
+ * \param[in] abs_path
+ * \param[in] allocator
+ * \return bool True if making the directory was successful, False otherwise
+ *              False if path is NULL
+ *              False if path is empty
+ *              False if path is not absolute
+ *              False if any intermediate directories don't exist
+ */
+RCUTILS_PUBLIC
```
mjcarroll: Is is worth noting that this function is likely not thread safe?  I think most folks would assume that is the case, given the underlying call, but could probably bear repeating.

clalancette: Hm, interesting question.  It sort of depends on our definition of thread-safe.  It only uses stack variables and the parameter passed in, so there is no risk of memory corruption/memory races.  On the other hand, there is the directory race that `mkdir` is subject to (described in the openat documentation): http://man7.org/linux/man-pages/man2/open.2.html .  So from that perspective, it isn't thread-safe, but I'm not sure how we want to describe it.

include/rcutils/filesystem.h
```diff
@@ -144,6 +144,25 @@ rcutils_to_native_path(
   const char * path,
   rcutils_allocator_t allocator);
 
+/// Create the specified directory.
+/**
+ * This function creates an absolutely-specified directory.  If any of the
```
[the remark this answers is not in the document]
clalancette: Thanks, fixed in 412dafb

include/rcutils/filesystem.h
mjcarroll: I think leaving it at "This function is not thread safe" is probably suitable.  If someone is deeply concerned about it, then can consult the underlying code and documentation?

clalancette: I ended up adding that with a minor note to go look at the openat documentation for more information.

clalancette:
All right, merging this one in as a pre-requisite to the spdlog logging framework.  Thanks for the reviews.

PR #173, opened 2019-08-08
fixed #170

### LINKED ISSUE #170: Potential buffer overflow
Function call `rcutils_find(str + i, '}')` can returns 0, if a malformed function.
if so `size_t token_len = chars_to_end_delim - 1;` will cause token_len =  18446744073709551615 and cause overflow
https://github.com/ros2/rcutils/blob/2497ff5e28f749a6c55d060d55faf438bf224e64/src/logging.c#L634-L648

danielwangksu:
fix #170 
Signed-off-by: Daniel Wang <daniel.wang@canonical.com>

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -644,6 +644,7 @@ rcutils_ret_t rcutils_logging_format_message(
     }
 
     // Found what looks like a token; determine if it's recognized.
+    assert(chars_to_end_delim > 0);
     size_t token_len = chars_to_end_delim - 1;  // Not including delimiters.
     memcpy(token, str + i + 1, token_len);  // Skip the start delimiter.
     token[token_len] = '\0';


clalancette:
@danielwangksu I thought we had determined that this wasn't a problem over in #170.  Why the change, then?

danielwangksu:
> @danielwangksu I thought we had determined that this wasn't a problem over in #170. Why the change, then?

Hi @clalancette sorry. I check it again, it seems rcutils_findn defines i each time the function is called. So it might return 0

hidmic:
Hmm, though this change is innocuous, `rcutils_findn`, as used by `rcutils_find` can only return 0 if `delimiter` is found at the beginning of `str`, which by virtue of https://github.com/ros2/rcutils/issues/170#issuecomment-519567328 cannot happen. Not sure what sort of static analysis are you running @danielwangksu.

danielwangksu:
> Hmm, though this change is innocuous, `rcutils_findn`, as used by `rcutils_find` can only return 0 if `delimiter` is found at the beginning of `str`, which by virtue of [#170 (comment)](https://github.com/ros2/rcutils/issues/170#issuecomment-519567328) cannot happen. Not sure what sort of static analysis are you running @danielwangksu.

i might missed something, is it possible a malformed string without `{` and only has `}` at beginning?

hidmic:
That is indeed possible, but the code that precedes the call to `rcutils_find()` in `logging.c` handles that case.

danielwangksu:
> That is indeed possible, but the code that precedes the call to `rcutils_find()` in `logging.c` handles that case.

Thank you for clarifying it. I will close this issue.

PR #181, opened 2019-09-13
Always log on stderr

### LINKED ISSUE #168: INFO and ERROR logging should come out in order by default

rotu:
This follows the convention for human-readable diagnostic output on Linux systems.

fix #168

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -75,9 +75,6 @@ bool g_rcutils_logging_severities_map_valid = false;
 
 int g_rcutils_logging_default_logger_level = 0;
 
-bool g_force_stdout_line_buffered = false;
-bool g_stdout_flush_failure_reported = false;
-
 enum rcutils_colorized_output g_colorized_output = RCUTILS_COLORIZED_OUTPUT_AUTO;
 
 rcutils_ret_t rcutils_logging_initialize(void)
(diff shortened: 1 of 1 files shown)


clalancette:
I'm going to weigh in on this, I took a look around at some popular logging utilities in different languages to see what they did.  This is obviously non-exhaustive, but hopefully somewhat representative.

Python [logging](https://docs.python.org/3.8/library/logging.html) module - Prints all output, at all levels, to stderr by default
C++ [log4cxx](https://logging.apache.org/log4cxx/latest_stable/usage.html) - Using the "BasicConfigurator", all output, at all levels, goes to stdout by default
C++ [spdlog](https://github.com/gabime/spdlog) - Using whatever defaults it uses, all output, at all levels, goes to stdout by default
Ruby [logger](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html) module - Needs to be told whether to use stdout or stderr, but all levels go to that stream by default
Rust [log](https://docs.rs/log/0.4.8/log/) crate - All output at all levels goes to stderr by default
Perl [log4Perl](https://perlmaven.com/logging-with-log4perl-the-easy-way) - All output at all levels goes to stdout by default
Java [log4j](https://logging.apache.org/log4j/2.x/) - All output at all levels goes to stdout by default

Given that survey, it seems to me that ROS 2 should *definitely* choose a single stream for all logging output by default.  Given the results above, it actually doesn't seem that clear to me which stream the default should be.  I don't care hugely one way or the other, but I'd lean towards making the default stderr (as in this PR), as in the case of a real error, you want to make sure the message gets flushed out immediately.  But we should also have a way to switch to the stdout stream, as nearly all loggers surveyed above have a way to do that.

So with all of that said, I'm going to do a review of this PR.

src/logging.c
```diff
@@ -683,14 +661,14 @@ rcutils_ret_t rcutils_logging_format_message(
 # define IS_STREAM_A_TTY(stream) (isatty(fileno(stream)) != 0)
 #endif
 
-#define IS_OUTPUT_COLORIZED(is_colorized, stream) \
```
clalancette: Here and elsewhere, instead of removing the stream, let's leave it configurable and add a new environment variable to let users choose between stderr and stdout.  I don't care hugely what its called, but something descriptive would be nice.  In my opinion, if the environment variable is not specified, the default should be stderr.

rotu: @clalancette Why should it be reconfigurable here?

* If the user wants all program output in a single stream, that's just `2>&1`.
* If it's about performance, then we should instead change the `stderr` buffering with `setvbuf`.
* If we don't expect stream redirection to be available (or we want to make sure the user can differentiate from other types of program output), we should additionally allow the user to specify an arbitrary file, not just a stream.

clalancette: > @clalancette Why should it be reconfigurable here?

So the 2>&1 isn't that straightforward to achieve when you are launching through a launch file.  While it would be kind of weird to use stdout instead of stderr in one process of a launch file in not another, it isn't invalid.

The other reason to have it is that all of the log libraries I surveyed have the ability to choose which stream to use.  Given that they all have it, presumably it is functionality that is useful in a logging library, so I think we should probably follow suit here.

clalancette: (also, because there is a small performance impact from using stderr, we should give users the ability to opt out of it and get buffered performance if they want)

clalancette:
I did some additional work on this (including a rebase), but I'm not allowed to push to the branch here.  I'm going to close out this PR, and open a new one with my new changes and replicating the discussion here.

rotu:
Thank you for taking this over, @clalancette!

Issue #30, opened 2017-06-13
Implement a thread safe version of rcutils_get_env


mikaelarguedas:
While ensuring thread safety may be impossible we could improve robustness by using thread local storage for the result of the rcutils_get_env function.
This would also be a good time to refactor every part of the code reading environment variables to use this function
Related ticket https://github.com/ros2/rcl/pull/97

wjwwood:
It would be possible to make it thread-safe with a global mutex, but perhaps it would be better to avoid doing that in rcutils_getenv, but protect higher level functions that use it using a mutex. For example, maybe the "right thing" to do is to just mark rcl_create_node as not thread safe, and let rclpy/rclcpp/rclc make it thread safe.

It (probably) wouldn't hurt to also have a thread-safe getenv here, but I'm not sure it should be used in create_node, for example. It would potentially hurt the portability of this library to start requiring mutexs

clalancette:
Looking at rcutils_get_env() in more detail, it looks like there are at least 2 current problems with the thread-safety.  The first problem is obvious and has to do with the global __env_buffer[] on Windows, which is clearly not thread-safe.  The second problem is a bit more subtle.  Since getenv()/getenv_s() both operate on the statically allocated environment for the program, it could be the case that the string is changing (via putenv()/setenv()) while reading it.  The Windows [documentation](https://msdn.microsoft.com/en-us/library/tb2sfw2z.aspx) says as much.

One way to tackle the first problem is to make the contract for rcutils_get_env() such that the string returned is malloc'ed, and responsibility for freeing the memory falls on the caller.  This also mitigates much of the second problem, though doesn't completely remove it.  We could completely mitigate the second problem by adding the mutex, but we would also have to be careful about adding a mutex to any putenv()/setenv() callers.  Since the only callers of putenv()/setenv() are in either tests, or the OpenSplice implementation, my feeling is that we should change the contract here to a malloc'ed string, fix the callers, and go from there.  If we start mucking with the environment more in the future, we can add the mutex then.  Thoughts on that approach?

wjwwood:
I'd rather just make it not thread-safe, but returning a malloc'ed string would be ok I guess. I was think that you could just make the `__env_buffer` thread local (has it's own issues avoiding malloc).

If it does return a copy of the environment variable's value, it should take an allocator.

----

[...]

PR #182, opened 2019-09-17
Modify rcutils_getenv to be thread safe

ivanpauno:
Complements https://github.com/ros2/rcl/pull/502

Documentation about thread local storage:
- https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Thread-Local.html
- https://docs.microsoft.com/en-us/cpp/cpp/thread?view=vs-2019
- https://clang.llvm.org/docs/AttributeReference.html#id289

### DIFF
--- a/include/rcutils/thread.h
+++ b/include/rcutils/thread.h
@@ -1,0 +1,33 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__THREAD_H_
+#define RCUTILS__THREAD_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#ifndef _WIN32
+# define THREAD_LOCAL_STORAGE __thread
+#else
+# define THREAD_LOCAL_STORAGE __declspec(thread)
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCUTILS__THREAD_H_
--- a/src/get_env.c
+++ b/src/get_env.c
@@ -21,10 +21,11 @@ extern "C"
 #include <stdlib.h>
 
 #include "rcutils/get_env.h"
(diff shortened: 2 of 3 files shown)


include/rcutils/thread.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#ifndef _WIN32
+# define THREAD_LOCAL_STORAGE __thread
+#else
+# define THREAD_LOCAL_STORAGE __declspec(thread)
```
wjwwood: https://github.com/ros2/rcutils/blob/e52f2cfc1aac935bb0e6c35a7f3ad332a56d6d71/include/rcutils/macros.h#L33?

src/get_env.c
```diff
@@ -21,10 +21,11 @@ extern "C"
 #include <stdlib.h>
 
 #include "rcutils/get_env.h"
+#include "rcutils/thread.h"
 
 #ifdef _WIN32
 # define WINDOWS_ENV_BUFFER_SIZE 2048
-static char __env_buffer[WINDOWS_ENV_BUFFER_SIZE];
+THREAD_LOCAL_STORAGE static char __env_buffer[WINDOWS_ENV_BUFFER_SIZE];
```
wjwwood: This may cause memory allocations where before they did not occur, see:

https://github.com/ros2/rcutils/blob/e70143a537a9e1540f77cdb74848346cd11f6c59/include/rcutils/error_handling.h#L101-L141

include/rcutils/thread.h
ivanpauno: It's the windows specific way of declaring a variable thread local. See https://docs.microsoft.com/en-us/cpp/cpp/thread?view=vs-2019.

ivanpauno: Oh, I see we already had this

wjwwood:
`getenv_s` requires that you use your own storage, whereas `getenv` returns just a pointer to it.

Also, neither is thread-safe as far as I know:

> This function is not required to be thread-safe. Another call to getenv, as well as a call to the POSIX functions setenv(), unsetenv(), and putenv() may invalidate the pointer returned by a previous call or modify the string obtained from a previous call.

https://en.cppreference.com/w/c/program/getenv

I'd recommend reading all of https://github.com/ros2/rcutils/issues/30 if you haven't already.

hidmic:
Alright, I see now. And I think the existing one is the best solution. In light of this, should #30 be closed as `won't do`?

wjwwood:
> In light of this, should #30 be closed as won't do?

I'm not sure. I think it's definitely not needed now, but maybe others have an opinion about it. We could close it or leave it open, doesn't matter to me. If we close it we can always reopen it.

clalancette:
I think in light of what we discussed in https://github.com/ros2/rclcpp/pull/987#issuecomment-591559842 , we should close this out in favor of a new PR.  I'm going to go ahead and do that, but feel free to re-open if you disagree.

PR #183, opened 2019-10-01
Make use of time source type for throttling logs

BMarchi:
Part of [797](https://github.com/ros2/rclcpp/issues/797). Gives meaning to the time_source_type argument to throttle macros.

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8497/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4382/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8350/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/6913/)


test/test_logging_macros.cpp
```diff
@@ -154,7 +154,7 @@ TEST_F(TestLoggingMacros, test_logging_throttle) {
     auto is_before_throttle_end =
       ((std::chrono::system_clock::now() - start) < throttle_time);
     RCUTILS_LOG_ERROR_THROTTLE(
-      RCUTILS_STEADY_TIME, throttle_time.count(), first ? "first" : "other");
+      rcutils_steady_time_now, throttle_time.count(), first ? "first" : "other");
```
hidmic: @BMarchi nit: why not keeping function behind the macro? `RCUTILS_STEADY_TIME` reads better.

rcutils/logging.py
```diff
@@ -54,7 +54,8 @@
 skipfirst_doc_lines = [
     'The first log call is being ignored but all subsequent calls are being processed.']
 throttle_params = OrderedDict((
-    ('time_source_type', 'The time source type of the time to be used'),
+    ('time_source_type', 'Function that returns rcutils_ret_t and expects a '
+        'rcutils_time_point_value_t pointer.'),
```
hidmic: @BMarchi nit: I wonder if `time_source` makes more sense than `time_source_type`. @dirk-thomas thoughts?

rcutils/logging.py
```diff
@@ -54,7 +54,8 @@
 skipfirst_doc_lines = [
     'The first log call is being ignored but all subsequent calls are being processed.']
 throttle_params = OrderedDict((
-    ('time_source_type', 'The time source type of the time to be used'),
+    ('time_source_type', 'Function that returns rcutils_ret_t and expects a '
```
dirk-thomas: How about renaming this parameter to match the new semantic, e.g. `get_time_point_value`?

hidmic: That goes against https://github.com/ros2/rcutils/pull/183#discussion_r330247026. Do we want to be explicit about the parameter being a function or do we want to hide it for the sake of API ergonomics? I'm slightly inclined towards the latter but it's not a strong stand.

dirk-thomas: The function parameter is being used as a function:

> if (time_source_type(&__rcutils_logging_now) != RCUTILS_RET_OK) {

There is no way you can just pass `RCUTILS_STEADY_TIME` (assuming it isn't a macro defining a function).

rcutils/logging.py
dirk-thomas: Well, "time source" has a very specific meaning in `rclcpp` and since this is not the same I would avoid using the same term for it. See my other comment for a suggestion.

BMarchi: So I should call it then `get_time_point_value`?

rcutils/logging.py
hidmic: That's precisely what I'm proposing, an alias for that function so that it reads better. But it's a detail and orthogonal to the actual macro parameter name.

hidmic: And that's what @BMarchi implemented in https://github.com/ros2/rcutils/pull/183/commits/e1b4344f90c79523908ece4454399e5743585eea. WDYT @dirk-thomas ?

dirk-thomas: My previous comment still applies: it is not a time source so it should not be named `time_source` imo.

rcutils/logging.py
```diff
@@ -168,7 +169,7 @@ def __init__(self, *, params=None, args=None, doc_lines=None):
             }, **name_args
         },
         doc_lines=skipfirst_doc_lines + throttle_doc_lines + name_doc_lines)),
-))
+    ))
```
dirk-thomas: This looks like an unrelated and unnecessary change?

include/rcutils/time.h
```diff
@@ -39,6 +39,8 @@ extern "C"
 #define RCUTILS_NS_TO_MS(nanoseconds) (nanoseconds / (1000LL * 1000LL))
 /// Convenience macro to convert nanoseconds to microseconds.
 #define RCUTILS_NS_TO_US(nanoseconds) (nanoseconds / 1000LL)
+/// Convenience macro to reference the function for system time.
```
hidmic: @BMarchi nit: ref that function

include/rcutils/time.h
```diff
@@ -39,6 +39,8 @@ extern "C"
 #define RCUTILS_NS_TO_MS(nanoseconds) (nanoseconds / (1000LL * 1000LL))
 /// Convenience macro to convert nanoseconds to microseconds.
 #define RCUTILS_NS_TO_US(nanoseconds) (nanoseconds / 1000LL)
+/// Convenience macro for rcutils_steady_time_now(rcutils_time_point_value_t *).
+#define RCUTILS_STEADY_TIME rcutils_steady_time_now
```
wjwwood: Is this being used?

BMarchi: Yes, it's used by the [tests](https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/test/test_logging_macros.cpp#L157).
It was there by default, before these changes.

wjwwood: I don't understand, how was it being set before, and why are you setting it here now? I only see this symbol (`RCUTILS_STEADY_TIME`) in one place in this pull request. Shouldn't remove it somewhere else or use it somewhere else for that to make sense?

hidmic: @wjwwood as @BMarchi mentioned earlier, it is used by tests:
https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/test/test_logging_macros.cpp#L156-L157

Up until this patch, `RCUTILS_STEADY_TIME` was undefined. That wasn't a problem because the throttled logging macro implementation ignored the corresponding macro argument (`time_source_type`) completely:

https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/resource/logging_macros.h.em#L176-L197

@BMarchi made that define expand to the right function to provide a nice API and keep those tests passing.

wjwwood: Oh, I had no idea this worked...

```
#define FOO(arg1, arg2, arg3) \
  arg2 = arg3;

int main()
{
  int a, b = 1;
  FOO(DOES_NOT_EXIST, a, b)
  return 0;
}
```

That's awful, haha.

----

So then why is there no system time equivalent?

Also, why do you need a macro here, why couldn't you just pass `rcutils_steady_time_now`?

BMarchi: @hidmic suggested on keep using the [macro](https://github.com/ros2/rcutils/pull/183#discussion_r330247026), so I just left it there

wjwwood: That's fine, though I still wonder if a system time version makes sense.

I don't think it should block this.

PR #186, opened 2019-11-10
Use Win32 wrapper around 64 bit atomic operations

theseankelly:
64 bit comiler intrinsics are unavailable on x86 and ARM platforms in
MSVC. This change uses the Win32/UWP wrapping API for the 64 bit
atomics. Per documentation on MSDN, the functions will leverage compiler
intrinsics if available. All other intrinsics used in this file are available on x64, x86, and ARM versions of MSVC and pose no issue.

I've tested this fix against Dashing & Crystal as well -- please consider for patches.


sloretz:
CI (windows, testing all packages up to `rclpy` to test all code using `rcutils` atomics)
* [Build Status](https://ci.ros2.org/job/ci_windows/8550/)

theseankelly:
Thank you!

PR #187, opened 2019-11-10
Fix uninitialized handle error

theseankelly:
When compiling for UWP, MSVC flags the uninitialized handles with error
C4703.

I've also tested for Crystal and Dashing, please consider for patches

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -742,7 +742,7 @@ rcutils_ret_t rcutils_logging_format_message(
   }
 # define SET_OUTPUT_COLOR_WITH_SEVERITY(status, severity, stream, output_array) \
   { \
-    WORD color; \
+    WORD color = COLOR_NORMAL; \
     HANDLE handle = INVALID_HANDLE_VALUE; \
     SET_COLOR_WITH_SEVERITY(status, severity, color) \
     GET_HANDLE_FROM_STREAM(status, handle, stream) \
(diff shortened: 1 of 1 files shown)


src/logging.c
```diff
@@ -743,15 +743,15 @@ rcutils_ret_t rcutils_logging_format_message(
 # define SET_OUTPUT_COLOR_WITH_SEVERITY(status, severity, stream, output_array) \
   { \
     WORD color; \
```
Karsten1987: When the `HANDLE` variable is being complained about, should `WORD` also be default initialized then? It would make sense to me to initialize it to [COLOR_NORMAL](https://github.com/ros2/rcutils/blob/0b47fc3f2e9598e3921bf0ce1ccd0d3e5a24d400/src/logging.c#L673)

theseankelly: The error in question (C4703) complains about uninitialized pointers/handles specifically. The uninitialized `color` isn't a problem. Though, say the word, and I'm happy to overload this PR and initialize it

Karsten1987: I would appreciate if you could initialized the color as well.

theseankelly: done; sorry for delay.

sloretz:
CI (testing just rcutils)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8640/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4483/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7047/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8559/)

theseankelly:
Thanks! Can this be marked a dashing backport as well?

Karsten1987:
see https://github.com/ros2/rcutils/pull/191

theseankelly:
Thanks @Karsten1987 !!

Issue #189, opened 2019-11-13
Default logging format design


SteveMacenski:
Was there a design discussion around the default logging format defined here? https://github.com/ros2/rcutils/blob/ff8ca5803d32ee7f91cd50c99cf36741f104c681/src/logging.c#L65

I'd think the default you'd want a unix timestamp to determine relative timing of when events or logs occur. I can submit a PR to change this is folks agree. I can say from experience working with Navigation2 that this has been a pain point, which I know can be worked around setting `RCUTILS_CONSOLE_OUTPUT_FORMAT` however I'd think a timestamp is a reasonable default behavior.

fujitatomoya:
@SteveMacenski 

i totally do agree with you.

SteveMacenski:
#190

PR #190, opened 2019-11-13
changing default logging format to include timestamp

SteveMacenski:
Covers https://github.com/ros2/rcutils/issues/189

Adds a timestamp to the default logging format. I think this is a reasonable default behavior.

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -62,7 +62,8 @@ enum rcutils_colorized_output
 bool g_rcutils_logging_initialized = false;
 
 char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
-static const char * g_rcutils_logging_default_output_format = "[{severity}] [{name}]: {message}";
+static const char * g_rcutils_logging_default_output_format =
+  "[{time}]: [{severity}] [{name}]: {message}";
 
 static rcutils_allocator_t g_rcutils_logging_allocator;
 
--- a/src/logging.c
+++ b/src/logging.c
@@ -63,7 +63,7 @@ bool g_rcutils_logging_initialized = false;
 
 char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
 static const char * g_rcutils_logging_default_output_format =
-  "[{time}]: [{severity}] [{name}]: {message}";
+  "[{severity}] [{time}] [{name}]: {message}";
 
 static rcutils_allocator_t g_rcutils_logging_allocator;
 


wjwwood:
Does anyone have time to summarize how it works in ROS 1? I want to say that the default for the console did not include the timestamps, but the timestamps were included in the file logging. So you could reduce clutter on the console (timestamp is slightly less necessary when you're observing it live), but you could always go to the file log for exact timestamps.

SteveMacenski:
>  I want to say that the default for the console did not include the timestamps

ROS1 did have timestamps to screen. I've stared at alot of log files at Simbe :laughing:

wjwwood:
> ROS1 did have timestamps to screen.

👍 Yeah, it's fine with me to add the timestamp by default.

src/logging.c
```diff
@@ -62,7 +62,8 @@ enum rcutils_colorized_output
 bool g_rcutils_logging_initialized = false;
 
 char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
-static const char * g_rcutils_logging_default_output_format = "[{severity}] [{name}]: {message}";
+static const char * g_rcutils_logging_default_output_format =
+  "[{time}]: [{severity}] [{name}]: {message}";
```
wjwwood: Looking at the ROS 1 style, would `[{severity}] [{time}] [{name}]: {message}` make more sense?

SteveMacenski:
Good point, done

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8707/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4541/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7111/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8625/)

SteveMacenski:
> assuming CI is good

If I've managed to mess up something to break CI with a 1 line string change, feel free to put me out to pasture

wjwwood:
> If I've managed to mess up something to break CI with a 1 line string change, feel free to put me out to pasture

It wouldn't be the first time for me. :)

My only real thought is that it might break some (imo fragile) tests that's expecting a certain output from the logger or something like that.

SteveMacenski:
aaarg. Linux failed.

dirk-thomas:
I would guess you need to rebase you branch on top of the latest state of the target branch.

SteveMacenski:
```
16:13:14 /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rclcpp/rclcpp/resource/logging.hpp.em:80: error: <class 'TypeError'>: can only concatenate tuple (not "str") to tuple
16:13:14 CMakeFiles/rclcpp.dir/build.make:61: recipe for target 'include/rclcpp/logging.hpp' failed
```

Lets try that.

SteveMacenski:
Yal are squashing this anyhow. Just pulled in.

wjwwood:
Yeah, it's really a limitation of our CI when testing forks (it doesn't try to merge with target branch).

New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8709/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4543/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7113/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8627/)

SteveMacenski:
**phew** saved by an updated target. I'm safe from the farm upstate for another day...

wjwwood:
Thanks @SteveMacenski!

dirk-thomas:
This should be documented on the release page (currently for F-turtle) to mentioned the change.

wjwwood:
Good point, I'll keep this tab open until we've announced the new name and have an F-turtle release page on ros2/ros2_documentation.

wjwwood:
https://github.com/ros2/ros2_documentation/pull/446

ivanpauno:
This caused some test failures, like the one corrected in https://github.com/ros2/ros2cli/pull/416. Also:

https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1705/testReport/junit/test_rclcpp/TestExecutableOutputAfterShutdown/test_signal_handler_before_shutdown/
https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1705/testReport/junit/(root)/projectroot/test_logging_demo__rmw_connext_cpp/

and maybe others.

wjwwood:
@SteveMacenski do you have time to look at the above test regressions that @ivanpauno pointed out?

I'll have a look next week if you can't, but we'd appreciate it.

SteveMacenski:
@wjwwood I hadn't looked at it. 

Maybe I'm missing the punchline, but I don't see anything in those links that makes me think there's an issue, can you point out specific lines or what you're referencing? 

I'll help fix issues from logging changes, but I'm pretty shocked there are tests created based on a logging format

wjwwood:
It's checking the output of the commands and they don't match anymore.

> can you point out specific lines or what you're referencing?

https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1705/testReport/junit/test_rclcpp/TestExecutableOutputAfterShutdown/test_signal_handler_before_shutdown/

That link has an assert error that the expected output did not match the actual output.

> I'll help fix issues from logging changes, but I'm pretty shocked there are tests created based on a logging format

Well, it could be improved by the tests manually setting the log output to ensure it doesn't deviate due to changes in the system default. Or the output matching can be relaxed to match any logging format (it can use regex).

SteveMacenski:
```
AssertionError: Did not find '['Registered custom signal handler.', 'Called rclcpp::init.', 'Waiting to give an opportunity for interrupt...', '[INFO] [rclcpp]: signal_handler(signal_value=2)', 'Custom sigint handler called.', 'Calling rclcpp::shutdown...', 'Called rclcpp::shutdown.', 'Waiting to give an opportunity for interrupt...', 'Exiting.']' in output for any of the matching processes: test_signal_handler-1
```

Thats... a little esoteric to pick up on the fact that that's a formatting error :laughing:. I'll add to my queue to look at next week.

nuclearsandwich:
I started updating some tests in https://github.com/ros2/demos/pull/420. I chose to specify a logging format in the test to prevent future changes to the defaults from breaking the tests again.

SteveMacenski:
OK so traced down [that link](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1705/testReport/junit/test_rclcpp/TestExecutableOutputAfterShutdown/test_signal_handler_before_shutdown/) and looks like @nuclearsandwich got to it [here](https://github.com/ros2/demos/pull/420).

First time pushing something this low into the ROS2 stack. Trying to think about future PRs how I can avoid these issues or fix things proactively on failure rather than making yal come to the rescue -- so where was the starting point that led you to problems being introduced here? If I know where to look, I can follow through down to these. I'm assuming it started with the nightly build, I see a bunch of options (release, debug, repeated, with specific RMWs) is there one you're looking at?

nuclearsandwich:
I have slightly different workflows depending on the situation but if a failure is in an environment I can easily reproduce like nightly_linux_{debug,release} I'll usually start there since I can work locally as well as with CI. https://index.ros.org/doc/ros2/Contributing/Build-Cop-and-Build-Farmer-Guide/ is our doc on being the build farmer but I can't recall when it was last touched. It would be a good idea for us to document the current nightly platforms and job variants for contributors.

---

The test failures weren't caught by Pull request CI because of the tight scoping for the branch builds to limit resource usage and time spent on CI. We try to limit CI to packages which are directly affected by the change. Because of the way tests were written, more packages were affected than tested.

SteveMacenski:
Got it, thanks!

PR #194, opened 2019-12-03
Split visibility macro project independent logic

thomas-moulard:
Until now, all ROS 2 packages had a long visibility_control.h header
which is a copy-and-paste of the rcutils one.

This commit reduces the size of visibility_control.h by moving out into
a new header called visibility_control_macros.h the macros re-usable by
other packages.

### DIFF
--- a/include/rcutils/visibility_control_macros.h
+++ b/include/rcutils/visibility_control_macros.h
@@ -1,0 +1,65 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__VISIBILITY_CONTROL_MACROS_H_
+#define RCUTILS__VISIBILITY_CONTROL_MACROS_H_
+
+// Defines macros to express whether a symbol is localed, imported, or exported
+//
+// Those macros are compatible with GCC, clang, and Microsoft Visual C++. They
+// can be used to enforce which symbols of a library are publicly accessible.
+//
+// RCUTILS_IMPORT, RCUTILS_EXPORT, and RCUTILS_LOCAL are respectively declaring
+// an imported, exported, or local symbol.
+// RCUTILS_LOCAL can be used directly. However, RCUTILS_IMPORT, and
+// RCUTILS_EXPORT may not be used directly. Every project need to provide
+// an additional header called `visibility_macros.h` containing:
+//
+// #ifdef <project>_BUILDING_DLL
+// # define <project>_PUBLIC RCUTILS_EXPORT
+// #else
+// # define <project>_PUBLIC RCUTILS_IMPORT
+// #endif // !<project>_BUILDING_DLL
+// #define <project>_LOCAL RCUTILS_LOCAL
+//
+// ...where "<project>" has been replaced by the project name, such as
+// "MY_PROJECT".
+// Your project CMakeLists.txt should also contain the following statement:
+//
+// target_compile_definitions(<your library> PRIVATE "<project>_BUILDING_DLL")
+//
+// A public (exported) class should then be tagged as <project>_PUBLIC, whereas
+// a non-public class should be tagged with <project>_LOCAL.
+//
+// See GCC documentation: https://gcc.gnu.org/wiki/Visibility
+
+#if defined(_MSC_VER) || defined(__CYGWIN__)
+// Use the Windows syntax when compiling with Microsoft Visual C++.
+//
+// GCC on Windows also support this syntax. When compiling with cygwin,
+// prefer using dllimport/dllexport are the semantic of those flags is closer
+// to msvc++ behavior. See GCC documentation:
+// https://gcc.gnu.org/onlinedocs/gcc/Microsoft-Windows-Function-Attributes.html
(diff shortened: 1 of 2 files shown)


thomas-moulard:
@ros2/aws-oncall @prajakta-gokhale @thomas-moulard - please run this CI job
Gist: https://gist.githubusercontent.com/thomas-moulard/81405cb16f40134822a2b6fbc432e3a5/raw/cfa55d79b52f7d2a59b332fd5b91421eb965de01/ros2.repos
BUILD args: --packages-up-to rcutils
TEST args: --packages-select rcutils
Job: ci_launcher

thomas-moulard:
I checked this PR is a no-op by listing `rcutils.so` exported symbols before/after this PR: `nm -D build/rcutils/librcutils.so > ~/[before|after].txt`

[after.txt](https://github.com/ros2/rcutils/files/3913955/after.txt)
[before.txt](https://github.com/ros2/rcutils/files/3913956/before.txt)

prajakta-gokhale:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8779/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4601/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7181/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8698/)

thomas-moulard:
@prajakta-gokhale Could you re-run a CI job? `uncrustify` should be ok now.

prajakta-gokhale:
CI rebuild:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8781/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4603/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7183/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8700/)

zmichaels11:
@tfoote please look at this

Karsten1987:
I commented the same thing on the respective discourse post, but I think the trade-off between having to modify less code vs introducing a new dependency for every package to `rcutils` is not justified in my opinion. 

from discourse (https://discourse.ros.org/t/add-visibility-control-h-to-the-api-as-standard-usage/11681/14?u=karsten):
> While your solution certainly works to deduplicate code, this brings another burden to the user. Now, every package has to depend on `rcutils`. In order to use the macro, every package has to set the dependencies straight in their `package.xml` and `CMakeLists.txt`.
The current situation is not ideal right now, but it's essentially one search-and-replace command to get the visibility macros to compile with your package.
One thing to mention is that `ros2 pkg create` [creates](https://github.com/ros2/ros2cli/blob/master/ros2pkg/ros2pkg/resource/cpp/visibility_control.h.em) that visibility header for you. So there is actually no need to copy paste anything in the first place when starting a new project.

emersonknapp:
Don't all ros2 packages have a dependency on rcutils already via rcl?

thomas-moulard:
@Karsten1987 I'm not sure what the concern is as it is an opt-in mechanism. The objective is not to enforce that on everyone. If one would rather not depend on `rcutils`, keeping the full header as it is today is possible.

Karsten1987:
I am not feeling super strongly about this. As I mentioned, the PR in itself certainly helps to reduce code duplication. Therefore, I am okay with this PR.

I am really just mentioning this as I feel that there should ideally only be one best-practice solution which works for all users. IMO, generating this file via `ros2 pkg create` does exactly this as users don't have to touch the visibility macros even once. But that's just my two cents.

jacobperron:
Are there plans change the output of `ros2 pkg create` to use the proposed change in this PR?

include/rcutils/visibility_control_macros.h
```diff
@@ -0,0 +1,65 @@
...
+// ...where "<project>" has been replaced by the project name, such as
+// "MY_PROJECT".
+// Your project CMakeLists.txt should also contain the following statement:
+//
+// target_compile_definitions(<your library> PRIVATE "<project>_BUILDING_DLL")
+//
+// A public (exported) class should then be tagged as <project>_PUBLIC, whereas
+// a non-public class should be tagged with <project>_LOCAL.
+//
+// See GCC documentation: https://gcc.gnu.org/wiki/Visibility
```
wjwwood: If you use `///` for this comment block, doxygen will pick it up as file documentation, which would be good, imo.

piraka9011: @wjwwood Done, could you take a look at this an lmk if it's good to merge?

wjwwood: This nitpick looks ok. But it would be good to have another approval from people who commented, and to have responses to questions asked, like this one:

https://github.com/ros2/rcutils/pull/194#issuecomment-570312699

piraka9011:
> Are there plans change the output of `ros2 pkg create` to use the proposed change in this PR?

Ditto Karsten's comment, there are no plans to change `ros2 pkg create` after these changes.

zmichaels11:
Are there any updates on this PR?

wjwwood:
I think it just needs CI to be merged. Are there connected prs that need review or something?

zmichaels11:
There haven't been any more commits since the last CI.
I can rerun it if you want.

wjwwood:
https://github.com/ros2/rcutils/pull/194/commits/6f6114eae733dc1379c6cad9ece267ec5f666acf is, is it not? I guess it could be trivial, but I didn't look at the content, I just assume if there's a new commit it needs new CI. Also, if it has been a while any thing merged to master could have broken CI, so "stale" is also a decent reason to re-run it.

I guess no one else will comment on it, then please do run CI again, and then I'll merge it.

zmichaels11:
Oh wow, I missed that one. Thanks for pointing it out.
I'll rerun the CI.

zmichaels11:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9138/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4860/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7462/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9060/)

PR #195, opened 2019-12-03
Feature/add security directory

BorjaOuterelo:
Fix https://github.com/ros2/rcl/issues/545 along with PR https://github.com/ros2/rcl/pull/544


jacobperron:
See my comments on the referenced ticket https://github.com/ros2/rcl/issues/545 about why I don't think moving the functions to `rcutils` is the right approach.

jacobperron:
Closing based on my above comment.

PR #196, opened 2019-12-13
All logging to the same stream

### LINKED ISSUE #168: INFO and ERROR logging should come out in order by default

clalancette:
This PR is a rehash of the now-closed #181.  In particular, this PR proposes to make it so that all output from all logging levels go to the same stream.  This replicates the behavior of basically all other logging systems; see https://github.com/ros2/rcutils/pull/181#issuecomment-546553594 for details.

Compared to #181, this PR is now rebased on master, and has made it so that the stream that all output goes to is configurable.  The default is stderr/unbuffered (so that error messages go out right away), but if users want a very minor performance improvement, they can switch it to stdout/buffered by setting `RCUTILS_CONSOLE_LINE_BUFFERED=1`.

Unfortunately, the change to stderr by default means that a lot of tests in the ROS 2 core (and probably beyond) are going to start failing.  That's because those tests are typically trying to capture stdout output, and there is no longer any output there.  Solutions I can think of:

1.  Make the default stream stdout (failing tests isn't a great reason to do this, in my opinion).
1.  Change all of the failing tests to set `RCUTILS_CONSOLE_LINE_BUFFERED=1` (this will fix it in the core, but still leave downstream consumers with failing tests).
1.  Change launch_testing to capture stderr along with stdout.
1.  ???

@ros2/team Opinions on how to proceed with fixing tests appreciated.

@rotu FYI

Fixes #168

### DIFF
--- a/include/rcutils/strerror.h
+++ b/include/rcutils/strerror.h
@@ -37,7 +37,7 @@ extern "C"
  */
 RCUTILS_PUBLIC
 void
-rcutils_safe_strerror(char * buffer, size_t buffer_length);
+rcutils_strerror(char * buffer, size_t buffer_length);
 
 #ifdef __cplusplus
 }
--- a/include/rcutils/strerror.h
+++ b/include/rcutils/strerror.h
@@ -1,0 +1,33 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__STRERROR_H_
+#define RCUTILS__STRERROR_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcutils/visibility_control.h"
+
+RCUTILS_PUBLIC
+void
+rcutils_safe_strerror(char * buffer, size_t buffer_length);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCUTILS__STRERROR_H_
--- a/include/rcutils/strerror.h
+++ b/include/rcutils/strerror.h
@@ -23,7 +23,8 @@ extern "C"
 #include "rcutils/visibility_control.h"
 
 /// Retrieve the string corresponding to the last system error.
-/** This function retrieves the value of errno, and calls the system-specific
+/**
+ * This function retrieves the value of errno, and calls the system-specific
  * equivalent of `strerror` on it, storing the output in the provided buffer.
  * If the error message is longer than the buffer, it will be truncated.
  * The memory for the c-string buffer that is passed in must be managed by the
--- a/include/rcutils/strerror.h
+++ b/include/rcutils/strerror.h
@@ -1,4 +1,4 @@
-// Copyright 2017 Open Source Robotics Foundation, Inc.
+// Copyright 2020 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,6 +22,18 @@ extern "C"
 
 #include "rcutils/visibility_control.h"
 
+/// Retrieve the string corresponding to the last system error.
+/* This function retrieves the value of errno, and calls the system-specific
+ * equivalent of `strerror` on it, storing the output in the provided buffer.
+ * If the error message is longer than the buffer, it will be truncated.
+ * The memory for the c-string buffer that is passed in must be managed by the
+ * caller.
+ *
+ * This function is thread-safe.
+ *
+ * \param[in/out] buffer the buffer in which to store the data
+ * \param[in] buffer_length the maximum length of the buffer
+ */
 RCUTILS_PUBLIC
 void
 rcutils_safe_strerror(char * buffer, size_t buffer_length);
--- a/include/rcutils/strerror.h
+++ b/include/rcutils/strerror.h
@@ -23,7 +23,7 @@ extern "C"
 #include "rcutils/visibility_control.h"
 
 /// Retrieve the string corresponding to the last system error.
-/* This function retrieves the value of errno, and calls the system-specific
+/** This function retrieves the value of errno, and calls the system-specific
  * equivalent of `strerror` on it, storing the output in the provided buffer.
  * If the error message is longer than the buffer, it will be truncated.
  * The memory for the c-string buffer that is passed in must be managed by the
--- a/src/logging.c
+++ b/src/logging.c
@@ -18,6 +18,7 @@ extern "C"
 #endif
 
 #include <ctype.h>
+#include <errno.h>
 #include <inttypes.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -77,9 +78,6 @@ bool g_rcutils_logging_severities_map_valid = false;
 int g_rcutils_logging_default_logger_level = 0;
 
 static FILE * g_output_stream = NULL;
-static bool g_output_flush_stream = false;
-// A one-time print if we fail to flush
-static bool g_stream_flush_failure_reported = false;
 
 enum rcutils_colorized_output g_colorized_output = RCUTILS_COLORIZED_OUTPUT_AUTO;
 
@@ -88,11 +86,20 @@ rcutils_ret_t rcutils_logging_initialize(void)
   return rcutils_logging_initialize_with_allocator(rcutils_get_default_allocator());
 }
 
+enum rcutils_get_env_retval
+{
+  RCUTILS_GET_ENV_ERROR = -1,
+  RCUTILS_GET_ENV_ZERO = 0,
+  RCUTILS_GET_ENV_ONE = 1,
+  RCUTILS_GET_ENV_DEFAULT = 2,
+};
+
 // A utility function to get zero or one from an environment variable.  Returns
-// -1 if we failed to get the environment variable, 0 if the value in the
-// environment variable is something other than "0" or "1", 0 if the environment
-// variable is "0" or "", or "1" if the environment variable is "1".
-static int rcutils_get_env_var_zero_or_one(
+// RCUTILS_GET_ENV_ERROR if we failed to get the environment variable or if it was
+// something we don't understand.  Return RCUTILS_GET_ENV_ZERO if the value in the
+// environment variable is "0", RCUTILS_GET_ENV_ONE if the value in the environment
+// variable is "1", or RCUTILS_GET_ENV_DEFAULT if the environment variables is empty.
+static enum rcutils_get_env_retval rcutils_get_env_var_zero_or_one(
   const char * name, const char * zero_semantic,
   const char * one_semantic)
 {
@@ -102,24 +109,27 @@ static int rcutils_get_env_var_zero_or_one(
     fprintf(
       stderr, "Error getting environment variable "
       "%s: %s\n", name, ret_str);
-    return -1;
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("Error getting environment variable: %s", ret_str);
+    return RCUTILS_GET_ENV_ERROR;
   }
 
-  if (strcmp(env_var_value, "0") == 0 || strcmp(env_var_value, "") == 0) {
-    return 0;
+  if (strcmp(env_var_value, "") == 0) {
+    return RCUTILS_GET_ENV_DEFAULT;
   }
-
+  if (strcmp(env_var_value, "0") == 0) {
+    return RCUTILS_GET_ENV_ZERO;
+  }
   if (strcmp(env_var_value, "1") == 0) {
-    return 1;
+    return RCUTILS_GET_ENV_ONE;
   }
 
   fprintf(
     stderr,
     "Warning: unexpected value [%s] specified for %s. "
-    "Valid values are 0 (%s) or 1 (%s).  Default value 0 will be used.\n",
+    "Valid values are 0 (%s) or 1 (%s).",
     env_var_value, name, zero_semantic, one_semantic);
 
-  return 0;
+  return RCUTILS_GET_ENV_ERROR;
 }
 
 rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allocator)
@@ -139,37 +149,59 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
     // are propagated immediately.  The user can choose to set the output stream
     // to stdout by setting the RCUTILS_CONSOLE_USE_STDOUT environment
     // variable to 1.
-    g_output_stream = stderr;
-    int retval = rcutils_get_env_var_zero_or_one(
+    enum rcutils_get_env_retval retval = rcutils_get_env_var_zero_or_one(
       "RCUTILS_CONSOLE_USE_STDOUT", "use stderr",
       "use stdout");
-    if (retval < 0) {
+    if (retval == RCUTILS_GET_ENV_ERROR) {
       return RCUTILS_RET_INVALID_ARGUMENT;
-    } else if (retval == 0) {
+    } else if (retval == RCUTILS_GET_ENV_DEFAULT || retval == RCUTILS_GET_ENV_ZERO) {
       g_output_stream = stderr;
-    } else if (retval == 1) {
+    } else if (retval == RCUTILS_GET_ENV_ONE) {
       g_output_stream = stdout;
     }
 
-    retval = rcutils_get_env_var_zero_or_one("RCUTILS_CONSOLE_FLUSH_STREAM", "no flush", "flush");
-    if (retval < 0) {
+    // Allow the user to choose how buffering on the stream works by setting
+    // RCUTILS_CONSOLE_BUFFERED_STREAM.  With an empty environment variable, use the
+    // default of the stream.  With a value of 0, force the stream to be unbuffered.
+    // With a value of 1, force the stream to be buffered.
+    retval =
+      rcutils_get_env_var_zero_or_one(
+      "RCUTILS_CONSOLE_BUFFERED_STREAM", "not buffered",
+      "buffered");
+    if (retval == RCUTILS_GET_ENV_ERROR) {
       return RCUTILS_RET_INVALID_ARGUMENT;
-    } else if (retval == 0) {
-      g_output_flush_stream = false;
-    } else if (retval == 1) {
-      g_output_flush_stream = true;
+    } else if (retval == RCUTILS_GET_ENV_ZERO) {
+      if (setvbuf(g_output_stream, NULL, _IONBF, 0) != 0) {
+        fprintf(
+          stderr, "Error setting stream buffering mode: %s\n", strerror(errno));
+        RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+          "Error setting stream buffering mode: %s", strerror(
+            errno));
+        return RCUTILS_RET_ERROR;
+      }
+    } else if (retval == RCUTILS_GET_ENV_ONE) {
+      if (setvbuf(g_output_stream, NULL, _IOLBF, 0) != 0) {
+        fprintf(
+          stderr, "Error setting stream buffering mode: %s\n", strerror(errno));
+        RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+          "Error setting stream buffering mode: %s", strerror(
+            errno));
+        return RCUTILS_RET_ERROR;
+      }
     }
+    // Note that for retval == RCUTILS_GET_ENV_DEFAULT, we do nothing to leave it at default.
 
     retval = rcutils_get_env_var_zero_or_one(
       "RCUTILS_COLORIZED_OUTPUT", "force color",
       "force no color");
-    if (retval < 0) {
+    if (retval == RCUTILS_GET_ENV_ERROR) {
       return RCUTILS_RET_INVALID_ARGUMENT;
-    } else if (retval == 0) {
+    } else if (retval == RCUTILS_GET_ENV_ZERO) {
       g_colorized_output = RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE;
-    } else if (retval == 1) {
+    } else if (retval == RCUTILS_GET_ENV_ONE) {
       g_colorized_output = RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE;
     }
+    // Note that for retval == RCUTILS_GET_ENV_DEFAULT, we do nothing to leave it at default.
 
     // Check for the environment variable for custom output formatting
     const char * output_format;
@@ -892,16 +924,6 @@ void rcutils_logging_console_output_handler(
 
   if (RCUTILS_RET_OK == status) {
     fprintf(g_output_stream, "%s\n", output_array.buffer);
-
-    if (g_output_flush_stream) {
-      int flush_result = fflush(g_output_stream);
-      if (flush_result != 0 && !g_stream_flush_failure_reported) {
-        g_stream_flush_failure_reported = true;
-        fprintf(
-          stderr, "Error: failed to perform fflush on stdout, fflush return code is: %d\n",
-          flush_result);
-      }
-    }
   }
 
   // Only does something in windows
--- a/src/logging.c
+++ b/src/logging.c
@@ -92,14 +92,15 @@ enum rcutils_get_env_retval
   RCUTILS_GET_ENV_ERROR = -1,
   RCUTILS_GET_ENV_ZERO = 0,
   RCUTILS_GET_ENV_ONE = 1,
-  RCUTILS_GET_ENV_DEFAULT = 2,
+  RCUTILS_GET_ENV_EMPTY = 2,
 };
 
-// A utility function to get zero or one from an environment variable.  Returns
-// RCUTILS_GET_ENV_ERROR if we failed to get the environment variable or if it was
-// something we don't understand.  Return RCUTILS_GET_ENV_ZERO if the value in the
-// environment variable is "0", RCUTILS_GET_ENV_ONE if the value in the environment
-// variable is "1", or RCUTILS_GET_ENV_DEFAULT if the environment variables is empty.
+// A utility function to get zero or one from an environment variable.
+// Returns RCUTILS_GET_ENV_ERROR if we failed to get the environment variable
+// or if it was something we don't understand.
+// Return RCUTILS_GET_ENV_ZERO if the value in the environment variable is "0",
+// RCUTILS_GET_ENV_ONE if the value in the environment variable is "1", or
+// RCUTILS_GET_ENV_EMPTY if the environment variables is empty.
 static enum rcutils_get_env_retval rcutils_get_env_var_zero_or_one(
   const char * name, const char * zero_semantic,
   const char * one_semantic)
@@ -117,7 +118,7 @@ static enum rcutils_get_env_retval rcutils_get_env_var_zero_or_one(
   }
 
   if (strcmp(env_var_value, "") == 0) {
-    return RCUTILS_GET_ENV_DEFAULT;
+    return RCUTILS_GET_ENV_EMPTY;
   }
   if (strcmp(env_var_value, "0") == 0) {
     return RCUTILS_GET_ENV_ZERO;
@@ -155,7 +156,7 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
         fprintf(
           stderr,
           "RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED is now ignored.  "
-          "Please set RCUTILS_CONSOLE_USE_STDOUT and RCUTILS_CONSOLE_BUFFERED_STREAM "
+          "Please set RCUTILS_LOGGING_USE_STDOUT and RCUTILS_LOGGING_BUFFERED_STREAM "
           "to control the stream and the buffering of log messages.\n");
       }
     } else {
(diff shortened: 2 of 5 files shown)


dirk-thomas:
> switch it to stdout/buffered by setting `RCUTILS_CONSOLE_LINE_BUFFERED=1`.

With the change in semantic the option name doesn't reflect what it does anymore. I would suggest to change the env var name to something indicating that it changes the default logging stream from `stderr` to `stdout` (or `mixed` if we want to allow the original behavior using a severity specific stream).

> Opinions on how to proceed with fixing tests appreciated.

Options 1 and 2 sound like hacks to me. I think we should a) certainly support option 3. in `launch_testing` and b) use that throughout our tests.

clalancette:
Another option that came to mind is to always use the stdout stream, but have the command-line option toggle between buffered and unbuffered mode.  Bonus points for making it 4 way configurable with stdout/stderr choice and buffered/unbuffered choice, with the default being stdout/unbuffered.

wjwwood:
I agree that the env var name no longer makes sense.

Why not update the tests to match output to stderr (and also set an environment variable to ensure that's the logger's behavior)?

It's possible to route stderr to stdout so they are captured together, but that option isn't exposed at the moment to the `ExecuteProcess` action:

https://github.com/ros2/launch/blob/7b6f4e021a8987baa4e7ccb957c6e1eace1339cb/launch/launch/actions/execute_process.py#L645

It looks like `launch_testing` was never built with support for checking output from stderr...

https://github.com/ros2/launch/blob/f73bc56caa9c7b3c53b264475728392d0e1ce9cf/launch_testing/launch_testing/io_handler.py#L163

@pbaughman can you comment on that?

But it should be possible, as `launch` generates events for each line of stderr as well as stdout, so it can be captured.

pbaughman:
@wjwwood I am happy to comment!

launch_testing actually captures [both stdout and stderr](https://github.com/ros2/launch/blob/f73bc56caa9c7b3c53b264475728392d0e1ce9cf/launch_testing/launch_testing/test_runner.py#L139-L144) in the proc_output handler but currently only has built-in support for [asserting about stdout](https://github.com/ros2/launch/blob/f73bc56caa9c7b3c53b264475728392d0e1ce9cf/launch_testing/launch_testing/asserts/assert_output.py#L81)

It would be a pretty simple improvement to add built-in assertions for stderr too.

jacobperron:
+1 for fixing our tests and launch_testing to work with stderr.

clalancette:
All right, I've done a bunch of work here to fix this up.  Besides this PR, we need (at least):

* https://github.com/ros2/demos/pull/430
* https://github.com/ros2/launch/pull/378

Full Linux CI: * Linux [Build Status](http://ci.ros2.org/job/ci_linux/9299/)

src/logging.c
```diff
@@ -86,6 +88,37 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+  } else if (strcmp(env_var_value, "1") == 0) {
+    return 1;
+  }
+
+  fprintf(
+    stderr,
+    "Warning: unexpected value [%s] specified for %s. "
+    "Valid values are 0 (%s) or 1 (%s).  Default value 0 will be used.\n",
+    env_var_value, name, zero_semantic, one_semantic);
+  return -1;
```
hidmic: @clalancette why not returning `0` then?

src/logging.c
```diff
@@ -869,10 +888,10 @@ void rcutils_logging_console_output_handler(
   SET_STANDARD_COLOR_IN_BUFFER(is_colorized, status, output_array)
 
   if (RCUTILS_RET_OK == status) {
-    fprintf(stream, "%s\n", output_array.buffer);
+    fprintf(g_output_stream, "%s\n", output_array.buffer);
 
-    if (g_force_stdout_line_buffered && stream == stdout) {
-      int flush_result = fflush(stream);
+    if (g_output_flush_stdout && g_output_stream == stdout) {
+      int flush_result = fflush(g_output_stream);
```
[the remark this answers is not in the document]
clalancette: Good call.  I'll also rename the environment variable to "FLUSH_STREAM".

Actually, we could make this slightly more generic by having two environment variables like RCUTILS_CONSOLE_USE_STDERR and RCUTILS_CONSOLE_SYNCHRONOUS, and then setup the stream as appropriate in each of those modes.  What do people think about that?

src/logging.c
```diff
@@ -86,6 +88,37 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
 
+// A utility function to get zero or one from an environment variable.  Returns
+// -2 if we failed to get the environment variable, -1 if the value in the
+// environment variable is something other than "0" or "1", 0 if the environment
+// variable is "0" or "", or "1" if the environment variable is "1".
+static int rcutils_get_env_var_zero_or_one(
+  const char * name, const char * zero_semantic,
+  const char * one_semantic)
+{
+  const char * env_var_value;
```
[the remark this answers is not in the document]
clalancette: Good catch, will do.

src/logging.c
```diff
@@ -86,6 +88,37 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+  const char * ret_str = rcutils_get_env(name, &env_var_value);
+  if (ret_str != NULL) {
```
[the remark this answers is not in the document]
clalancette: Ug, I hate that style, but will do.

src/logging.c
```diff
@@ -86,6 +88,37 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+    fprintf(
+      stderr, "Error getting env. variable "
+      "%s: %s\n", name, ret_str);
+    return -2;
+  }
+
+  if (strcmp(env_var_value, "0") == 0 || strcmp(env_var_value, "") == 0) {
+    return 0;
+  } else if (strcmp(env_var_value, "1") == 0) {
```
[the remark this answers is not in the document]
clalancette: Heh, true, will remove.

src/logging.c
```diff
@@ -86,6 +88,37 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+// variable is "0" or "", or "1" if the environment variable is "1".
+static int rcutils_get_env_var_zero_or_one(
+  const char * name, const char * zero_semantic,
+  const char * one_semantic)
+{
+  const char * env_var_value;
+  const char * ret_str = rcutils_get_env(name, &env_var_value);
+  if (ret_str != NULL) {
+    fprintf(
+      stderr, "Error getting env. variable "
```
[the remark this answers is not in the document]
clalancette: That was a copy-n-paste, but I agree with you.  Will fix.

src/logging.c
clalancette: That's an excellent point, and simplifies the return structure from this function.  Will do that, thanks for the idea.

src/logging.c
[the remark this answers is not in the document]
clalancette: I ended up revamping this with two environment variables: RCUTILS_CONSOLE_USE_STDOUT, which is 0 for `stderr` and 1 for `stdout`, and RCUTILS_CONSOLE_BUFFERED_STREAM, which is 0 for `unbuffered` and 1 for `stdout`. If one or the other isn't specified, then defaults are used.

clalancette:
I've responded to all of the comments, and I've also added a message about backwards compatibility.  I'm going to run a full Linux CI on this again to see where we are at.

Linux CI: [Build Status](http://ci.ros2.org/job/ci_linux/9316/)

src/logging.c
```diff
@@ -99,50 +147,82 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+    enum rcutils_get_env_retval retval = rcutils_get_env_var_zero_or_one(
+      "RCUTILS_CONSOLE_USE_STDOUT", "use stderr",
+      "use stdout");
+    if (retval == RCUTILS_GET_ENV_ERROR) {
+      return RCUTILS_RET_INVALID_ARGUMENT;
+    } else if (retval == RCUTILS_GET_ENV_DEFAULT || retval == RCUTILS_GET_ENV_ZERO) {
+      g_output_stream = stderr;
+    } else if (retval == RCUTILS_GET_ENV_ONE) {
+      g_output_stream = stdout;
+    }
```
hidmic: @clalancette nit: why not a `switch` statement? Otherwise, we should flip those operands :)

src/logging.c
```diff
@@ -803,23 +883,18 @@ void rcutils_logging_console_output_handler(
...
-      stream = stdout;
-      break;
     case RCUTILS_LOG_SEVERITY_WARN:
     case RCUTILS_LOG_SEVERITY_ERROR:
     case RCUTILS_LOG_SEVERITY_FATAL:
-      stream = stderr;
       break;
     default:
       fprintf(stderr, "unknown severity level: %d\n", severity);
-      return;
```
hidmic: @clalancette why removing this?

src/logging.c
```diff
@@ -803,23 +883,18 @@ void rcutils_logging_console_output_handler(
       "call to rcutils_logging_console_output_handler failed.\n");
     return;
   }
-  FILE * stream = NULL;
   switch (severity) {
```
hidmic: @clalancette what's the point of the `switch` statement now? to ensure the given severity level is valid?

src/logging.c
```diff
@@ -99,50 +147,82 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+      "RCUTILS_COLORIZED_OUTPUT", "force color",
+      "force no color");
+    if (retval == RCUTILS_GET_ENV_ERROR) {
+      return RCUTILS_RET_INVALID_ARGUMENT;
+    } else if (retval == RCUTILS_GET_ENV_ZERO) {
+      g_colorized_output = RCUTILS_COLORIZED_OUTPUT_FORCE_DISABLE;
+    } else if (retval == RCUTILS_GET_ENV_ONE) {
+      g_colorized_output = RCUTILS_COLORIZED_OUTPUT_FORCE_ENABLE;
+    }
+    // Note that for retval == RCUTILS_GET_ENV_DEFAULT, we do nothing to leave it at default.
```
hidmic: @clalancette here and above, you may be able to better describe intent using a `switch` statement (i.e. instead of adding comments).

src/logging.c
clalancette: Oh, I think that's a bug.  I'll restore it.

src/logging.c
clalancette: Yeah, good point, I'll switch to a switch statement (see what I did there? :)

src/logging.c
clalancette: Yeah, exactly.  Once I restore the `return` in the default case, it is just error checking.

clalancette:
More review fixes, and one more test fix.  Another Linux CI:

Linux CI: [Build Status](http://ci.ros2.org/job/ci_linux/9332/)

clalancette:
With https://github.com/ros2/ros2/pull/869 ,  this should be much closer to green.  I'm going to kick off a full round of CI on all platforms to see if there is anything else I missed:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9342/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5025/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7648/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9261/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/130/)

clalancette:
All right.  All of the reported failures in the CI job were also on the buildfarm, so I don't think this series of PRs is causing the problem.  This is ready for another round of review, along with the connected PRs.

brawner:
Not sure if you caught this, but there are new MSBuild warnings with this build:

https://ci.ros2.org/job/ci_windows-container/130/warnings43Result/new/package.-1236367313/

clalancette:
> Not sure if you caught this, but there are new MSBuild warnings with this build:

I always miss them.  Thanks for pointing it out, I'll look to see what we do elsewhere.

clalancette:
All right, this is getting to be a bit of an octopus, but I believe I've solved the Windows errors now.  Let's try another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9358/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5041/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7664/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9278/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/146/)

src/logging.c
```diff
@@ -86,6 +87,54 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+  RCUTILS_GET_ENV_ONE = 1,
+  RCUTILS_GET_ENV_DEFAULT = 2,
+};
+
+// A utility function to get zero or one from an environment variable.  Returns
+// RCUTILS_GET_ENV_ERROR if we failed to get the environment variable or if it was
+// something we don't understand.  Return RCUTILS_GET_ENV_ZERO if the value in the
+// environment variable is "0", RCUTILS_GET_ENV_ONE if the value in the environment
+// variable is "1", or RCUTILS_GET_ENV_DEFAULT if the environment variables is empty.
+static enum rcutils_get_env_retval rcutils_get_env_var_zero_or_one(
```
hidmic: @clalancette nit: any reason not to use `Doxygen` syntax?

include/rcutils/strerror.h
```diff
@@ -0,0 +1,33 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcutils/visibility_control.h"
+
+RCUTILS_PUBLIC
+void
+rcutils_safe_strerror(char * buffer, size_t buffer_length);
```
hidmic: @clalancette missing docblock

src/logging.c
```diff
@@ -99,49 +148,96 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+      "RCUTILS_CONSOLE_BUFFERED_STREAM", "not buffered",
+      "buffered");
+    switch (retval) {
+      case RCUTILS_GET_ENV_ERROR:
+        return RCUTILS_RET_INVALID_ARGUMENT;
+      case RCUTILS_GET_ENV_DEFAULT:
+        break;
+      case RCUTILS_GET_ENV_ZERO:
+        if (setvbuf(g_output_stream, NULL, _IONBF, 0) != 0) {
+          char error_string[1024];
```
hidmic: @clalancette meta: nothing against stack allocated char arrays, but I wonder if we should have a statically allocated array that subsequent calls to `rcutils_safe_strerror()` can use, like we do for `rcutils_get_env()`.

src/logging.c
clalancette: > @clalancette nit: any reason not to use `Doxygen` syntax?

Not particularly.  This is an internal function, so I didn't think it was necessary, but I'll claim ignorance on what our policy is for comments on internal functions.

include/rcutils/strerror.h
clalancette: Oops, yeah, good call.  Added in 9200c9fe916bb89d3e2721bd9d340cd83da598f2

src/logging.c
clalancette: I think it is a bad idea for threading reasons.  Keeping it stack allocated like this ensures that this can get called from multiple threads.  Granted, this should only be happening at startup time, but I'd still prefer to keep this thread-safe if at all possible.

Alternatives here are to shrink the size of the stack allocated array, or to make it a heap allocation.  I don't really feel strongly about either one (or leaving it as it is).

clalancette:
CI for new code:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5070/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7685/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9306/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/170/)

dirk-thomas:
Nitpick: each sentence in a comment / docblock should start on a new line.

src/logging.c
```diff
@@ -86,6 +87,54 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+      stderr, "Error getting environment variable "
+      "%s: %s\n", name, ret_str);
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Error getting environment variable %s: %s", name,
+      ret_str);
+    return RCUTILS_GET_ENV_ERROR;
+  }
+
+  if (strcmp(env_var_value, "") == 0) {
+    return RCUTILS_GET_ENV_DEFAULT;
```
dirk-thomas: This function doesn't have any concept of a default value. Therefore I would argue this return code should instead be "empty/not-set". It is up to the caller to determine the semantic of it.

dirk-thomas:
Regarding the env var names: I am not convinced the prefix `RCUTILS_CONSOLE_` is a good choice (I understand it has been used for the previous env vars already). I am missing some kind of hint that these options are affecting `logging`.

src/logging.c
```diff
@@ -99,49 +148,96 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+      return RCUTILS_RET_ERROR;
+    }
+
+    // Set the default output stream for all severities to stderr so that errors
+    // are propagated immediately.  The user can choose to set the output stream
+    // to stdout by setting the RCUTILS_CONSOLE_USE_STDOUT environment
+    // variable to 1.
+    enum rcutils_get_env_retval retval = rcutils_get_env_var_zero_or_one(
+      "RCUTILS_CONSOLE_USE_STDOUT", "use stderr",
+      "use stdout");
```
dirk-thomas: Nitpick: why wrap this line?

src/logging.c
```diff
@@ -99,49 +148,96 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+      case RCUTILS_GET_ENV_ONE:
+        if (setvbuf(g_output_stream, NULL, _IOLBF, 0) != 0) {
+          char error_string[1024];
+          rcutils_safe_strerror(error_string, sizeof(error_string));
+          fprintf(
+            stderr, "Error setting stream buffering mode: %s\n", error_string);
+          RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+            "Error setting stream buffering mode: %s", error_string);
+          return RCUTILS_RET_ERROR;
+        }
```
dirk-thomas: How about de-duplicating these two blocks and choosing the argument to `setvbuf` based on `ZERO` / `ONE`?

include/rcutils/strerror.h
dirk-thomas: How about removing `safe` from the name of the function?

include/rcutils/strerror.h
```diff
@@ -0,0 +1,45 @@
...
+#include "rcutils/visibility_control.h"
+
+/// Retrieve the string corresponding to the last system error.
+/* This function retrieves the value of errno, and calls the system-specific
+ * equivalent of `strerror` on it, storing the output in the provided buffer.
+ * If the error message is longer than the buffer, it will be truncated.
+ * The memory for the c-string buffer that is passed in must be managed by the
+ * caller.
+ *
+ * This function is thread-safe.
```
dirk-thomas: Why is this function thread-safe where the underlying platform specific implementation doesn't guarantee to prevent data races?

src/logging.c
clalancette: That's fine.  I'll change the name of the enum to `RCUTILS_GET_ENV_EMPTY`.

src/logging.c
clalancette: That's what either cpplint or uncrustify wanted.

include/rcutils/strerror.h
clalancette: As far as I know, it does guarantee that.  We are using `strerror_r` on Linux/macOS, which is thread-safe.  On Windows, we are using `strerror_s`, which I believe is an implementation of the C11 version of the same name: https://wiki.sei.cmu.edu/confluence/display/c/CON33-C.+Avoid+race+conditions+when+using+library+functions .  This is also guaranteed to be thread-safe.

include/rcutils/strerror.h
clalancette: I took the name from https://github.com/ros2/rclcpp/blob/2d9c6ea3a77103b91f7984269d163037961a4c04/rclcpp/src/rclcpp/signal_handler.cpp#L164 .  But I don't have a strong opinion, so we can just switch it to `rcutils_strerror` if you prefer.

src/logging.c
clalancette: The issue is that we also have to be careful about "default", since the default is different based on whether we are using `stdout` or `stderr`.  The code would have to go something like:

```
switch (retval) {
    case ERROR:
        return;
    case ONE:
        arg = _IONBF;
    case ZERO:
        arg = _IOLBF;
}

if (retval != DEFAULT) {
    setvbuf(g_output_stream, NULL, arg, 0);
    // error checking
}
```

I don't really think that is better, but if you think we should change to that, I can.

clalancette:
> Regarding the env var names: I am not convinced the prefix `RCUTILS_CONSOLE_` is a good choice (I understand it has been used for the previous env vars already). I am missing some kind of hint that these options are affecting `logging`.

That's true.  I think I'll go with something like `RCUTILS_LOGGING_USE_STDOUT` and `RCUTILS_LOGGING_BUFFERED_STREAM`.

src/logging.c
dirk-thomas: How about without any switch? E.g.:

```
if (ERRO == retval) {
  return;
}
if (ZERO == retval || ONE == retval) {
  auto arg = ONE == retval ? _IONBF : _IOLBF;
  setvbuf(g_output_stream, NULL, arg, 0);

File: src/logging.c (hunk 15 above)
Comment (dirk-thomas) [reply src/logging.c]: I don't see why that would be the case. The line containing all arguments (line 175+176) contains way less than 80 characters.

File: src/logging.c (hunk 13 above)
Comment (hidmic) [reply src/logging.c]: True that `strerror_r` is thread-safe (because `errno` is), unlike the process environment. Disregard.

File: src/logging.c [hunk 18]
```diff
@@ -99,49 +149,97 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
-    ret_str = rcutils_get_env("RCUTILS_COLORIZED_OUTPUT", &colorized_output);
+    // Allow the user to choose how buffering on the stream works by setting
+    // RCUTILS_LOGGING_BUFFERED_STREAM.
+    // With an empty environment variable, use the default of the stream.
+    // With a value of 0, force the stream to be unbuffered.
+    // With a value of 1, force the stream to be line buffered.
+    retval =
+      rcutils_get_env_var_zero_or_one(
+      "RCUTILS_LOGGING_BUFFERED_STREAM", "not buffered",
+      "buffered");
```
wjwwood: nitpick: this is weird indentation I would expect:

```c
retval = rcutils_get_env_var_zero_or_one(
  ...);
```

or

```c
retval =
  rcutils_get_env_var_zero_or_one(...);
```

But not both...

If you must do it like this, I don't know why the linter doesn't make you do:

```c
retval =
  rcutils_get_env_var_zero_or_one(
    ...);
```

include/rcutils/strerror.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcutils/visibility_control.h"
+
+/// Retrieve the string corresponding to the last system error.
+/* This function retrieves the value of errno, and calls the system-specific
```
wjwwood: nitpick: this isn't keeping with the style in `rcutils`, e.g. https://github.com/ros2/rcutils/blob/32486f43f36e9fb09e103c2d10345d876161ea17/include/rcutils/allocator.h#L31 or https://github.com/ros2/rcutils/blob/32486f43f36e9fb09e103c2d10345d876161ea17/include/rcutils/find.h#L27, just to randomly select a few.

src/logging.c
```diff
@@ -86,6 +87,55 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+// RCUTILS_GET_ENV_ONE if the value in the environment variable is "1", or
+// RCUTILS_GET_ENV_EMPTY if the environment variables is empty.
+static enum rcutils_get_env_retval rcutils_get_env_var_zero_or_one(
+  const char * name, const char * zero_semantic,
+  const char * one_semantic)
+{
+  const char * env_var_value = NULL;
+  const char * ret_str = rcutils_get_env(name, &env_var_value);
+  if (NULL != ret_str) {
+    fprintf(
```
wjwwood: Please use our fwrite macro instead, as some users will want to scrub use of `fprintf`:

https://github.com/ros2/rcutils/blob/32486f43f36e9fb09e103c2d10345d876161ea17/include/rcutils/error_handling.h#L45

I realize other parts of this code use `fprintf`, but we're trying to curb that in new code. The use case is to be able to avoid using the stdout or stderr file handles by changing as few lines of code as possible. This is needed if, for safety reasons, you cannot use the system file handles, or on platforms that do not provide a console.

If you still want to use string formatting, you can combine fwrite with `rcutils_snprintf`, like here:

https://github.com/ros2/rcutils/blob/32486f43f36e9fb09e103c2d10345d876161ea17/include/rcutils/error_handling.h#L215-L222

src/logging.c
```diff
@@ -86,6 +87,55 @@ rcutils_ret_t rcutils_logging_initialize(void)
...
+    return RCUTILS_GET_ENV_EMPTY;
+  }
+  if (strcmp(env_var_value, "0") == 0) {
+    return RCUTILS_GET_ENV_ZERO;
+  }
+  if (strcmp(env_var_value, "1") == 0) {
+    return RCUTILS_GET_ENV_ONE;
+  }
+
+  fprintf(
```
wjwwood: And here.

src/logging.c
```diff
@@ -99,49 +149,97 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+          char error_string[1024];
+          rcutils_safe_strerror(error_string, sizeof(error_string));
+          fprintf(
+            stderr, "Error setting stream buffering mode: %s\n", error_string);
+          RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+            "Error setting stream buffering mode: %s", error_string);
+          return RCUTILS_RET_ERROR;
+        }
+        break;
+    }
```
wjwwood: Why no default case? I'm not sure if the compiler will warn you (certainly will not error I think) if you added a field to the enum later and didn't update the switch statements?

src/logging.c
clalancette: Again, I have no idea why cpplint and uncrustify ask for the things they do.  I write code, and then I fight them until they pass.  I can fight them some more to make it seem more like real code.

include/rcutils/strerror.h
clalancette: Arg.  The style is inconsistent, and I always happen to pick the examples that are "different".  I can change it to the `**`.

src/logging.c
clalancette: So, there's a lot of uses of `fprintf` in this file.  I guess my question is whether we should switch them all?  If that is the case, then I'd suggest a follow-up PR for all of them.

src/logging.c
clalancette: Good call.  I've added a default case to all switches.

src/logging.c
dirk-thomas: There is nothing weird about the above. An indentation happens for every opened parenthesis. The fact that `retval =` and `rcutils_get_env_var_zero_or_one(` are on two separate lines doesn't change the fact that the following arguments are within parenthesis and therefore one level indented. What William was pointing out was that having `rcutils_get_env_var_zero_or_one(` on the same level as the arguments isn't desired since it makes the code hard to visually parse. The first snipped William posted will pass the linter just fine and is what we use throughout the code.

wjwwood: Honestly I don't care that much, I just thought I'd mention it in case it was unintentional.

I don't think what's there now should pass the linter, as for me things inside parenthesis should be indented once _from the line_ that the parenthesis open on, not the line the statement starts on. But again, I'm not going to waste time on it.

src/logging.c
wjwwood: I think it would be nice to avoid extending the problem, but you're right that the other calls should be changed at some point and that work is out of scope for this pr. I'd still vote for not adding fprintf's, but again, I'm not going to block this pr on that.

include/rcutils/strerror.h
```diff
@@ -23,7 +23,7 @@ extern "C"
 #include "rcutils/visibility_control.h"
 
 /// Retrieve the string corresponding to the last system error.
-/* This function retrieves the value of errno, and calls the system-specific
+/** This function retrieves the value of errno, and calls the system-specific
```
wjwwood: Actually this still doesn't follow the style... sorry.

It should be:

```
/// Brief ... description.
/**
 * Start detailed description here.
 */
...
```

src/logging.c
```diff
@@ -185,44 +184,41 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
       case RCUTILS_GET_ENV_ONE:
         g_output_stream = stdout;
         break;
+      default:
+        fprintf(
+          stderr, "Invalid return from environment fetch\n");
+        RCUTILS_SET_ERROR_MSG(
+          "Invalid return from environment fetch");
+        return RCUTILS_RET_ERROR;
```
wjwwood: I'm not a huge fan of setting the error state and printing to stderr. In my opinion, libraries shouldn't print to stderr unless there's no other way to raise an issue. This function can raise an issue with the return code and error state.

Again, I know this is the case already in this file, but I felt I should bring it up as long as we're looking at this file.

include/rcutils/strerror.h
clalancette: This should be fixed now.

src/logging.c
clalancette: This is done now.

src/logging.c
clalancette: I've done...something else here that seems to make uncrustify happy.

clalancette:
I've now responded to most of the open comments.  The ones remaining are as follows:

1.  Change `rcutils_safe_strerror` -> `rcutils_strerror`.  I think having the safe in here is a good additional hint that this is a thread-safe version of strerror, so I'm going to leave it.
1.  Don't use `fprintf` directly, and instead use `RCUTILS_SAFE_FWRITE_TO_STDERR`.  As I see it, this whole file violates that, so I'd rather not make this (already getting large) PR any larger, and file an issue for a follow-up to fix it.
1.  Don't both print to stderr and return an error, and instead just set and return an error.  Again, this whole file violates that, so I'd rather keep it internally consistent and then file an issue to follow-up to fix it.

So with that, I think this is ready for another round of review.

clalancette:
Another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9403/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5080/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7695/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9320/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/180/)

clalancette:
All right, all PRs approved.  One more CI to make sure everything is happy, then I'll merge:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9478/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5206/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7751/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9402/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/243/)

PR #197, opened 2020-01-08
calculate file and directory size

Karsten1987:
connects to https://github.com/ros2/rosbag2/issues/248
Signed-off-by: Karsten Knese <karsten@openrobotics.org>


src/filesystem.c
```diff
@@ -210,6 +212,53 @@ rcutils_mkdir(const char * abs_path)
   return success;
 }
 
+size_t
+rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_t allocator)
+{
+  size_t dir_size = 0;
+#ifdef _WIN32
+  char * path = rcutils_joint_path(directory_path, "*", allocator);
```
clalancette: This should be `rcutils_join_path`.

Karsten1987:
This code is currently already being used within rosbag2. As per top-level issue description, there are currently three places within ROS2 (`rosbag2`, `rcpputils`, rcutils`) where filesystem helper functions are implemented. 
rosbag2 can solely rely on the other two packages except that the two calculate functions are currently unavailable in `rcutils`.

> What about the two different kinds of symlinks on Windows? What happens if one file in the directory is very large, but has its permissions set so the current user can't read it? Also, it seems pretty easy to Denial Of Service the application by creating a directory with many little files. Maybe we can figure out a different way to do this.

Is there a safe & cross-platform way of getting the file size as well as directory size? I am currently not aware of any.

src/filesystem.c
Karsten1987: nice catch. Thanks

src/filesystem.c
```diff
@@ -210,6 +212,53 @@ rcutils_mkdir(const char * abs_path)
...
+      allocator.deallocate(file_path, allocator.state);
+    }
+  }
+  closedir(dir);
+  return dir_size;
+#endif
+}
+
+size_t
+rcutils_get_file_size(const char * file_path)
```
piraka9011: Does this need to check if the path exists?

src/filesystem.c
```diff
@@ -210,6 +212,53 @@ rcutils_mkdir(const char * abs_path)
...
+{
+  size_t dir_size = 0;
+#ifdef _WIN32
+  char * path = rcutils_join_path(directory_path, "*", allocator);
+  WIN32_FIND_DATA data;
+  HANDLE handle = FindFirstFile(path, &data);
+  allocator.deallocate(path, allocator.state);
+  if (handle != INVALID_HANDLE_VALUE) {
+    do {
+      if (strcmp(data.cFileName, ".") != 0 && strcmp(data.cFileName, "..") != 0) {
```
piraka9011: Maybe make this a separate function (`directory_entry_is_clean`) or better docs since it's not directly clear what this is doing IMO.

piraka9011:
Some of this was proposed in rcpputils as well [here](https://github.com/ros2/rcpputils/pull/28).

src/filesystem.c
Karsten1987: I am not convinced that a separate function for this makes sense. I could see a function to check whether a directory is empty, but that wouldn't really apply here as this check really only skips the two file handles (`.` = local folder, `..` = parent folder).

I can update the function with some inline comments if that helps.

src/filesystem.c
```diff
@@ -210,6 +212,69 @@ rcutils_mkdir(const char * abs_path)
   return success;
 }
 
+size_t
+rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_t allocator)
+{
+  size_t dir_size = 0;
+
+  if (!rcutils_is_directory(directory_path)) {
+    fprintf(stderr, "Path is no directory: %s\n", directory_path);
```
piraka9011: Typo

src/filesystem.c
```diff
@@ -210,6 +212,69 @@ rcutils_mkdir(const char * abs_path)
...
+  closedir(dir);
+  return dir_size;
+#endif
+}
+
+size_t
+rcutils_get_file_size(const char * file_path)
+{
+  if (!rcutils_is_file(file_path)) {
+    fprintf(stderr, "Path is no file: %s\n", file_path);
```
piraka9011: Typo: `not a`

Karsten1987:
@clalancette please let me know if you have any further comments on this. CI turns out to be green and both PRs are approved. I just wanted to make sure we can easily address your concerns here in this PR. If it turns out to be a more complex change, I'd propose we open a follow up ticket/PR to this.

Karsten1987:
@clalancette I am going ahead and merge. Please feel free to come up with a follow up issue for addressing your concerns.

PR #202, opened 2020-01-30
add quality declaration

wjwwood:
This is a quality declaration for `rcutils` as described in the proposed REP-2004 (see https://github.com/ros-infrastructure/rep/pull/218).

This declaration is aspirational, in that it states things which aren't currently true, but we plan to do soon so we can get it up to the highest level. I'll use it as a kind of TODO list for fixing up `rcutils`, using line comments to keep track of progress of certain tasks.

This will most likely continue to evolve with the REP as well.

### DIFF
--- a/QUALITY_DECLARATION.md
+++ b/QUALITY_DECLARATION.md
@@ -1,0 +1,132 @@
+This document is a declaration of software quality for the `rcutils` package, based on the guidelines in [REP-2004](https://www.ros.org/reps/rep-2004.html).
+
+# `rcutils` Quality Declaration
+
+The package `rcutils` claims to be in the **Quality Level 1** category.
+
+Below are the rationales, notes, and caveats for this claim, organized by each requirement listed in the [Package Requirements for Quality Level 1 in REP-2004](https://www.ros.org/reps/rep-2004.html).
+
+## Version Policy
+
+### Version Scheme
+
+`rcutils` uses `semver` according to the recommendation for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#versioning), and is at or above a stable version, i.e. `>= 1.0.0`.
+
+### API Stability Within a Released ROS Distribution
+
+`rcutils` will not break public API within a released ROS distribution, i.e. no major releases once the ROS distribution is released.
+
+### ABI Stability Within a Released ROS Distribution
+
+`rcutils` contains C code and therefore must be concerned with ABI stability, and will maintain ABI stability within a ROS distribution.
+
+### Public API Declaration
+
+All symbols in the installed headers are considered part of the public API.
+
+All installed headers are in the `include` directory of the package, headers in any other folders are not installed and considered private.
+
+Additionally, there are generated header files which are installed and therefore part of the public API.
+The source templates for these generated headers are in the `resource` folder.
+
+## Change Control Process
+
+TODO fix link
+
+`rcutils` follows the recommended guidelines for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#change_control_process).
+
+This includes:
+
+- all changes occur through a pull request
+- all pull request have two peer reviews
+- all pull request must pass CI on all [tier 1 platforms](https://www.ros.org/reps/rep-2000.html#support-tiers)
+- all pull request must resolve related documentation changes before merging
+
+## Documentation
+
+### Feature Documentation
+
+TODO fix link
+
+`rcutils` has a [feature list](TODO) and each item in the list links to the corresponding feature documentation.
+There is documentation for all of the features, and new features require documentation before being added.
+
+### Public API Documentation
+
+TODO fix link
+
+`rcutils` has embedded API documentation and it is generated using doxygen and is [hosted](TODO) along side the feature documentation.
+There is documentation for all of the public API, and new additions to the public API require documentation before being added.
+
+### License
+
+The license for `rcutils` is Apache 2.0, and a summary is in each source file, the type is declared in the `package.xml` manifest file, and a full copy of the license is in the `LICENSE` file.
+
+There is an automated test which runs a linter that ensures each file has a license statement.
+
+### Copyright Statements
+
+The copyright holders each provide a statement of copyright in each source code file in `rcutils`.
+
+There is an automated test which runs a linter that ensures each file has at least one copyright statement.
+
+## Testing
+
+### Feature Testing
+
+Each feature in `rcutils` has corresponding tests which simulate typical usage, and they are located in the `test` directory.
+New features are required to have tests before being added.
+
+### Public API Testing
+
+Each part of the public API have tests, and new additions or changes to the public API require tests before being added.
+The tests aim to cover both typical usage and corner cases, but are quantified by contributing to code coverage.
+
+### Coverage
+
+TODO fix link
+
+`rcutils` follows the recommendations for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#coverage), and opts to use branch coverage instead of line coverage.
+
+This includes:
+
+- tracking and reporting branch coverage statistics
+- achieving and maintaining branch coverage at or above 95%
+- no lines are manually skipped in coverage calculations
+
+Changes are required to make a best effort to keep or increase coverage before being accepted, but decreases are allowed if properly justified and accepted by reviewers.
+
+Current coverage statistics can be viewed here:
+
+TODO FIXME
+
+### Performance
+
+TODO fix link
+
+`rcutils` follows the recommendations for performance testing of C code in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#performance-c), and opts to do performance analysis on each release rather than each change.
+
+TODO how to run perf tests, where do they live, etc.
+
+TODO exclusions of parts of the code from perf testing listed here? or linked to?
+
+### Linters and Static Analysis
+
+TODO fix link
+
+`rcutils` uses and passes all the standard linters and static analysis tools for a C package as described in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#linters-and-static-analysis).
+
+TODO any qualifications on what "passing" means for certain linters
+
+## Dependencies
+
+`rcutils` has no run-time or build-time dependencies that need to be considered for this declaration.
+
+It has several "buildtool" dependencies, which do not affect the resulting quality of the package, because they do not contribute to the public library API.
+It also has several test dependencies, which do not affect the resulting quality of the package, because they are only used to build and run the test code.
+
+## Platform Support
+
+`rcutils` supports all of the tier 1 platforms as described in [REP-2000](https://www.ros.org/reps/rep-2000.html#support-tiers), and tests each change against all of them.
+
+TODO make additional statements about non-tier 1 platforms?
--- a/README.md
+++ b/README.md
@@ -2,6 +2,12 @@
 
 `rcutils` is a C API consisting of macros, functions, and data structures used through out the ROS 2 code base.
 
+## Quality Declaration
+
+This package claims to be in the **Quality Level 1** category, see the [Quality Declaration](./QUALITY_DECLARATION.md) for more details.
+
+## API
+
 The API is a combination of parts:
 
 - Allocator concept, used to inject the allocating and deallocating methods into a function or type.


QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+The package `rcutils` claims to be in the **Quality Level 1** category.
+
+Below are the rationales, notes, and caveats for this claim, organized by each requirement listed in the [Package Requirements for Quality Level 1 in REP-2004](https://www.ros.org/reps/rep-2004.html).
+
+## Version Policy
+
+### Version Scheme
+
+`rcutils` uses `semver` according to the recommendation for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#versioning), and is at or above a stable version, i.e. `>= 1.0.0`.
```
wjwwood: The link is valid, but the content at the link needs to be reviewed and updated most likely.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+All installed headers are in the `include` directory of the package, headers in any other folders are not installed and considered private.
+
+Additionally, there are generated header files which are installed and therefore part of the public API.
+The source templates for these generated headers are in the `resource` folder.
+
+## Change Control Process
+
+TODO fix link
+
+`rcutils` follows the recommended guidelines for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#change_control_process).
```
wjwwood: The ROS 2 developer guide needs a common change control process section for us to link to and not repeat ourselves too much.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+- all pull request must pass CI on all [tier 1 platforms](https://www.ros.org/reps/rep-2000.html#support-tiers)
+- all pull request must resolve related documentation changes before merging
+
+## Documentation
+
+### Feature Documentation
+
+TODO fix link
+
+`rcutils` has a [feature list](TODO) and each item in the list links to the corresponding feature documentation.
```
wjwwood: Currently the feature list is in the README, which is then used by doxygen, but honestly it should probably be moved into a separate file (set of files) in the `doc` directory:

https://github.com/ros2/rcutils/blob/8f2d2b9781fe254810deb2188bfa0dad8900bb4e/Doxyfile#L11

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+There is documentation for all of the features, and new features require documentation before being added.
+
+### Public API Documentation
+
+TODO fix link
+
+`rcutils` has embedded API documentation and it is generated using doxygen and is [hosted](TODO) along side the feature documentation.
```
wjwwood: Not 100% sure where we should link. docs.ros2.org for now I guess? For example:

https://docs.ros2.org/eloquent/api/rcutils/index.html

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+### Public API Testing
+
+Each part of the public API have tests, and new additions or changes to the public API require tests before being added.
+The tests aim to cover both typical usage and corner cases, but are quantified by contributing to code coverage.
+
+### Coverage
+
+TODO fix link
+
+`rcutils` follows the recommendations for ROS Core packages in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#coverage), and opts to use branch coverage instead of line coverage.
```
wjwwood: The developer guide needs a section on coverage too.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+- tracking and reporting branch coverage statistics
+- achieving and maintaining branch coverage at or above 95%
+- no lines are manually skipped in coverage calculations
+
+Changes are required to make a best effort to keep or increase coverage before being accepted, but decreases are allowed if properly justified and accepted by reviewers.
+
+Current coverage statistics can be viewed here:
+
+TODO FIXME
```
wjwwood: Once we have some hosted coverage history... link to it.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Performance
+
+TODO fix link
+
+`rcutils` follows the recommendations for performance testing of C code in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#performance-c), and opts to do performance analysis on each release rather than each change.
```
wjwwood: Developer guide needs a section on performance tests (like the coverage one).

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+TODO how to run perf tests, where do they live, etc.
+
+TODO exclusions of parts of the code from perf testing listed here? or linked to?
+
+### Linters and Static Analysis
+
+TODO fix link
+
+`rcutils` uses and passes all the standard linters and static analysis tools for a C package as described in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#linters-and-static-analysis).
```
wjwwood: Developer guide needs a section about standard linters and static analyzers.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+TODO fix link
+
+`rcutils` has embedded API documentation and it is generated using doxygen and is [hosted](TODO) along side the feature documentation.
+There is documentation for all of the public API, and new additions to the public API require documentation before being added.
+
+### License
+
+The license for `rcutils` is Apache 2.0, and a summary is in each source file, the type is declared in the `package.xml` manifest file, and a full copy of the license is in the `LICENSE` file.
+
+There is an automated test which runs a linter that ensures each file has a license statement.
```
chapulina: Consider adding links to the files mentioned (`package.xml` and `LICENSE`), as well as a link to the results of the automated test.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+The copyright holders each provide a statement of copyright in each source code file in `rcutils`.
+
+There is an automated test which runs a linter that ensures each file has at least one copyright statement.
+
+## Testing
+
+### Feature Testing
+
+Each feature in `rcutils` has corresponding tests which simulate typical usage, and they are located in the `test` directory.
```
chapulina: Consider adding a full link to the `test` directory, for the case when this declaration is rendered outside the repo.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+New features are required to have tests before being added.
+
+### Public API Testing
+
+Each part of the public API have tests, and new additions or changes to the public API require tests before being added.
```
chapulina:
```suggestion
Each part of the public API has tests, and new additions or changes to the public API require tests before being added.
```

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+TODO exclusions of parts of the code from perf testing listed here? or linked to?
+
+### Linters and Static Analysis
+
+TODO fix link
+
+`rcutils` uses and passes all the standard linters and static analysis tools for a C package as described in the [ROS 2 Developer Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#linters-and-static-analysis).
+
+TODO any qualifications on what "passing" means for certain linters
```
chapulina: Also add a link to the linter results.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+`rcutils` has no run-time or build-time dependencies that need to be considered for this declaration.
+
+It has several "buildtool" dependencies, which do not affect the resulting quality of the package, because they do not contribute to the public library API.
+It also has several test dependencies, which do not affect the resulting quality of the package, because they are only used to build and run the test code.
+
+## Platform Support
+
+`rcutils` supports all of the tier 1 platforms as described in [REP-2000](https://www.ros.org/reps/rep-2000.html#support-tiers), and tests each change against all of them.
+
+TODO make additional statements about non-tier 1 platforms?
```
chapulina: It would also be nice to add a link to supporting evidence here. CI results? Binary packages?

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+### License
+
+The license for `rcutils` is Apache 2.0, and a summary is in each source file, the type is declared in the `package.xml` manifest file, and a full copy of the license is in the `LICENSE` file.
+
+There is an automated test which runs a linter that ensures each file has a license statement.
+
+### Copyright Statements
+
+The copyright holders each provide a statement of copyright in each source code file in `rcutils`.
```
mjcarroll: Do we need to mention DCO here?

QUALITY_DECLARATION.md
Blast545: Although it's a good idea to have the common rules for versioning/documentation in the developer guide, I think that the quality declaration document should be as self contained as possible and at least contain a brief summary of those guidelines. That would avoid users jumping back and forth to the developer guide when reading the packages documentation.

QUALITY_DECLARATION.md
Blast545: [Developer guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#test-coverage) says that the metric to be used when measuring coverage will be branch coverage. However, [REP2004](https://github.com/ros-infrastructure/rep/blob/c2ba03bac23b8f313e47820833fad000dbbe058b/rep-2004.rst) says that metric used in ROS Core packages will be line coverage, aiming for percentage above 95% (in fact talks about branch coverage as something optional).

This comes with two issues:
1. There's an inconsistency to be solved, maybe discussed, for the ROS2 core packages. Will we use branch or line coverage? 
2. There's no percentage coverage required explicitly for packages outside the ROS2 organization, and maybe this has to be changed in the REP2004 to set a desired standard.

Please check this @chapulina

QUALITY_DECLARATION.md
Blast545: I think for this would be a good idea to have a link pointing to the latest success jobs in packaging for each of the tier1 platforms. Like the panel shown on the main page of [CI](https://ci.ros2.org/) but only showing some metrics, maybe including test metrics here too. Not sure if this can be generated easily.

QUALITY_DECLARATION.md
```diff
@@ -0,0 +1,132 @@
...
+
+There is an automated test which runs a linter that ensures each file has at least one copyright statement.
```
Blast545: The automated test checks only for the license statement in each source file, the copyright statement is checked with DCO through github, so maybe this line should be removed and DCO would be mentioned in the line above where @mjcarroll commented.

QUALITY_DECLARATION.md
Blast545: The linters have some documentation in their repositories [Like this one](https://github.com/ament/ament_lint/blob/master/ament_cmake_lint_cmake/doc/index.rst), I think it should be made some effort to include links to all those directly in this section of the [tutorials](https://index.ros.org/doc/ros2/Tutorials/Ament-CMake-Documentation/#testing-and-linting), in a section of the developer guide or maybe in a different index.ros2 article.

QUALITY_DECLARATION.md
wjwwood: Hmm, I don't completely agree, but I think we're mostly on the same page. I think the document should make sense on its own, but details definitely should be collected in one place if possible. What I don't want to have to do is that if we change our recommendation from 1-review into 2-reviews for each pull request that I have to touch every repository manually.

In the cases where the linked to guidelines leave some flexibility, then I think it's important for the declaration that is referencing it to clarify what it is choosing to do, e.g. if it chooses line versus branch coverage.

QUALITY_DECLARATION.md
wjwwood: The developer guide is out of date still, see the original proposed pr (https://github.com/ros2/ros2_documentation/pull/460) that led to REP-2004's proposal:

> The end goal would be to move most of this into a REP, and then replace this section with the prescriptions on how to achieve the categories for ROS core packages.

Basically, consider the developer guidelines out-of-date until we separate the requirements into the REP and the prescriptions into a (as yet non-existent) pull request against the developer guide lines.

----

That being said, I don't feel strongly about the line versus branch coverage. My thinking is that in some cases it might not be technically feasible to do branch coverage, so I left it as "at least line, but may use branch".

----

With respect to the percentage, that's still being discussed but we will have some recommendation for ROS core packages via the developer guide. What's already there is out-of-date and should be ignored for now.

QUALITY_DECLARATION.md
wjwwood: Each job has a badge already, we can use those from the packaging jobs, like these:

- https://ci.ros2.org/view/packaging/job/packaging_linux/ [Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux/)
- https://ci.ros2.org/view/packaging/job/packaging_windows/ [Build Status](https://ci.ros2.org/view/packaging/job/packaging_windows/)

But those are for all of the things in the `ros2.repos` file, not just this package...

We also have the `.deb` packaging jobs, which are per package, but they are only for Linux and aren't built nightly (right now), e.g.:

- http://build.ros2.org/view/Ebin_uB64/job/Ebin_uB64__rcutils__ubuntu_bionic_amd64__binary/ [Build Status](http://build.ros2.org/view/Ebin_uB64/job/Ebin_uB64__rcutils__ubuntu_bionic_amd64__binary/)

There are dev jobs, but those are per repository, not per package, e.g.:

- http://build.ros2.org/job/Edev__rcutils__ubuntu_bionic_amd64/ [Build Status](http://build.ros2.org/job/Edev__rcutils__ubuntu_bionic_amd64/)

QUALITY_DECLARATION.md
wjwwood: > The automated test checks only for the license statement in each source file, the copyright statement is checked with DCO through github

That's not correct as far as I know... This linter checks for copyright claims in each file:

https://github.com/ament/ament_lint/blob/master/ament_copyright/doc/index.rst

The DCO check makes sure that each commit has been signed-off by the author, which is related, but different.

QUALITY_DECLARATION.md
wjwwood: What do you think we should mention here? Wouldn't that belong under the "change process" part? The DCO doesn't have anything to do with copyright as far as I know (though it is related), instead thee DCO checks that each commit is signed-off by the person who authored it.

wjwwood: Also, it would be good to comment about this on the REP-2004 pr instead: https://github.com/ros-infrastructure/rep/pull/218

QUALITY_DECLARATION.md
wjwwood: Yeah, links to how to use the linters, as well as which ones are on by default, and links to each of those own documentation, should all be part of what's in the developer guide, to which this will link.

QUALITY_DECLARATION.md
Blast545: Yes, didn't thought about the cases where the recommended guidelines are changed, seems good then to have a 'Recommended guidelines' document to reference

QUALITY_DECLARATION.md
Blast545: Yeah, I thought about those, in the sense that these links can be referenced and checked easily, but done for each package. 

Maybe we could point to the latest successful build tests of each package, per platform, something like what's found here:
[Test results ament_index_python nightly_osx_debug](https://ci.ros2.org/job/nightly_osx_debug/lastSuccessfulBuild/testReport/ament_index_python.test/).

QUALITY_DECLARATION.md
Blast545: You are right, I thought that one only checked for the license header. Just tested it and it also looks for the copyright notice using regex with a year

ros-discourse:

https://discourse.ros.org/t/rfc-rep-2004-package-quality-categories/13150/1

chapulina:
This has been superseded by #240 . Please reopen if that's not the case.

PR #203, opened 2020-01-31
code style only: wrap after open parenthesis if not in one line

dirk-thomas:
Style update to match the ROS 2 development guide and pass with the updated linter configuration from ament/ament_lint#210.

### DIFF
--- a/test/test_time.cpp
+++ b/test/test_time.cpp
@@ -59,7 +59,8 @@ TEST_F(TestTimeFixture, test_rcutils_system_time_now) {
   rcutils_reset_error();
   // Check for normal operation (not allowed to alloc).
   rcutils_time_point_value_t now = 0;
-  EXPECT_NO_MEMORY_OPERATIONS({
+  EXPECT_NO_MEMORY_OPERATIONS(
+  {
     ret = rcutils_system_time_now(&now);
   });
   EXPECT_EQ(ret, RCUTILS_RET_OK) << rcutils_get_error_string().str;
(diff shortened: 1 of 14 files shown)


test/test_time.cpp
```diff
@@ -102,7 +105,8 @@ TEST_F(TestTimeFixture, test_rcutils_steady_time_now) {
   std::this_thread::sleep_for(std::chrono::milliseconds(100));
   // Then take a new timestamp with each and compare.
   rcutils_time_point_value_t later;
-  EXPECT_NO_MEMORY_OPERATIONS({
+  EXPECT_NO_MEMORY_OPERATIONS(
+  {
```
wjwwood: Btw, these are actually wrong. This should become:

```c++
EXPECT_NO_MEMORY_OPERATIONS(
  {
    ret = rcutils_steady_time_now(&later);
  });
```

Because if you replaced the first argument with a variable name it would be:

```c++
EXPECT_NO_MEMORY_OPERATIONS(
  argument);
```

And not this:

```c++
EXPECT_NO_MEMORY_OPERATIONS(
argument);
```

dirk-thomas: I don't disagree. This is just what `uncrustify` can do atm. The options are simple:
* report the problem upstream (or even better contribute a PR to support this in the way we want it)
* change the configuration options we use which seems to sacrifice other parts of the style
* not use `uncrustify` at all if we are unhappy with what it does and just not enforce that level of style

wjwwood: I know, but I was just pointing it out in case we revisit this in the future. I don't care for the new state, but have time to argue to change it back or fix uncrustify.

PR #206, opened 2020-02-11
Split bug fix

Blast545:
This is a bug I encountered adding tests for the split.c module. When calling the split function with the following str `"my//hello//world/////"` and this delimiter `'/'`, the module crashes trying to free a memory not allocated. 

In [Line 81](https://github.com/ros2/rcutils/blob/44d00371833253238925cf29f77a2d640dc387e4/src/split.c#L81) it can be seen the behavior expected for consecutive delimiters, no deallocation is required there.

### DIFF
--- a/test/test_split.cpp
+++ b/test/test_split.cpp
@@ -67,41 +67,48 @@ TEST(test_split, split) {
   ASSERT_EQ(RCUTILS_RET_OK, ret);
 
   rcutils_string_array_t tokens1 = test_split("hello_world", '/', 1);
+  ASSERT_EQ(tokens1.size, 1u);
   EXPECT_STREQ("hello_world", tokens1.data[0]);
   ret = rcutils_string_array_fini(&tokens1);
   ASSERT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
 
   rcutils_string_array_t tokens2 = test_split("hello/world", '/', 2);
+  ASSERT_EQ(tokens2.size, 2u);
   EXPECT_STREQ("hello", tokens2.data[0]);
   EXPECT_STREQ("world", tokens2.data[1]);
   ret = rcutils_string_array_fini(&tokens2);
   ASSERT_EQ(RCUTILS_RET_OK, ret);
 
   rcutils_string_array_t tokens3 = test_split("/hello/world", '/', 2);
+  ASSERT_EQ(tokens3.size, 2u);
   EXPECT_STREQ("hello", tokens3.data[0]);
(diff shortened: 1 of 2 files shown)


Blast545:
I added the case for the failing condition, I'm posting CI results as well:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9304/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4987/)
* macOS (CI is not working on macOS at the moment)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9222/)

test/test_split.cpp
```diff
@@ -128,6 +128,13 @@ TEST(test_split, split) {
   EXPECT_STREQ("world", tokens10.data[2]);
   ret = rcutils_string_array_fini(&tokens10);
   ASSERT_EQ(RCUTILS_RET_OK, ret);
+
+  rcutils_string_array_t tokens11 = test_split("///my//hello//world/////", '/', 3);
+  EXPECT_STREQ("my", tokens11.data[0]);
```
hidmic: @Blast545 consider to `ASSERT_EQ(tokens11.size, 3);` before indexing.

Blast545: @hidmic I followed the standard of the previous tests, should I add that to all?

hidmic: That'd indeed be better.

test/test_split.cpp
```diff
@@ -67,67 +67,85 @@ TEST(test_split, split) {
   ASSERT_EQ(RCUTILS_RET_OK, ret);
 
   rcutils_string_array_t tokens1 = test_split("hello_world", '/', 1);
+  ASSERT_EQ(tokens1.size, 1u);
```
clalancette: Are these necessary?  As far as I can tell, the `test_split()` function already does this check based on the final argument.  So I'd suggest we just remove these (unless I'm missing something).

hidmic: Oh, my bad. I missed that when I suggested adding assertions. My apologies @Blast545, but @clalancette is right.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9321/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5011/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7631/) (MacOS jobs are not working at the moment)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9244/)

PR #207, opened 2020-02-12
Fixes in comments.

clalancette:

I just noticed these misspellings while reviewing other PRs.


clalancette:
Thanks for the review.  I've verified that this builds and passes tests locally.  Since it is only a change in comments, I'm going to merge without CI.

PR #214, opened 2020-03-06
Remove test for large allocation failure.

clalancette:
This has potential performance implications, and also causes
build warnings on 32-bit platforms since the size is too big
to fit into a size_t.  It's also a pretty artificial test,
so just remove it.

This should fix one of the build warnings in https://ci.ros2.org/view/nightly/job/nightly_linux-armhf_debug/409/warnings23Result/new/ .  @Blast545 FYI.

### DIFF
--- a/test/test_array_list.cpp
+++ b/test/test_array_list.cpp
@@ -368,12 +368,6 @@ TEST_F(ArrayListTest, init_list_bad_allocator_fail) {
   EXPECT_EQ(RCUTILS_RET_BAD_ALLOC, ret) << rcutils_get_error_string().str;
 }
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI just on armhf to prove that the warning is gone:

* Linux armhf [Build Status](http://ci.ros2.org/job/ci_linux-armhf/78/)

Blast545:
@clalancette Thanks for tagging me, I'll use a different approach to test the case covered by that portion of code.

clalancette:
> Thanks for tagging me, I'll use a different approach to test the case covered by that portion of code.

Sounds good, thanks for understanding.

PR #215, opened 2020-03-10
Included utils to load, unload and get symbols from shared libraries

ahcorde:
As discussed in this PR https://github.com/ros/class_loader/pull/139. It makes sense to include the logic of loading/unloading shared libraries and symbols in `rcutils` . These functions will be used in `class_loader`, `rmw_implementation`, `rosidl_typesupport` and `rosbag2`.

This other PR makes use of this new capability https://github.com/ros/class_loader/pull/139

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -368,7 +368,7 @@ endif()
 
 ament_export_dependencies(ament_cmake)
 ament_export_include_directories(include)
-ament_export_libraries(${PROJECT_NAME})
+ament_export_libraries(${PROJECT_NAME} ${CMAKE_DL_LIBS})
 ament_package()
 
 install(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,7 +326,7 @@ if(BUILD_TESTING)
   )
 
   if(TARGET test_shared_library)
-    add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.cpp)
+    add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
     target_link_libraries(test_shared_library ${PROJECT_NAME})
   endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,11 +315,7 @@ if(BUILD_TESTING)
     target_link_libraries(test_repl_str ${PROJECT_NAME})
   endif()
 
-  if(WIN32)
-    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
-  else()
-    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
-  endif()
+  set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
 
   ament_add_gtest(test_shared_library test/test_shared_library.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ set(rcutils_sources
   src/logging.c
   src/process.c
   src/repl_str.c
+  src/shared_library.c
   src/snprintf.c
   src/split.c
   src/strdup.c
@@ -98,6 +99,8 @@ add_library(
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(${PROJECT_NAME} PRIVATE "RCUTILS_BUILDING_DLL")
+
+target_link_libraries(${PROJECT_NAME} ${CMAKE_DL_LIBS})
 
 # Needed if pthread is used for thread local storage.
 if(IOS AND IOS_SDK_VERSION LESS 10.0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,6 +315,21 @@ if(BUILD_TESTING)
     target_link_libraries(test_repl_str ${PROJECT_NAME})
   endif()
 
+  if(WIN32)
+    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
+  else()
+    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
+  endif()
+
+  ament_add_gtest(test_shared_library test/test_shared_library.cpp
+    APPEND_LIBRARY_DIRS "${append_library_dirs}"
+  )
+
+  if(TARGET test_shared_library)
+    add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.cpp)
+    target_link_libraries(test_shared_library ${PROJECT_NAME})
+  endif()
+
   rcutils_custom_add_gtest(test_time
     test/test_time.cpp
     ENV ${memory_tools_test_env_vars})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,6 +323,11 @@ if(BUILD_TESTING)
 
   if(TARGET test_shared_library)
     add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
+    if(WIN32)
+      # Causes the visibility macros to use dllexport rather than dllimport
+      # which is appropriate when building the dll but not consuming it.
+      target_compile_definitions(dummy_shared_library PRIVATE "DUMMY_SHARED_LIBRARY_BUILDING_DLL")
+    endif()
     target_link_libraries(test_shared_library ${PROJECT_NAME})
   endif()
 
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -130,6 +130,18 @@ RCUTILS_WARN_UNUSED
 rcutils_ret_t
 rcutils_unload_shared_library(rcutils_shared_library_t * lib);
 
+/// Get the library name for the compiled platform
+/**
+ * \param[in] library_name library base name (without prefix and extension)
+ * \param[out] library_name_platform library name for the compiled platform
+ * \return `RCUTILS_RET_OK` if successful, or
+ * \return `RCUTILS_RET_ERROR` if an unknown error occurs
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_ret_t
+rcutils_get_platform_library_name(const char * library_name, char * library_name_platform);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -49,7 +49,6 @@ typedef struct RCUTILS_PUBLIC_TYPE rcutils_shared_library_t
 /// Return an empty shared library struct.
 /*
  * This function returns an empty and zero initialized shared library struct.
- * The default allocator is set.
  *
  * Example:
  *
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -120,7 +120,7 @@ rcutils_has_symbol(const rcutils_shared_library_t * lib, const char * symbol_nam
 
 /// Unload the shared library.
 /**
- * \param[inout] lib rcutils_shared_library_t to be finalized
+ * \param[in] lib rcutils_shared_library_t to be finalized
  * \return `RCUTILS_RET_OK` if successful, or
  * \return `RCUTILS_RET_INVALID_ARGUMENT` for invalid arguments, or
  * \return `RCUTILS_RET_ERROR` if an unknown error occurs
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -56,13 +56,16 @@ typedef struct RCUTILS_PUBLIC_TYPE rcutils_shared_library_t
  * ```c
  * // Do not do this:
  * // rcutils_shared_library_t foo;
- * // rcutils_load_shared_library(&foo, "library_name"); // undefined behavior!
+ * // rcutils_load_shared_library(
+ * //     &foo,
+ * //    "library_name",
+ * //    rcutils_get_default_allocator()); // undefined behavior!
  * // or
  * // rcutils_unload_shared_library(&foo); // undefined behavior!
  *
  * // Do this instead:
  * rcutils_shared_library_t bar = rcutils_get_zero_initialized_shared_library();
- * rcutils_load_shared_library(&bar, "library_name"); // ok
+ * rcutils_load_shared_library(&bar, "library_name",rcutils_get_default_allocator()); // ok
  * void * symbol = rcutils_get_symbol(&bar, "bazinga"); // ok
  * bool is_bazinga_symbol = rcutils_has_symbol(&bar, "bazinga"); // ok
  * rcutils_unload_shared_library(&bar); // ok
@@ -77,6 +80,7 @@ rcutils_get_zero_initialized_shared_library(void);
 /**
  * \param[inout] lib struct with the shared library pointer and shared library path name
  * \param[in] library_path string with the path of the library
+ * \param[in] allocator to be used to allocate and deallocate memory
  * \return `RCUTILS_RET_OK` if successful, or
  * \return `RCUTILS_RET_BAD_ALLOC` if memory allocation fails, or
  * \return `RCUTILS_RET_ERROR` if an unknown error occurs, or
@@ -85,7 +89,10 @@ rcutils_get_zero_initialized_shared_library(void);
 RCUTILS_PUBLIC
 RCUTILS_WARN_UNUSED
 rcutils_ret_t
-rcutils_load_shared_library(rcutils_shared_library_t * lib, const char * library_path);
+rcutils_load_shared_library(
+  rcutils_shared_library_t * lib,
+  const char * library_path,
+  rcutils_allocator_t allocator);
 
 /// Return shared library symbol pointer.
 /**
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -25,12 +25,9 @@ extern "C"
 #ifndef _WIN32
 #include <dlfcn.h>
 typedef void * rcutils_shared_library_handle_t;
-# include <limits.h>
-# define RCUTILS_DIR_PATH_MAX PATH_MAX
 #else
 #include <windows.h>
 typedef HINSTANCE rcutils_shared_library_handle_t;
-# define RCUTILS_DIR_PATH_MAX MAX_PATH
 #endif  // _WIN32
 
 #include "rcutils/allocator.h"
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -56,7 +56,7 @@ typedef struct RCUTILS_PUBLIC_TYPE rcutils_shared_library_t
  * // Do not do this:
  * // rcutils_shared_library_t foo;
  * // rcutils_allocator_t allocator = rcutils_get_default_allocator();
- * // rcutils_unload_library(&foo, allocator); // undefined behavior!
+ * // rcutils_unload_shared_library(&foo, allocator); // undefined behavior!
  *
  * // Do this instead:
  * rcutils_shared_library_t bar = rcutils_get_zero_initialized_shared_library();
@@ -115,7 +115,7 @@ rcutils_has_symbol(const rcutils_shared_library_t * lib, const char * symbol_nam
 RCUTILS_PUBLIC
 RCUTILS_WARN_UNUSED
 rcutils_ret_t
-rcutils_unload_library(rcutils_shared_library_t * lib);
+rcutils_unload_shared_library(rcutils_shared_library_t * lib);
 
 #ifdef __cplusplus
 }
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -55,19 +55,22 @@ typedef struct RCUTILS_PUBLIC_TYPE rcutils_shared_library_t
  * ```c
  * // Do not do this:
  * // rcutils_shared_library_t foo;
- * // rcutils_load_shared_library(
+ * // rcutils_ret_t ret = rcutils_load_shared_library(
  * //     &foo,
  * //    "library_name",
  * //    rcutils_get_default_allocator()); // undefined behavior!
  * // or
- * // rcutils_unload_shared_library(&foo); // undefined behavior!
+ * // rcutils_ret_t ret = rcutils_unload_shared_library(&foo); // undefined behavior!
  *
  * // Do this instead:
  * rcutils_shared_library_t bar = rcutils_get_zero_initialized_shared_library();
  * rcutils_load_shared_library(&bar, "library_name", rcutils_get_default_allocator()); // ok
  * void * symbol = rcutils_get_symbol(&bar, "bazinga"); // ok
  * bool is_bazinga_symbol = rcutils_has_symbol(&bar, "bazinga"); // ok
- * rcutils_unload_shared_library(&bar); // ok
+ * rcutils_ret_t ret = rcutils_unload_shared_library(&bar); // ok
+ * if (ret != RCUTILS_RET_ERROR) {
+ *   // error handling
+ * }
  * ```
  * */
 RCUTILS_PUBLIC
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -24,53 +24,98 @@ extern "C"
 
 #ifndef _WIN32
 #include <dlfcn.h>
+typedef void * shared_library_type;
 #else
 #include <windows.h>
+typedef HINSTANCE shared_library_type;
 #endif  // _WIN32
 
+#include "rcutils/allocator.h"
+#include "rcutils/types/rcutils_ret.h"
 #include "rcutils/macros.h"
 #include "rcutils/visibility_control.h"
 
-/// The structure
-typedef struct rcutils_shared_library_t
+/// Handle to a loaded shared library.
+typedef struct RCUTILS_PUBLIC_TYPE rcutils_shared_library_t
 {
-  /// The pointer to the shared library.
-  #ifndef _WIN32
-  void * lib_pointer;
-  #else
-  HINSTANCE lib_pointer;
-  #endif
+  /// The pointer to the shared library
+  shared_library_type lib_pointer;
   /// The path of the shared_library
   char * library_path;
 } rcutils_shared_library_t;
 
+/// Return an empty shared library struct.
+/*
+ * This function returns an empty and zero initialized shared library struct.
+ *
+ * Example:
+ *
+ * ```c
+ * // Do not do this:
+ * // rcutils_shared_library_t foo;
+ * // rcutils_allocator_t allocator = rcutils_get_default_allocator();
+ * // rcutils_unload_library(&foo, allocator); // undefined behavior!
+ *
+ * // Do this instead:
+ * rcutils_shared_library_t bar = rcutils_get_zero_initialized_shared_library();
+ * rcutils_allocator_t allocator = rcutils_get_default_allocator();
+ * rcutils_unload_library(&bar, allocator); // ok
+ * ```
+ * */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_shared_library_t
+rcutils_get_zero_initialized_shared_library(void);
+
 /// Return shared library pointer.
 /**
- * \param[in] library_path path to the library
- * \return void* shared library pointer.
- *                nullptr if library doesn't exist
+ *
+ * the memory of the library path inside the rcutils_shared_library_t struct should be
+ * reserved and defined outside this function
+ *
+ * \param[in] lib struct with the shared library pointer and shared library path name
+ * \return `RCUTILS_RET_OK` if successful, or
+ * \return `RCUTILS_RET_ERROR` if an unknown error occurs
  */
 RCUTILS_PUBLIC
-rcutils_shared_library_t *
-rcutils_get_shared_library(const char * library_path);
+RCUTILS_WARN_UNUSED
+rcutils_ret_t
+rcutils_get_shared_library(rcutils_shared_library_t * lib);
 
 /// Return shared library symbol pointer.
 /**
+ * \param[in] lib struct with the shared library pointer and shared library path name
  * \param[in] symbol_name name of the symbol inside the shared library
- * \return void* symbol pointer.
- *                nullptr if symbol doesn't exist
+ * \return shared library symbol pointer.
  */
 RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
 void *
 rcutils_get_symbol(rcutils_shared_library_t * lib, const char * symbol_name);
 
-/// Unload the shared library.
+/// Return if the shared library contains a specific symbolname .
 /**
- * \param[in] lib struct to the shared library
+ * \param[in] lib struct with the shared library pointer and shared library path name
+ * \param[in] symbol_name name of the symbol inside the shared library
+ * \return returns true on success, and false otherwise.
  */
 RCUTILS_PUBLIC
-void
-rcutils_unload_library(rcutils_shared_library_t * lib);
+RCUTILS_WARN_UNUSED
+bool
+rcutils_has_symbol(rcutils_shared_library_t * lib, const char * symbol_name);
+
+/// Unload the shared library.
+/**
+ * \param[inout] lib rcutils_shared_library_t to be finalized
+ * \param[in] allocator the allocator to use for allocation
+ * \return `RCUTILS_RET_OK` if successful, or
+ * \return `RCUTILS_RET_INVALID_ARGUMENT` for invalid arguments, or
+ * \return `RCUTILS_RET_ERROR` if an unknown error occurs
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_ret_t
+rcutils_unload_library(rcutils_shared_library_t * lib, rcutils_allocator_t allocator);
 
 #ifdef __cplusplus
 }
--- a/include/rcutils/shared_library.h
+++ b/include/rcutils/shared_library.h
@@ -25,9 +25,12 @@ extern "C"
 #ifndef _WIN32
 #include <dlfcn.h>
 typedef void * rcutils_shared_library_handle_t;
+# include <limits.h>
+# define RCUTILS_DIR_PATH_MAX PATH_MAX
 #else
 #include <windows.h>
 typedef HINSTANCE rcutils_shared_library_handle_t;
+# define RCUTILS_DIR_PATH_MAX MAX_PATH
 #endif  // _WIN32
 
 #include "rcutils/allocator.h"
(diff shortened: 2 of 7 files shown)


src/shared_library.c
```diff
@@ -0,0 +1,93 @@
...
+  }
+#endif  // _WIN32
+  return lib_symbol;
+}
+
+void
+rcutils_unload_library(rcutils_shared_library_t * lib)
+{
+#ifndef _WIN32
+  dlclose(lib->lib_pointer);
```
dirk-thomas: The return code should be checked and probably be exposed.

src/shared_library.c
```diff
@@ -0,0 +1,93 @@
...
+  free(lib->library_path);
+  free(lib);
+#else
+  FreeLibrary(lib->lib_pointer);
+  free(lib->library_path);
+  free(lib);
```
dirk-thomas: Nitpick: these two lines are platform independent and should be deduplicated / moved outside of the conditional blocks.

src/shared_library.c
```diff
@@ -0,0 +1,93 @@
...
+rcutils_get_symbol(rcutils_shared_library_t * lib, const char * symbol_name)
+{
+#ifndef _WIN32
+  void * lib_symbol = dlsym(lib->lib_pointer, symbol_name);
+  const char * dlsym_error = dlerror();
+  if (dlsym_error) {
+    fprintf(
+      stderr, "Cannot load symbol '%s' in shared library '%s'",
+      symbol_name, lib->library_path);
+    dlclose(lib->lib_pointer);
```
dirk-thomas: Why should the library be unloaded in this case? Maybe the caller only wants to check if a specific symbol is available.

For the very same reason I don't think the function should print to `stderr`. Instead the error message should be exposed through API if the calling code wants to retrieve it.

src/shared_library.c
```diff
@@ -0,0 +1,93 @@
...
+#include "rcutils/shared_library.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+
+rcutils_shared_library_t *
+rcutils_get_shared_library(const char * library_path)
+{
+  rcutils_shared_library_t * lib = (rcutils_shared_library_t *)
+    malloc(sizeof(rcutils_shared_library_t));
```
dirk-thomas: Should this support a custom allocator?

Same below.

src/shared_library.c
```diff
@@ -0,0 +1,93 @@
...
+#else
+  lib->lib_pointer = LoadLibrary(library_path);
+#endif  // _WIN32
+  if (!lib->lib_pointer) {
+    free(lib);
+    return NULL;
+  }
+
+  // +1 to accomodate for the null terminator
+  lib->library_path = (char *) malloc((strlen(library_path) + 1) * sizeof(char));
```
dirk-thomas: Return value must be checked.

CMakeLists.txt
```diff
@@ -99,6 +100,8 @@ add_library(
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(${PROJECT_NAME} PRIVATE "RCUTILS_BUILDING_DLL")
 
+target_link_libraries(${PROJECT_NAME} ${CMAKE_DL_LIBS})
```
dirk-thomas: Wouldn't `CMAKE_DL_LIBS` need to be exported to downstream packages?

src/shared_library.c
ahcorde: I can create another method called `rcutils_library_has_symbol`, what do you think?

src/shared_library.c
```diff
@@ -0,0 +1,91 @@
...
+
+void
+rcutils_unload_library(rcutils_shared_library_t * lib)
+{
+#ifndef _WIN32
+  dlclose(lib->lib_pointer);
+#else
+  FreeLibrary(lib->lib_pointer);
+#endif  // _WIN32
+  free(lib->library_path);
```
dirk-thomas: NItpick: this pointer should be set to `0` just in case the caller tries to access it after the fact.

src/shared_library.c
dirk-thomas: That doesn't change the fact that this function shouldn't unload the library. And with that the current function can already be used for that exact use case by checking if the returned pointer is `0` or not.

src/shared_library.c
ahcorde: There is no code error in dlclose()

> The function dlclose() decrements the reference count on the dynamic library handle handle. If the reference count drops to zero and no other loaded libraries use symbols in it, then the dynamic library is unloaded.

---- 

### Windows

> If the function succeeds, the return value is nonzero.
> If the function fails, the return value is zero. To get extended error information, call the GetLastError function.

But I can do something in Windows

dirk-thomas: You might want to also cite the next sentence after paragraph you copy-n-pasted (https://linux.die.net/man/3/dlclose):

> The function dlclose() returns 0 on success, and nonzero on error.

wjwwood: According to https://linux.die.net/man/3/dlclose there is a non-zero return code on error with `dlclose()`:

> dlclose()
>
> The function dlclose() decrements the reference count on the dynamic library handle handle. If the reference count drops to zero and no other loaded libraries use symbols in it, then the dynamic library is unloaded.
>
> The function dlclose() returns 0 on success, and nonzero on error. 

And then there is `dlerror` to check the error:

> dlerror()
>
> The function dlerror() returns a human readable string describing the most recent error that occurred from dlopen(), dlsym() or dlclose() since the last call to dlerror(). It returns NULL if no errors have occurred since initialization or since it was last called. 

So I think it should check the return code, set the error state using `RCUTILS_SET_ERROR_MSG(msg)` and then return an `rcutils_ret_t` in the event that a non-zero is returned from `dlclose()`.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+
+/// The structure
+typedef struct rcutils_shared_library_t
+{
+  /// The pointer to the shared library.
+  #ifndef _WIN32
+  void * lib_pointer;
+  #else
+  HINSTANCE lib_pointer;
+  #endif
```
wjwwood: Can you use a typedef for the handle type and then only declare `lib_pointer` once?

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+  /// The path of the shared_library
+  char * library_path;
+} rcutils_shared_library_t;
+
+/// Return shared library pointer.
+/**
+ * \param[in] library_path path to the library
+ * \return void* shared library pointer.
+ *                nullptr if library doesn't exist
```
wjwwood: If it returns various values you can have multiple `\return` entries, which I would prefer in this case, e.g.:

https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/time.h#L68-L70

wjwwood: Also, a nitpick about the style, when wrapping these, please only indent two spaces, e.g. I would rewrite this:

```
 * \return type pointer to a shared library handle, or
 *   nullptr if the library does not exist.
```

Please also note consistent use of punctuation.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+  #else
+  HINSTANCE lib_pointer;
+  #endif
+  /// The path of the shared_library
+  char * library_path;
+} rcutils_shared_library_t;
+
+/// Return shared library pointer.
+/**
+ * \param[in] library_path path to the library
```
wjwwood: It would be nice to know:

- does this string get copied into the returned structure?
  - if so how is it allocated?
- how is the returned structure allocated?

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+ * \return void* shared library pointer.
+ *                nullptr if library doesn't exist
+ */
+RCUTILS_PUBLIC
+rcutils_shared_library_t *
+rcutils_get_shared_library(const char * library_path);
```
wjwwood: I think it would be better to have this signature:

```c
rcutils_ret_t
rcutils_get_shared_library(rcutils_shared_library_t * shared_library, const char * library_path);
```

For two reasons:

- this puts the effort of allocating the `rcutils_shared_library_t` onto the user (they can choose heap or stack)
- it allows you to return more precise error codes if that makes sense

wjwwood: If you do this, I recommend a function like `rcutils_shared_library_t rcutils_get_zero_initialized_shared_library()` which the user may use to ensure the struct is initialized (even if we add a field in the future). For example:

https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/types/string_map.h#L55-L56

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+RCUTILS_PUBLIC
+void *
+rcutils_get_symbol(rcutils_shared_library_t * lib, const char * symbol_name);
+
+/// Unload the shared library.
+/**
+ * \param[in] lib struct to the shared library
+ */
+RCUTILS_PUBLIC
+void
```
wjwwood: As discussed elsewhere, this function can fail and in which case it should return a `rcutils_ret_t` most likely.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+rcutils_get_shared_library(const char * library_path);
+
+/// Return shared library symbol pointer.
+/**
+ * \param[in] symbol_name name of the symbol inside the shared library
+ * \return void* symbol pointer.
+ *                nullptr if symbol doesn't exist
+ */
+RCUTILS_PUBLIC
+void *
```
wjwwood: Similarly, this should probably be moved into the arguments as an output argument (`void **`) and `rcutils_ret_t` returned here, unless there are no known error states, in which case returning `nullptr` can stay.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+#ifndef _WIN32
+#include <dlfcn.h>
+#else
+#include <windows.h>
+#endif  // _WIN32
+
+#include "rcutils/macros.h"
+#include "rcutils/visibility_control.h"
+
+/// The structure
```
wjwwood: This doesn't tell me much... How about:

`/// Handle to a loaded shared library.`

src/shared_library.c
wjwwood: 👍, if allocation cannot be avoided, then an `rcutils_allocator_t` should be one of the arguments.

src/shared_library.c
wjwwood: > I can create another method called rcutils_library_has_symbol, what do you think?

That's probably useful.

> That doesn't change the fact that this function shouldn't unload the library.

👍

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,79 @@
...
+  char * library_path;
+} rcutils_shared_library_t;
+
+/// Return shared library pointer.
+/**
+ * \param[in] library_path path to the library
+ * \return void* shared library pointer.
+ *                nullptr if library doesn't exist
+ */
+RCUTILS_PUBLIC
```
wjwwood: You should probably have `RCUTILS_WARN_UNUSED` on these as well.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,124 @@
...
+/// Return if the shared library contains a specific symbolname .
+/**
+ * \param[in] lib struct with the shared library pointer and shared library path name
+ * \param[in] symbol_name name of the symbol inside the shared library
+ * \return returns true on success, and false otherwise.
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+bool
+rcutils_has_symbol(rcutils_shared_library_t * lib, const char * symbol_name);
```
wjwwood: Should `lib` be `const` here? Maybe not, I just don't know, but it seems like a lookup method could be const.

src/shared_library.c
```diff
@@ -0,0 +1,120 @@
...
+rcutils_ret_t
+rcutils_get_shared_library(rcutils_shared_library_t * lib)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib->library_path, RCUTILS_RET_INVALID_ARGUMENT);
+
+#ifndef _WIN32
+  lib->lib_pointer = dlopen(lib->library_path, RTLD_LAZY);
+  if (!lib->lib_pointer) {
+    RCUTILS_SET_ERROR_MSG(dlerror);
```
wjwwood: Whoa this seems wrong, maybe:

```suggestion
    RCUTILS_SET_ERROR_MSG(dlerror());
```

src/shared_library.c
```diff
@@ -0,0 +1,120 @@
...
+    return RCUTILS_RET_ERROR;
+  }
+#else
+  lib->lib_pointer = LoadLibrary(lib->library_path);
+  if (!lib->lib_pointer) {
+    DWORD error = GetLastError();
```
wjwwood: This is not used?

src/shared_library.c
```diff
@@ -0,0 +1,120 @@
...
+  }
+#endif  // _WIN32
+  return RCUTILS_RET_OK;
+}
+
+void *
+rcutils_get_symbol(rcutils_shared_library_t * lib, const char * symbol_name)
+{
+#ifndef _WIN32
+  void * lib_symbol = dlsym(lib->lib_pointer, symbol_name);
```
wjwwood: What if `lib` or `symbol_name` is nullptr?

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,124 @@
...
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_shared_library_t
+rcutils_get_zero_initialized_shared_library(void);
+
+/// Return shared library pointer.
+/**
+ *
+ * the memory of the library path inside the rcutils_shared_library_t struct should be
+ * reserved and defined outside this function
```
wjwwood: This is really asymmetric behavior, the user has to allocate the string, but the unload library call deallocates the string. Either unload should not deallocate the string or (my preference) this function should take an allocator, copy the string using the allocator, store the allocator in the `rcutils_shared_library_t`, unload should no longer take an allocator, and unload can deallocate the string using the original allocator passed via the this function. That way the user doesn't have to allocate the string and we ensure the string is allocated and deallocated with the same allocator.

ahcorde: @wjwwood Done ;)

src/shared_library.c
```diff
@@ -0,0 +1,142 @@
...
+
+#ifndef _WIN32
+  void * lib_symbol = dlsym(lib->lib_pointer, symbol_name);
+#else
+  void * lib_symbol = GetProcAddress(lib->lib_pointer, symbol_name);
+#endif  // _WIN32
+  if (!lib_symbol) {
+    return false;
+  }
+  return true;
```
dirk-thomas: There is no need for an `if` - simply `return lib_symbol != 0`.

src/shared_library.c
```diff
@@ -0,0 +1,142 @@
...
+  // The function dlclose() returns 0 on success, and nonzero on error.
+  int error_code = dlclose(lib->lib_pointer);
+  if (error_code) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("dlclose error: %s", dlerror());
+    ret = RCUTILS_RET_ERROR;
+  }
+#else
+  // zero if the function succeeds
+  int error_code = FreeLibrary(lib->lib_pointer);
+  if (!error_code) {
```
dirk-thomas: This condition doesn't match the comment. If the function succeeds and returns zero this expression evaluates to `true` and signals an error.

src/shared_library.c
ahcorde: Fixed https://github.com/ros2/rcutils/pull/215/commits/e07d449bb638bb08661d73969fabb5b9246c9b04

src/shared_library.c
ahcorde: The comment was wrong. Fixed https://github.com/ros2/rcutils/pull/215/commits/e07d449bb638bb08661d73969fabb5b9246c9b04

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,124 @@
...
+ * ```
+ * */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_shared_library_t
+rcutils_get_zero_initialized_shared_library(void);
+
+/// Return shared library pointer.
+/**
+ * \param[inout] lib struct with the shared library pointer and shared library path name
```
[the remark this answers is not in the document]
ahcorde: I assign the allocator in the `rcutils_get_zero_initialized_shared_library`.

https://github.com/ros2/rcutils/blob/55fdb7d9d436236b942daabb908a5137e162ba4e/src/shared_library.c#L27-L34

Is it correct? or should we assign in the lib argument before passing it ?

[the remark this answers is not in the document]
wjwwood: I mean if they want to override it. Other functions instead zero initialize the allocator too, and the init function will take an allocator and assign it. Basically the user would never access the allocator member of the struct directly even if providing a custom one.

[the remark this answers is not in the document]
ahcorde: your suggestions it's to remove this line:

```cpp
zero_initialized_shared_library.allocator = rcutils_get_default_allocator();
```

and then change the headers of this two functions?

```cpp
rcutils_load_shared_library(
      rcutils_shared_library_t * lib, 
      const char * library_path);
rcutils_load_shared_library(
      rcutils_shared_library_t * lib,
      const char * library_path);
```
to 
```cpp
rcutils_load_shared_library(
      rcutils_shared_library_t * lib,
      const char * library_path,
rcutils_load_shared_library(
      rcutils_shared_library_t * lib,
      const char * library_path,
```

is it correct?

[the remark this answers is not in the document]
wjwwood: Yes, that's close to what I'd propose:

```c
rcutils_ret_t
rcutils_load_shared_library(
  rcutils_shared_library_t * library,
  const char * library_path,

rcutils_ret_t
rcutils_unload_shared_library(
  rcutils_shared_library_t * library);
```

To be used as such:

```c
rcutils_shared_library_t library = rcutils_get_zero_initialized_shared_library();

rcutils_ret_t ret = rcutils_load_shared_library(&library, "...", rcutils_get_default_allocator());
if (RCUTILS_RET_OK != ret) {
  // error handling
}

// ... later ...

ret = rcutils_unload_shared_library(&library);
if (RCUTILS_RET_OK != ret) {
  // error handling
}
```

CMakeLists.txt
```diff
@@ -312,6 +315,21 @@ if(BUILD_TESTING)
     target_link_libraries(test_repl_str ${PROJECT_NAME})
   endif()
 
+  if(WIN32)
+    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
+  else()
+    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
```
dirk-thomas: Does this need to be a conditional or does `TARGET_FILE_DIR` also work on non-Windows systems?

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+#include "rcutils/shared_library.h"
+#include "rcutils/strdup.h"
+
+rcutils_shared_library_t
+rcutils_get_zero_initialized_shared_library(void)
+{
+  rcutils_shared_library_t zero_initialized_shared_library;
+  zero_initialized_shared_library.library_path = NULL;
+  zero_initialized_shared_library.lib_pointer = NULL;
+  return zero_initialized_shared_library;
```
dirk-thomas: Allocator still needs to be zero initialized here.

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+rcutils_ret_t
+rcutils_load_shared_library(
+  rcutils_shared_library_t * lib,
+  const char * library_path,
+  rcutils_allocator_t allocator)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib, RCUTILS_RET_INVALID_ARGUMENT);
```
dirk-thomas: The function should also check `library_path`.

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+
+  lib->allocator = allocator;
+
+  lib->library_path = rcutils_strdup(library_path, lib->allocator);
```
dirk-thomas: It needs to be either checked that `lib->library_path` is `NULL` before or it needs to be deallocated before.

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+  }
+#else
+  lib->lib_pointer = LoadLibrary(lib->library_path);
+  if (!lib->lib_pointer) {
+    lib->allocator.deallocate(lib->library_path, lib->allocator.state);
+    lib->library_path = NULL;
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("LoadLibrary error: %lu", GetLastError());
+    return RCUTILS_RET_ERROR;
+  }
+#endif  // _WIN32
```
dirk-thomas: Significant parts in the `#ifndef` and `#else` are identifical. Please deduplicate.

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+  return GetLastError() == 0 && lib_symbol != 0;
+#endif  // _WIN32
+}
+
+rcutils_ret_t
+rcutils_unload_shared_library(rcutils_shared_library_t * lib)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib->lib_pointer, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib->library_path, RCUTILS_RET_INVALID_ARGUMENT);
```
dirk-thomas: Since the function later uses `lib->allocator` it should check that it isn't empty.

src/shared_library.c
```diff
@@ -0,0 +1,158 @@
...
+    ret = RCUTILS_RET_ERROR;
+  }
+#else
+  // If the function succeeds, the return value is nonzero.
+  int error_code = FreeLibrary(lib->lib_pointer);
+  if (!error_code) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("FreeLibrary error: %lu", GetLastError());
+    ret = RCUTILS_RET_ERROR;
+  }
+#endif  // _WIN32
```
dirk-thomas: Please deduplicate the parts which are the same.

test/test_shared_library.cpp
```diff
@@ -0,0 +1,88 @@
...
+  symbol = rcutils_get_symbol(&lib, "symbol");
+  EXPECT_TRUE(symbol == NULL);
+
+  symbol = rcutils_get_symbol(&lib, "print_name");
+  EXPECT_TRUE(symbol != NULL);
+
+  // unload shared_library
+  ret = rcutils_unload_shared_library(&lib);
+  ASSERT_EQ(RCUTILS_RET_OK, ret);
+}
```
dirk-thomas: The tests should also exercise non-standard sequences of function calls, e.g. load twice, get_symbol before loading, unload twice, etc.

CMakeLists.txt
ahcorde: I checked `class_loader` and `pluginlib`. These lines come from there. I tried on my Windows machine and it's the way to make it work.

I will try again to remove this part: `TARGET_FILE_DIR`

dirk-thomas: You will certainly need the generator expression on Windows. I was asking if the same also works on non-Windows so we don't need a conditional choice here.

src/shared_library.c
ahcorde: the only duplicate line between the `if` statements is
```
ret = RCUTILS_RET_ERROR;
```

dirk-thomas: And the `if` and the closing parenthesis and the definition of the `error_code`...

wjwwood: The `if` is not duplicated? The "declaration" of the `error_code` is the same (`int`), but the "definition" is not (as I understand the terms in this case), one calls `dlclose` and the other calls `FreeLibrary`.

I actually think this code is fine as-is and needs no more deduplication.

test/test_shared_library.cpp
ahcorde: I added more tests https://github.com/ros2/rcutils/pull/215/commits/791214afc1a389a57bde249956c9efdd52c77676

ahcorde:
friendly ping @dirk-thomas @wjwwood

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,134 @@
...
+ * // rcutils_load_shared_library(
+ * //     &foo,
+ * //    "library_name",
+ * //    rcutils_get_default_allocator()); // undefined behavior!
+ * // or
+ * // rcutils_unload_shared_library(&foo); // undefined behavior!
+ *
+ * // Do this instead:
+ * rcutils_shared_library_t bar = rcutils_get_zero_initialized_shared_library();
+ * rcutils_load_shared_library(&bar, "library_name",rcutils_get_default_allocator()); // ok
```
wjwwood:
```suggestion
 * rcutils_load_shared_library(&bar, "library_name", rcutils_get_default_allocator()); // ok
```

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,134 @@
...
+ * void * symbol = rcutils_get_symbol(&bar, "bazinga"); // ok
+ * bool is_bazinga_symbol = rcutils_has_symbol(&bar, "bazinga"); // ok
+ * rcutils_unload_shared_library(&bar); // ok
```
wjwwood: You should show the capture and checking of the return type from `rcutils_unload_shared_library()`.

src/shared_library.c
```diff
@@ -0,0 +1,157 @@
...
+  if (lib_symbol == NULL) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Error getting the symbol '%s'. Error '%d'",
+      symbol_name, GetLastError());
+    return NULL;
+  }
+#endif  // _WIN32
+  if (!lib_symbol) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "symbol '%s' doesnt' exit in library '%s'",
```
wjwwood:
```suggestion
      "symbol '%s' does not exist in the library '%s'",
```

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,137 @@
...
+
+/// Return shared library pointer.
+/**
+ * \param[inout] lib struct with the shared library pointer and shared library path name
+ * \param[in] library_path string with the path of the library
+ * \param[in] allocator to be used to allocate and deallocate memory
+ * \return `RCUTILS_RET_OK` if successful, or
+ * \return `RCUTILS_RET_BAD_ALLOC` if memory allocation fails, or
+ * \return `RCUTILS_RET_ERROR` if an unknown error occurs, or
+ * \return `RCUTILS_RET_INVALID_ARGUMENT` for invalid arguments
```
dirk-thomas: We usually use a single `\return` even if the text is wrapped into multiple lines.

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,137 @@
...
+ * \return if symbols exists returns true, otherwise returns false.
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+bool
+rcutils_has_symbol(const rcutils_shared_library_t * lib, const char * symbol_name);
+
+/// Unload the shared library.
+/**
+ * \param[inout] lib rcutils_shared_library_t to be finalized
```
dirk-thomas: How is `lib` an "out" argument? While the content of lib will be updated the pointer of lib can't be changed on the caller side.

include/rcutils/shared_library.h
wjwwood: That’s not true...

dirk-thomas: You seem to be very certain that the number of times multiple `\return` statements are used is higher than the number of times we use a single statement and still wrap the content. I am not going to count - marked as resolved.

include/rcutils/shared_library.h
ahcorde: Fixed af0d73e87237267b1236f690468357f22544ace0

test/test_shared_library.cpp
dirk-thomas: E.g. the case of calling `rcutils_load_shared_library()` twice isn't covered yet. And I assume the current implementation would leak memory in that case.

ahcorde: Avoided the leak https://github.com/ros2/rcutils/pull/215/commits/8dc66df8868220f0d3052d1b3e560ac20275b1dc and added the test

src/shared_library.c
```diff
@@ -0,0 +1,161 @@
...
+  rcutils_allocator_t allocator)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(lib, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(library_path, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RCUTILS_RET_INVALID_ARGUMENT);
+
+  lib->allocator = allocator;
+
+  if (lib->library_path != NULL) {
+    lib->allocator.deallocate(lib->library_path, lib->allocator.state);
```
dirk-thomas: The deallocate needs to happen before the allocator is being overwritten. Otherwise the deallocator might happen with a different allocator than the one which allocated the string.

ahcorde: fixed the order https://github.com/ros2/rcutils/pull/215/commits/e154f9c0a4556f45d930f99a824ac8ebee57c21b

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9837/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5454/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8000/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9693/)

test/test_shared_library.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // running test has failed.
+    rcutils_reset_error();
+    lib = rcutils_get_zero_initialized_shared_library();
+
+#ifdef __linux__
+    library_path = std::string("libdummy_shared_library.so");
+#elif __APPLE__
+    library_path = std::string("libdummy_shared_library.dylib");
+#elif _WIN32
+    library_path = std::string("dummy_shared_library.dll");
```
dirk-thomas: For calling code having to know about these platform specific naming details makes the API rather difficult to use. A function to provide the platform specific library name when passing in this case `dummy_shared_library` would be helpful.

ahcorde: Added a function to provide the platform specific library name https://github.com/ros2/rcutils/pull/215/commits/d941e44755bffda67808ca955b753c445cfcffb4

include/rcutils/shared_library.h
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * \param[in] library_name library base name (without prefix and extension)
+ * \param[out] library_name_platform library name for the compiled platform
+ * \return `RCUTILS_RET_OK` if successful, or
+ * \return `RCUTILS_RET_ERROR` if an unknown error occurs
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_ret_t
+rcutils_get_platform_library_name(const char * library_name, char * library_name_platform);
```
dirk-thomas: The API doesn't define how large `library_name_platform` is so this can easily result in writing into wrong memory.

src/shared_library.c
```diff
@@ -0,0 +1,185 @@
...
+rcutils_get_platform_library_name(const char * library_name, char * library_name_platform)
+{
+  int written = 0;
+
+#ifdef __linux__
+  written = rcutils_snprintf(
+    library_name_platform, strlen(library_name) + 7, "lib%s.so", library_name);
+#elif __APPLE__
+  written = rcutils_snprintf(
+    library_name_platform, strlen(library_name) + 8, "lib%s.dylib", library_name);
```
dirk-thomas: Why 8? Shouldn't it be 10?

CMakeLists.txt
```diff
@@ -312,6 +315,17 @@ if(BUILD_TESTING)
...
   endif()
 
+  set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
+
+  ament_add_gtest(test_shared_library test/test_shared_library.cpp
+    APPEND_LIBRARY_DIRS "${append_library_dirs}"
+  )
+
+  if(TARGET test_shared_library)
+    add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
```
wjwwood: Here is where you'd need to define `DUMMY_SHARED_LIBRARY_BUILDING_DLL`..

ahcorde: thank you @wjwwood https://github.com/ros2/rcutils/pull/215/commits/66187a37d0db7787aa20d48104b451dc8a10f6cb

include/rcutils/shared_library.h
ahcorde: fixed https://github.com/ros2/rcutils/pull/215/commits/510bf355f0cee666d1c096075cae1e6b255babbd

src/shared_library.c
ahcorde: fixed https://github.com/ros2/rcutils/pull/215/commits/26b7157ac4e13d3574fd97a793b6d082aaa05a6c

src/shared_library.c
```diff
@@ -0,0 +1,204 @@
...
+rcutils_ret_t
+rcutils_get_platform_library_name(
+  const char * library_name,
+  char * library_name_platform,
+  unsigned int buffer_size)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(library_name, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(library_name_platform, RCUTILS_RET_INVALID_ARGUMENT);
+
+  if (strlen(library_name) > RCUTILS_DIR_PATH_MAX)
```
dirk-thomas, line 168: What is driving this limitation?

src/shared_library.c
```diff
@@ -0,0 +1,204 @@
...
+  {
+    RCUTILS_SET_ERROR_MSG(
+      "library_name is too long more than the maximum allowed size\n");
+    return RCUTILS_RET_ERROR;
+  }
+
+  int written = 0;
+
+#ifdef __linux__
+  if (buffer_size > (strlen(library_name) + 7)) {
```
dirk-thomas, line 178: Why does the buffer need to be larger and equal size isn't enough?

src/shared_library.c
ahcorde: I was thinking about the full path of the library but it's just the name. Removed

src/shared_library.c
ahcorde: fixed https://github.com/ros2/rcutils/pull/215/commits/5cc468ea931e1ed07a1c4c973bc181904820b8cf

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9870/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5486/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9726/)

PR #218, opened 2020-03-27
Added directory and file utils to replace tinydir

ahcorde:
This PR is part of the effort to reduce the number of dependencies in ROS 2. Tinydir is only used in RCL in [one place](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/security_directory.c#L87) to find a folder that match with a string.

I have created a new file `directory_and_file_reader.[hc]` maybe I can include this code in `filesystem.[hc]`.

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/include/rcutils/directory_and_file_reader.h
+++ b/include/rcutils/directory_and_file_reader.h
@@ -1,0 +1,198 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__DIRECTORY_AND_FILE_READER_H_
+#define RCUTILS__DIRECTORY_AND_FILE_READER_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcutils/allocator.h"
+#include "rcutils/types/rcutils_ret.h"
+#include "rcutils/macros.h"
+#include "rcutils/visibility_control.h"
+
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#ifndef _WIN32
+#include <dirent.h>
+#include <limits.h>
+# define RCUTILS_DIR_PATH_MAX PATH_MAX
+#else
+#include <windows.h>
+#include <tchar.h>
+#include <direct.h>
+# define RCUTILS_DIR_PATH_MAX MAX_PATH
+#endif  // _WIN32
+
+#ifndef RCUTILS_DIR_PATH_MAX
+#define RCUTILS_DIR_PATH_MAX 4096
+#endif
+#define RCUTILS_FILENAME_MAX 256
+/// Handle directories library.
+
+/// Handle files.
+typedef struct rcutils_file_t
+{
+  /// path of the file
+  char * path;
+  /// filename
+  char * name;
+  /// is the file a directory ?
+  int is_dir;
+  //  information about the file
+  #ifndef _WIN32
+  struct stat _s;
+  #else
+  struct _stat _s;
+  #endif
+} rcutils_file_t;
+
+/// Handle directories.
+typedef struct rcutils_dir_t
+{
+  /// path of the directory
+  char * path;
+  // is there any file in the folder.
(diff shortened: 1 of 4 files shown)


mjcarroll:
Just a note that this shouldn't have a major documentation impact as `tinydir` is currently built/installed via a vendor package.

include/rcutils/directory_and_file_reader.h
```diff
@@ -0,0 +1,198 @@
...
+/*
+ * This function returns an empty and zero initialized directory struct.
+ *
+ * Example:
+ * // Do not do this:
+ * // rcutils_dir_t foo;
+ * // rcutils_open_dir(&foo, "directory", rcutils_get_default_allocator()); // undefined behavior!
+ * // rcutils_ret_t ret = retrcutils_next_dir(&foo); // undefined behavior!
+ *
+ * // Do this instead:
```
mjcarroll: I would lead with the correct example, then provide negative examples.  I'm guilty of copying the first bit of code I see in the documentation.

include/rcutils/directory_and_file_reader.h
```diff
@@ -0,0 +1,198 @@
...
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+rcutils_dir_t
+rcutils_get_zero_initialized_dir(void);
+
+/// Return an empty file struct.
+/*
+ * This function returns an empty and zero initialized file struct.
+ *
+ * Example:
```
mjcarroll: Same here

dirk-thomas:
>  Does this need additional test coverage?

The new API certainly needs test coverage.

src/directory_and_file_reader.c
```diff
@@ -0,0 +1,249 @@
...
+  zero_initialized_file.path = NULL;
+  zero_initialized_file.name = NULL;
+  zero_initialized_file.is_dir = 0;
+  return zero_initialized_file;
+}
+
+void
+rcutils_file_fini(rcutils_file_t * file, rcutils_allocator_t allocator)
+{
+  allocator.deallocate(file->path, allocator.state);
```
dirk-thomas: The function needs to check that the `file` argument is valid. And if it is not it would be good to be able to signal that error using a return value.

ahcorde:
Added test coverage https://github.com/ros2/rcutils/pull/218/commits/98a64ec405bf89e66c0bc85617657f449b773269

include/rcutils/directory_and_file_reader.h
ahcorde: changed the order of the examples https://github.com/ros2/rcutils/pull/218/commits/1ff3651bb7398794bee1da36510f5d1abe28bbfa

ahcorde:
According with the comment of @mikaelarguedas in rcl https://github.com/ros2/rcl/pull/605#issuecomment-606700678 this code probably makes no sense.

@dirk-thomas should we close both PRs?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9838/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5455/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8001/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9694/)

src/directory_and_file_reader.c
```diff
@@ -0,0 +1,289 @@
...
+    dir->allocator.deallocate(dir->path, dir->allocator.state);
+  }
+
+  if (dir->dir != NULL) {
+    errno = 0;
+    #ifndef _WIN32
+    int error = closedir(dir->dir);
+    // The closedir() function returns 0 on success
+    if (error) {
+      #else
```
wjwwood: nitpick: this indent is weird too

src/directory_and_file_reader.c
```diff
@@ -0,0 +1,289 @@
...
+    // If the function succeeds, the return value is nonzero.
+    if (!FindClose(dir->dir)) {
+    #endif
```
wjwwood: GitHub is failing to render this, maybe that's the source of your uncrustify issue too?

src/directory_and_file_reader.c
```diff
@@ -0,0 +1,289 @@
...
+    dir->allocator.deallocate(file->name, dir->allocator.state);
+    RCUTILS_SET_ERROR_MSG("unable to allocate memory");
+    return RCUTILS_RET_BAD_ALLOC;
+  }
+
+#ifndef _WIN32
+  if (lstat(
+#else
+  if (_tstat(
+#endif
```
wjwwood: Actually, this stanza is likely the issue... I'd recommend using a `#define` for this function name if that's all that really is different.

dirk-thomas:
> According with the comment of @mikaelarguedas in rcl ros2/rcl#605 (comment) this code probably makes no sense.
> should we close both PRs?

If the API is not used anymore, yes, I think it should just be removed.

PR #223, opened 2020-04-08
add freebsd support

dodsonmg:
Cross-compiling for FreeBSD often results in satisfying `__GNUC__` conditionals if no `__FreeBSD__` conditionals are provided.  In most cases, it should execute the same code as the `__APPLE__` conditionals.

This PR ensures the correct code is executed for FreeBSD.  This is necessary in one case (process.c:50).  In a second case (process.c:74), it is added for completeness, just in case the `__APPLE__` and `__GNUC__` cases diverge in the future.

macros.h:34 also checks if `__APPLE__` is defined, but separate code to support FreeBSD does not appear to be necessary in this case, as the conditional appears to be iOS-specific.


dirk-thomas:
Thanks for the patch.

PR #226, opened 2020-04-08
Fixed condition in rcutils_get_platform_library_name

ahcorde:
Fixed condition in rcutils_get_platform_library_name

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/src/shared_library.c
+++ b/src/shared_library.c
@@ -183,7 +183,7 @@ rcutils_get_platform_library_name(
       library_name_platform, strlen(library_name) + 5, "%s.dll", library_name);
   }
 #endif
-  if (written < 0) {
+  if (written <= 0) {
     RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "failed to format library name: '%s'\n",
       library_name);


ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9990/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5590/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9873/)

src/shared_library.c
```diff
@@ -183,7 +183,7 @@ rcutils_get_platform_library_name(
       library_name_platform, strlen(library_name) + 5, "%s.dll", library_name);
   }
 #endif
-  if (written < 0) {
+  if (written <= 0) {
```
wjwwood, line 186: Wait, why this change?

If you're trying to detect an error case, then just less than zero is correct I think. I believe 0 is a valid, but unlikely, answer here.

Are you trying to include truncation in this error checking?

ahcorde: we need to check also with zero because there is no way to know if there is an error with the length of the strings otherwise the function will return an empty string

wjwwood: That doesn't seem to agree with this test: https://github.com/ros2/rcutils/blob/04456c7540362cf60a01b19d0e505fded4156795/test/test_snprintf.cpp#L32-L34

ahcorde: I understand what you mean, but want also to check the other condition which is checking if it's possible to fill the buffer

 https://github.com/ros2/rcutils/blob/130801fafb71d37a84df5923050d739664bba431/src/shared_library.c#L171

https://github.com/ros2/rcutils/blob/130801fafb71d37a84df5923050d739664bba431/src/shared_library.c#L176

https://github.com/ros2/rcutils/blob/130801fafb71d37a84df5923050d739664bba431/src/shared_library.c#L181

I think this is not affecting the detection of the error from rcutils_snprintf

wjwwood: Ah, I see. Got it.

PR #232, opened 2020-04-10
Don't leak memory on realloc failing.

### LINKED ISSUE #231: char_array is leaking memory
The most recent ASAN builds for Linux are failing [nightly_linux_address_sanitizer#373](https://ci.ros2.org/view/nightly/job/nightly_linux_address_sanitizer/373).

Failing tests:

* [ArrayCharTest.vsprintf_fail](https://github.com/ros2/rcutils/blob/70d828c658410e790a811cd5db365606320c0faa/test/test_char_array.cpp#L100)
* [TestLoggingCustomEnv.test_logging](https://github.com/ros2/rcutils/blob/70d828c658410e790a811cd5db365606320c0faa/test/test_logging_custom_env.cpp#L31)

It looks like it's coming from [`rcutils_char_array_init`](https://github.com/ros2/rcutils/blob/a6306c65c34fecf6f91310ac9830b694a768fed7/src/char_array.c#L52) and [`rcutils_char_array_strncat`](https://github.com/ros2/rcutils/blob/a6306c65c34fecf6f91310ac9830b694a768fed7/src/char_array.c#L214)  in `src/char_array.c` .

## Proposed solution 

1. ?
2. Use `char_array->buffer_length` instead of `strlen(char_array->buffer)`.

clalancette:
When reallocing a memory region, it is possible for the allocation
to fail.  If it does, and you are directly assigning to the
"original" pointer, you'll lose the pointer and hence leak
the memory.  Store the result of the reallocation in a temporary,
do error checking, and then only reassign char_array->buffer
once we are confident it points to real memory.

This should fix #231

### DIFF
--- a/src/char_array.c
+++ b/src/char_array.c
@@ -107,11 +107,12 @@ rcutils_char_array_resize(rcutils_char_array_t * char_array, size_t new_size)
   size_t old_length = char_array->buffer_length;
 
   if (char_array->owns_buffer) {  // we own the buffer, we can do whatever we want
-    char_array->buffer = rcutils_reallocf(char_array->buffer, new_size * sizeof(char), allocator);
+    char * new_buf = rcutils_reallocf(char_array->buffer, new_size * sizeof(char), allocator);
     RCUTILS_CHECK_FOR_NULL_WITH_MSG(
-      char_array->buffer,
+      new_buf,
       "failed to reallocate memory for char array",
       return RCUTILS_RET_BAD_ALLOC);
+    char_array->buffer = new_buf;
   } else {  // we don't realloc memory we don't own. instead, we alloc some new space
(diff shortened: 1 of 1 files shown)


clalancette:
After looking further, this will solve the leaks in test_char_array.c, but not in test_logging_custom_env.cpp.  That will need another fix (which I have ready).  I still think we should proceed with this simple change, as it is a step forward, and then I'll open another PR with the other fixes.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5659/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8189/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9972/)

PR #233, opened 2020-04-10
Make sure to initialize buffers for logging testing.

clalancette:
Also ensure that we cleanup afterwards.  This removes warnings
on asan where we were accessing invalid memory, and also warnings
that this was leaking memory.

Along with #232, solves #231

### DIFF
--- a/test/test_logging_custom_env.cpp
+++ b/test/test_logging_custom_env.cpp
@@ -42,9 +42,7 @@ TEST(CLASSNAME(TestLoggingCustomEnv, RMW_IMPLEMENTATION), test_logging) {
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   rcutils_char_array_t msg_buf, output_buf;
   ret = rcutils_char_array_init(&msg_buf, 1024, &allocator);
-  msg_buf.buffer[0] = '\0';
   ret = rcutils_char_array_init(&output_buf, 1024, &allocator);
-  output_buf.buffer[0] = '\0';
   rcutils_time_point_value_t now = 0;
 
   ret = rcutils_logging_format_message(
--- a/test/test_logging_custom_env.cpp
+++ b/test/test_logging_custom_env.cpp
@@ -42,10 +42,15 @@ TEST(CLASSNAME(TestLoggingCustomEnv, RMW_IMPLEMENTATION), test_logging) {
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   rcutils_char_array_t msg_buf, output_buf;
   ret = rcutils_char_array_init(&msg_buf, 1024, &allocator);
+  msg_buf.buffer[0] = '\0';
   ret = rcutils_char_array_init(&output_buf, 1024, &allocator);
+  output_buf.buffer[0] = '\0';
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5660/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8190/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9973/)

test/test_logging_custom_env.cpp
```diff
@@ -42,10 +42,15 @@ TEST(CLASSNAME(TestLoggingCustomEnv, RMW_IMPLEMENTATION), test_logging) {
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   rcutils_char_array_t msg_buf, output_buf;
   ret = rcutils_char_array_init(&msg_buf, 1024, &allocator);
+  msg_buf.buffer[0] = '\0';
```
dirk-thomas, line 45: Why is this necessary? Should this maybe happen within the `init` function?

clalancette: As far as I can tell, the contract for `rcutils_char_array_init` says nothing about initializing the memory it gives to you.  It just allocates memory for you, and it is up to you to properly fill it in.

Two other ways to fix this would be:
1.  Do a `rcutils_char_array_{strcpy,memcpy,vsprintf}` before using the buffer, which would ensure we have valid data in it and a \0
1.  Use `rcutils_get_zero_initialized_char_array`, which sets up an char_array to use calloc instead of malloc.

I didn't use the second because the contract for it seems weird, and in both (test) places it is used I think it is used incorrectly.  I actually think we should remove it since it a) isn't used in any real code, only tests, and b) doesn't seem to do the right thing.  I can do that in a follow-up.

dirk-thomas: Since there function is called `init` and not `alloc` I think it needs to do the base minimum initialization. Otherwise the bigger can't be safely use (without doing that extra initialization step manually).

clalancette: It wasn't at all clear to me that this was only meant for character data, especially when there are methods like `memcpy` in there.  I guess looking around it is used only for character data, so we could safely do this.  I'll make the change to the init, but I'm also going to do a follow-up where we remove some of these methods that don't make sense for a character array.

piraka9011:
@clalancette Let's get this in if it's ready!

clalancette:
Let me run another CI on it since I changed the code:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10125/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5693/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8228/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10020/)

piraka9011:
@clalancette friendly ping

clalancette:
Sorry, fell off my radar.  Merging.

PR #237, opened 2020-04-14
update rcutils_get_env to always use getenv

### LINKED ISSUE #30: Implement a thread safe version of rcutils_get_env

Suyash458:
fixes #30 
Signed-off-by: Suyash458 <suyash.behera458@gmail.com>

### DIFF
--- a/src/get_env.c
+++ b/src/get_env.c
@@ -22,13 +22,7 @@ extern "C"
 
 #include "rcutils/get_env.h"
 
-#ifdef _WIN32
-# include <errno.h>
-// all environment variables live together in a single memory block
-// which has a limit of 32767 characters
-# define WINDOWS_ENV_BUFFER_SIZE 32767
-#endif  // _WIN32
-
+#pragma warning(disable : 4996)
 
 const char *
 rcutils_get_env(const char * env_name, const char ** env_value)
@@ -39,29 +33,12 @@ rcutils_get_env(const char * env_name, const char ** env_value)
   if (NULL == env_value) {
     return "argument env_value is null";
   }
-  *env_value = NULL;
-#ifdef _WIN32
-  size_t required_size;
-  char * __env_buffer = (char*) malloc(WINDOWS_ENV_BUFFER_SIZE * sizeof(char));
-  errno_t ret = getenv_s(&required_size, __env_buffer, sizeof(__env_buffer), env_name);
-  switch (ret) {
-    case 0:
-      __env_buffer[WINDOWS_ENV_BUFFER_SIZE - 1] = '\0';
-      *env_value = __env_buffer;
-      break;
-    case EINVAL:
-      return "invalid arguments when reading environment variable";
-    case ERANGE:
-      return "insufficient buffer size to read environment variable";
-    default:
-      return "unknown error code reading environment variable";
-  }
-#else
   *env_value = getenv(env_name);
+  // Note: getenv is deprecated; consider using getenv_s instead
+
   if (NULL == *env_value) {
     *env_value = "";
   }
-#endif  // _WIN32
   return NULL;
 }
 
--- a/src/get_env.c
+++ b/src/get_env.c
@@ -27,7 +27,6 @@ extern "C"
 // all environment variables live together in a single memory block
 // which has a limit of 32767 characters
 # define WINDOWS_ENV_BUFFER_SIZE 32767
-static char __env_buffer[WINDOWS_ENV_BUFFER_SIZE];
 #endif  // _WIN32
 
 
@@ -43,6 +42,7 @@ rcutils_get_env(const char * env_name, const char ** env_value)
   *env_value = NULL;
 #ifdef _WIN32
   size_t required_size;
+  char * __env_buffer = (char*) malloc(WINDOWS_ENV_BUFFER_SIZE * sizeof(char));
   errno_t ret = getenv_s(&required_size, __env_buffer, sizeof(__env_buffer), env_name);
   switch (ret) {
     case 0:
--- a/src/get_env.c
+++ b/src/get_env.c
@@ -22,7 +22,9 @@ extern "C"
 
 #include "rcutils/get_env.h"
 
+#ifdef _WIN32
 #pragma warning(disable : 4996)
+#endif
 
 const char *
 rcutils_get_env(const char * env_name, const char ** env_value)
@@ -33,8 +35,9 @@ rcutils_get_env(const char * env_name, const char ** env_value)
   if (NULL == env_value) {
     return "argument env_value is null";
   }
+
+  // TODO(Suyash458): getenv is deprecated on Windows; consider using getenv_s instead
   *env_value = getenv(env_name);
-  // Note: getenv is deprecated; consider using getenv_s instead
 
   if (NULL == *env_value) {
     *env_value = "";


src/get_env.c
```diff
@@ -43,6 +42,7 @@ rcutils_get_env(const char * env_name, const char ** env_value)
   *env_value = NULL;
 #ifdef _WIN32
   size_t required_size;
+  char * __env_buffer = (char*) malloc(WINDOWS_ENV_BUFFER_SIZE * sizeof(char));
```
clalancette, line 45: So I think this is the wrong way to go.  The problem is that on Linux/macOS, this would return a string that the caller does *not* free, and on Windows it would return a string that the caller *does* free.  That's much too difficult to get right, so we should make it consistent.

But let's back up a second.  There are three levels of thread-safety that are currently being violated by this current version of `rcutils_get_env`.  The first level is that, on Windows, we are returning a pointer to a global array, so if two threads try to *get* an environment variable at the same time, they'll stomp all over each other.  The second level is that, on all platforms, if a thread gets an environment variable, and then somewhere down the line something else calls `setenv`, that can cause the pointer to be invalidated.  The third level of thread safety being violated is that if something calls `{un}setenv` at the same time that something else is calling `getenv`, a race and/or crash can occur.

With that being said, I propose that we split this into three different patches:
1.  We remove the global Windows buffer completely, and switch from `getenv_s` to `getenv` on Windows.  That will make the behavior consistent between Linux, macOS, and Windows, and won't cause us to change this function's signature.  This doesn't cause us to be any less thread-safe then we are on Linux and macOS now, and improves thread safety for Windows.  (I'll suggest we just modify this simple PR to accomplish this, but we can open a new one as well).
2.  Later do a follow-up where we getenv and copy the string *for all platforms*.  This will solve the second thread-safety issue, since a later `setenv` won't affect outstanding pointers.  That will cause us to change the signature and the semantic of `rcutils_get_env` so we can have an allocator, and so downstream users free the buffer we give them.
3.  Even later, do an audit of any places that call `rcutils_get_env` and `setenv`, and add locking as appropriate.

Doing it this way addresses our most pressing thread-safety issue (number 1), while leaving the less likely ones to follow-up later.

Does this make sense?  @hidmic @ivanpauno @wjwwood thoughts?

wjwwood: That sounds right to me.

Suyash458: @clalancette just wanted to point out that [`getenv` is deprecated](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=vs-2019) so, should we just go ahead with 2 and 3 instead?

clalancette: > @clalancette just wanted to point out that [`getenv` is deprecated](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=vs-2019) so, should we just go ahead with 2 and 3 instead?

Personally, I'd still go with using the deprecated `getenv` for Windows (with appropriate changes to silence the compiler warnings).  If you want to go straight for 2, go for it, but I warn you that it will probably be complicated and span several repositories.

ivanpauno: Sounds good to me, and I agree with @clalancette that option 1 is the first thing to do.
The deprecated warning can be silenced using `pragma warning` appropriately.

src/get_env.c
```diff
@@ -40,28 +33,12 @@ rcutils_get_env(const char * env_name, const char ** env_value)
...
-    case EINVAL:
-      return "invalid arguments when reading environment variable";
-    case ERANGE:
-      return "insufficient buffer size to read environment variable";
-    default:
-      return "unknown error code reading environment variable";
-  }
-#else
   *env_value = getenv(env_name);
+  // Note: getenv is deprecated; consider using getenv_s instead
```
clalancette, line 37: This note should be above the line it talks about, and should be something more like:

```suggestion
  // TODO(Suyash458): getenv is deprecated on Windows; consider using getenv_s instead
```

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10211/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5759/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8300/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10100/)

src/get_env.c
```diff
@@ -22,14 +22,7 @@ extern "C"
...
 
-#ifdef _WIN32
-# include <errno.h>
-// all environment variables live together in a single memory block
-// which has a limit of 32767 characters
-# define WINDOWS_ENV_BUFFER_SIZE 32767
-static char __env_buffer[WINDOWS_ENV_BUFFER_SIZE];
-#endif  // _WIN32
-
+#pragma warning(disable : 4996)
```
clalancette, line 25: Ah, CI is showing unknown pragmas from GCC and clang.  You'll need to wrap this in an `#ifdef _WIN32`.

src/get_env.c
Suyash458: Sure, will update this and create a new issue

Suyash458: @clalancette do we want to close [this](https://github.com/ros2/rcutils/issues/30) and create a new issue or just update it with these discussions?

clalancette: Let's close #30 (which will happen just by merging this PR) and open a new one.

src/get_env.c
Suyash458: I've updated this bit and the comment

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10377/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5892/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8457/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10274/)

dirk-thomas:
This looks like a step back to me. There is a reason why this function is deprecated and why we moved away from it in the past.

clalancette:
> This looks like a step back to me. There is a reason why this function is deprecated and why we moved away from it in the past.

I disagree.  https://github.com/ros2/rcutils/pull/237#discussion_r408151034 lays it out in a lot of detail; using `getenv_s` as we are currently using it is worse than using `getenv`.  Change 2 above will be even better, but this change removes the worst offender of thread unsafety we have.

ivanpauno:
> This looks like a step back to me. There is a reason why this function is deprecated and why we moved away from it in the past.

`getenv` on Windows isn't actually more "unsafe" than on Linux.
Windows marks `getenv` as deprecated not because is going to be removed, but because they think that `getenv_s` signature allows to catch more errors (see https://docs.microsoft.com/en-us/cpp/c-runtime-library/security-features-in-the-crt?view=vs-2019).
The same logic also applies on Linux, but all `*_s` functions weren't implemented there (which is an optional C11 extension AFAIK).

IMO, this solves the problem of calling `rcutils_getenv` concurrently (it doesn't solve the problem of also modifying the environment concurrently with `putenv` or `unsetenv`, which is a different one).

We can also add another version of `rcutils_getenv` mimicking `getenv_s` (which on Linux/osX will just use `getenv`), and deprecate the old one if we don't want to ignore windows CRT warnings (which IMO, they can safely be ignored).

clalancette:
Bah, my github gist is now out-of-date.  Updating it and kicking off another round of CI.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5903/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8472/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10286/)

src/get_env.c
```diff
@@ -40,28 +35,13 @@ rcutils_get_env(const char * env_name, const char ** env_value)
...
-    case EINVAL:
-      return "invalid arguments when reading environment variable";
-    case ERANGE:
-      return "insufficient buffer size to read environment variable";
-    default:
-      return "unknown error code reading environment variable";
-  }
-#else
+
+  // TODO(Suyash458): getenv is deprecated on Windows; consider using getenv_s instead
```
hidmic, line 39: @Suyash458 aren't we consciously ignoring that deprecation? Why keeping a TODO?

clalancette: I asked for that.  It's mostly so that we can make the even better (but bigger) change in step 2 in https://github.com/ros2/rcutils/pull/237#discussion_r408151034 .  At that point, we most likely will switch back to `getenv_s`.

clalancette:
I'm going to go ahead and merge this as we have green CI.  I'll also open a follow-up issue so that we can do more thread-safetyness of this function in the future.

PR #255, opened 2020-05-22
QD Security Vulnerability Declaration: REP 2006

clalancette:
This PR adds a link to REP-2006 (the Security Vulnerability Declaration) to the Quality Declaration for this repository.

Connects to ros2/ros2#924.

### DIFF
--- a/QUALITY_DECLARATION.md
+++ b/QUALITY_DECLARATION.md
@@ -154,7 +154,7 @@ Currently nightly build status can be seen here:
 
 ## Vulnerability Disclosure Policy [7.i]
 
-This package does not yet have a Vulnerability Disclosure Policy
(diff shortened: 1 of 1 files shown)


QUALITY_DECLARATION.md
```diff
@@ -205,5 +205,3 @@ Comparing this table with the [Quality Level Comparison Chart of REP2004](https:
 Missing to quality level 3:
```
chapulina, line 205: The table above also needs updating.

Alternatively, it could be removed so the declaration matches most of the other ones.

clalancette: Good point.  I removed some pieces of this so it looks more like the other packages.

PR #256, opened 2020-06-01
Update version stability section of quality declaration for 1.0.

nuclearsandwich:
1.0 for this package was bumped in https://github.com/ros2/rcutils/commit/59773572d1e467c30861e11274b2d0b1056e2f55 so this updates the quality declaration.

It looks to me like this might make the package level 2 or level 3. I'd appreciate a comment on that front from someone whose got more familiarity with what's part of each quality level.

### DIFF
--- a/QUALITY_DECLARATION.md
+++ b/QUALITY_DECLARATION.md
@@ -165,7 +165,7 @@ The chart below compares the requirements in the REP-2004 with the current state
 |--|--|--|
 |1| **Version policy** |---|
 |1.i|Version Policy available | ✓ |
-|1.ii|Stable version |☓|
+|1.ii|Stable version |✓||
 |1.iii|Declared public API|✓|
 |1.iv|API stability policy|✓|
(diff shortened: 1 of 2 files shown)


ahcorde:
coverage is not yet ready. It's QD 3.

nuclearsandwich:
Bumped to QL3.

nuclearsandwich:
> README.md needs to be updated too

I think it's really unfortunate we duplicate this info. What do we think of just linking to the quality declaration? Yeah it's "one more click" but it's one less way to be wrong.

Blast545:
We have discussed we should have some sort of badges similar to the ones shown with CI results, but for the time being is yet too be implemented.

In fact I just realized with this PR you should update too the table at the end of the QD with a check for version stability.

nuclearsandwich:
> In fact I just realized with this PR you should update too the table at the end of the QD with a check for version stability.

:facepalm:

PR #257, opened 2020-06-01
Remove failure check of RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED

brawner:
While looking for places to increase testing coverage, I noticed that logic in logging.c required the `rcutils_get_env` call for RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED to at least succeed. As currently implemented, it will always succeed and return NULL thus the else statement is dead code. But even if it could result in an error, the else clause doesn't seem useful since the environment variable use has been deprecated.

Importantly for foxy considerations, this will not change the API of the function, since the environment variable was unused and the rcutils_get_env call will always succeed.

`colcon build --packages-up-to rcutils && colcon test --packages-select rcutils`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10962/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6302/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8924/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10865/)

### DIFF
--- a/src/format_string.c
+++ b/src/format_string.c
@@ -39,7 +39,7 @@ rcutils_format_string_limit(
     return NULL;
   }
   RCUTILS_CHECK_ALLOCATOR(&allocator, return NULL);
-  // extract the variadic arguments twice, once for length calculatio and once for formatting.
+  // extract the variadic arguments twice, once for length calculation and once for formatting.
   va_list args1;
   va_start(args1, format_string);
   va_list args2;
--- a/src/logging.c
+++ b/src/logging.c
@@ -147,18 +147,12 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
 
     const char * line_buffered = NULL;
     const char * ret_str = rcutils_get_env("RCUTILS_CONSOLE_STDOUT_LINE_BUFFERED", &line_buffered);
-    if (NULL == ret_str) {
-      if (strcmp(line_buffered, "") != 0) {
-        fprintf(
-          stderr,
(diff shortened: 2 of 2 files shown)


brawner:
In the case where `rcutils_get_env` returns an error with the inputs that logging_initialize passes it, then it would be a programming error and not bad input or system error. What do you think of using an assert to check ret is null?

i.e.
```
assert(ret_str == NULL);
if (strcmp(line_buffered, "") != 0) {
      fprintf(...);
}
```

clalancette:
> In the case where `rcutils_get_env` returns an error with the inputs that logging_initialize passes it, then it would be a programming error and not bad input or system error. What do you think of using an assert to check ret is null?

Oh, I see.  Yeah, for right now, you are correct.  But we are looking at changing the signature of `rcutils_get_env` so it is more threadsafe: https://github.com/ros2/rcutils/issues/242 .  When we do that, `rcutils_get_env` will be able to return errors related to system problems (in particular, it will allocate memory, which may fail).

While I don't usually like to hold code up on something that may happen in the future, in this case it will be pretty easy to forget that we baked this knowledge in here.  So in the name of defensive programming, I'd still suggest that we keep the distinction between something going wrong in the system and the thing being empty.

brawner:
I just wanted to float this PR as an idea. I realize that a lot of repetitive checks are happening in the code base and I don't have a problem with that. This just seemed like a situation where if the API of `rcutils_get_env` changed and started throwing errors, then this part of the code base shouldn't really care. Either way, we get to 95% coverage without removing these lines in #258 so I'm not too concerned about this chunk.

PR #258, opened 2020-06-02
Increase testing coverage of rcutils to 95%

brawner:
I've added a bit more test cases to the unit tests in rcutils to boost the coverage to 95% (just barely).

The big question about this PR is the desired location of `time_bomb_allocator`. @hidmic had suggested putting it into a centralized location so it can be used in other packages. I'm thinking I might want to do that for the allocator_testing_utils.h as well. But I don't think rcutils/include is the desired end location.

The remaining 5% of untested logic pretty much consists of failure checks that won't fail, generally because bad parameters have already been sanitized by the time the function under test calls a function that is checked for failure. In a few cases it might be because the system call cannot be made to fail reliably or even at all in a simple unit test.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10965/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6305/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8927/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10875/)

### DIFF
--- a/test/test_filesystem.cpp
+++ b/test/test_filesystem.cpp
@@ -44,6 +44,9 @@ public:
 TEST_F(TestFilesystemFixture, get_cwd_nullptr) {
   EXPECT_FALSE(rcutils_get_cwd(NULL, sizeof(this->cwd)));
   EXPECT_FALSE(rcutils_get_cwd(this->cwd, 0));
+
+  // ERANGE, including a null terminating character, cwd should always be longer than 1 char
+  EXPECT_FALSE(rcutils_get_cwd(this->cwd, 1));
 }
 
 TEST_F(TestFilesystemFixture, join_path) {
@@ -234,6 +237,16 @@ TEST_F(TestFilesystemFixture, is_writable) {
 TEST_F(TestFilesystemFixture, is_readable_and_writable) {
   {
     char * path = rcutils_join_path(this->test_path, "dummy_folder", g_allocator);
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      g_allocator.deallocate(path, g_allocator.state);
+    });
+    ASSERT_FALSE(nullptr == path);
+    EXPECT_TRUE(rcutils_is_readable_and_writable(path));
+  }
+  {
+    char * path =
+      rcutils_join_path(this->test_path, "dummy_readable_file.txt", g_allocator);
     OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
     {
       g_allocator.deallocate(path, g_allocator.state);
--- a/test/test_get_env.cpp
+++ b/test/test_get_env.cpp
@@ -50,15 +50,22 @@ TEST(TestGetEnv, test_get_env) {
 
 TEST(TestGetEnv, test_get_home) {
   EXPECT_STRNE(NULL, rcutils_get_home_dir());
-
   const char * home = NULL;
-  const char * ret = rcutils_get_env("HOME", &home);
-  EXPECT_EQ(NULL, ret);
 
 #ifdef _WIN32
+  // Assert pre-condition that USERPROFILE is defined
+  const char * ret = rcutils_get_env("USERPROFILE", &home);
+  ASSERT_EQ(NULL, ret);
+
+  // Check USERPROFILE is not defined
   EXPECT_TRUE(rcutils_set_env("USERPROFILE", NULL));
   EXPECT_EQ(NULL, rcutils_get_home_dir());
 #else
+  // Assert pre-condition that HOME is defined
+  const char * ret = rcutils_get_env("HOME", &home);
+  ASSERT_EQ(NULL, ret);
+
+  // Check HOME is not defined
   EXPECT_TRUE(rcutils_set_env("HOME", NULL));
   EXPECT_EQ(NULL, rcutils_get_home_dir());
 #endif
--- a/test/test_get_env.cpp
+++ b/test/test_get_env.cpp
@@ -16,6 +16,7 @@
 
 #include <string>
 
+#include "rcutils/env.h"
 #include "rcutils/get_env.h"
 
 /* Tests rcutils_get_env.
@@ -49,4 +50,16 @@ TEST(TestGetEnv, test_get_env) {
 
 TEST(TestGetEnv, test_get_home) {
   EXPECT_STRNE(NULL, rcutils_get_home_dir());
+
+  const char * home = NULL;
+  const char * ret = rcutils_get_env("HOME", &home);
+  EXPECT_EQ(NULL, ret);
+
+#ifdef _WIN32
+  EXPECT_TRUE(rcutils_set_env("USERPROFILE", NULL));
+  EXPECT_EQ(NULL, rcutils_get_home_dir());
+#else
+  EXPECT_TRUE(rcutils_set_env("HOME", NULL));
+  EXPECT_EQ(NULL, rcutils_get_home_dir());
+#endif
 }
--- a/test/test_logging_bad_env.cpp
+++ b/test/test_logging_bad_env.cpp
@@ -1,0 +1,22 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "rcutils/logging.h"
+
+TEST(TestBadCustomEnv, test_initialize) {
+  EXPECT_FALSE(g_rcutils_logging_initialized);
+  ASSERT_EQ(RCUTILS_RET_INVALID_ARGUMENT, rcutils_logging_initialize());
+}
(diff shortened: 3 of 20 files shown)


test/test_get_env.cpp
```diff
@@ -49,4 +50,16 @@ TEST(TestGetEnv, test_get_env) {
 
 TEST(TestGetEnv, test_get_home) {
   EXPECT_STRNE(NULL, rcutils_get_home_dir());
+
+  const char * home = NULL;
+  const char * ret = rcutils_get_env("HOME", &home);
+  EXPECT_EQ(NULL, ret);
```
hidmic, line 56: @brawner mind to explain why can it be assumed that `HOME` is not in the environment?

test/test_logging_bad_env.cpp
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "rcutils/logging.h"
+
+TEST(TestBadCustomEnv, test_initialize) {
+  EXPECT_FALSE(g_rcutils_logging_initialized);
+  ASSERT_EQ(RCUTILS_RET_INVALID_ARGUMENT, rcutils_logging_initialize());
```
hidmic, line 21: @brawner uh, `RCUTILS_RET_INVALID_ARGUMENT` for a function that takes no arguments?

test/test_logging_console_output_handler.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+TEST(TestLoggingConsoleOutputHandler, typical_inputs) {
+  ASSERT_EQ(RCUTILS_RET_OK, rcutils_logging_initialize());
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    EXPECT_EQ(RCUTILS_RET_OK, rcutils_logging_shutdown());
+  });
+
+  rcutils_log_location_t log_location = {
```
hidmic, line 41: @brawner what follows seems akin to a smoke test to me. Shall we leave a comment stating so? Same for the other test case below.

test/test_repl_str.cpp
```diff
@@ -62,4 +62,31 @@ TEST(test_repl_str, nominal) {
...
+    EXPECT_EQ(NULL, out);
+  }
+
+  // Force cache_sz_inc to exceed cache_sz_inc_max
+  {
+    std::stringstream ss;
+    for (size_t i = 0; i < 1048576; ++i) {
+      ss << "f";
+    }
+    char * out = rcutils_repl_str(ss.str().c_str(), "f", "longer replacementment", &allocator);
```
hidmic, line 88: @brawner typo? 
```suggestion
    char * out = rcutils_repl_str(ss.str().c_str(), "f", "longer replacement", &allocator);
```

test/test_strerror.cpp
```diff
@@ -43,4 +43,12 @@ TEST(test_strerror, get_error) {
 
   rcutils_strerror(error_string, sizeof(error_string));
   ASSERT_STREQ(error_string, "No such file or directory");
+
+#if (!defined(_WIN32)) && (!( \
+    defined(_GNU_SOURCE) && (!defined(ANDROID) || __ANDROID_API__ >= 23)))
+  // Hopefully this does not become a valid errno.
+  errno = 12345;
+  rcutils_strerror(error_string, sizeof(error_string));
+  ASSERT_STREQ(error_string, "Failed to get error");
```
hidmic, line 52: @brawner a message to hint our future selves in case of failure would be nice.

test/test_get_env.cpp
brawner: The return of `rcutils_get_env` is an error string. If it's null then the call succeeded. The actual variable value is assigned to `home`.

test/test_logging_bad_env.cpp
brawner: They are set in the CMakeLists.txt file. But `rcutils_logging_initialize()` checks environment variables and will return an invalid argument in that case.

test/test_logging_console_output_handler.cpp
brawner: Added a more informative comment at the top of the test function.

test/test_repl_str.cpp
brawner: Fixed.

test/test_strerror.cpp
brawner: Added a message.

brawner:
Separate PR to fix Windows bug:
https://github.com/ros2/rcutils/pull/259

test/test_get_env.cpp
hidmic: Oops, nevermind.

test/test_logging_bad_env.cpp
hidmic: I see. It's out of scope for this PR, but I find the error code selection weird. Environment variables are not function arguments.

test/test_logging_console_output_handler.cpp
hidmic: Cool. Mocks would come in handy here...

brawner:
After merging #259, I've rebased this PR. Also for API considerations I moved the time_bomb allocator into the test directory. We can leave that discussion for post-foxy discussions.

brawner:
Retesting:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11009/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6344/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8968/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10935/)

brawner:
Also without #257 this shows up as only 94%. At 1666/1763, it just needs one more line of coverage or one less line of uncovered code 🤷‍♂️

test/test_filesystem.cpp
```diff
@@ -241,6 +244,16 @@ TEST_F(TestFilesystemFixture, is_readable_and_writable) {
...
     EXPECT_TRUE(rcutils_is_readable_and_writable(path));
   }
+  {
+    char * path =
+      rcutils_join_path(this->test_path, "dummy_readable_file.txt", g_allocator);
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      g_allocator.deallocate(path, g_allocator.state);
+    });
+    ASSERT_FALSE(nullptr == path);
```
cottsay, line 254:
```suggestion
    ASSERT_NE(nullptr, path);
```

test/test_get_env.cpp
cottsay: What is the point of this test though? Won't it fail if `HOME` isn't set on Windows, for example?

test/test_filesystem.cpp
brawner: After private discussion, leaving this as is to match the rest the file.

test/test_get_env.cpp
brawner: The original intent of this check was to verify that "HOME" was defined, so that we can be sure that the expected post condition below was a result of setting each environment variable to NULL. However, I forgot to update this when I added in the #ifdef for windows.

I have since updated this to check the appropriate HOME/USERPROFILE variables in the ifdef below.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11011/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6346/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8970/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10938/)

test/test_logging_console_output_handler.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+  va_list args;
+  va_start(args, format);
+  rcutils_logging_console_output_handler(location, severity, name, timestamp, format, &args);
+  va_end(args);
+}
+
+// There are no outputs of the handler function, and the only result are fprintf() calls.
+// This is just a smoke test to check that the code can handle simple inputs cleanly.
+TEST(TestLoggingConsoleOutputHandler, typical_inputs) {
```
hidmic, line 36: @brawner meta: thinking again about this test, you _could_ mock `stdout` and/or `stderr` for the duration of the test. Won't block on it, but it is doable.

test/test_logging_console_output_handler.cpp
```diff
@@ -0,0 +1,97 @@
...
+    &log_location, RCUTILS_LOG_SEVERITY_WARN, log_name, timestamp, format, "part1", "part2");
+  call_handler(
+    &log_location, RCUTILS_LOG_SEVERITY_ERROR, log_name, timestamp, format, "part1", "part2");
+  call_handler(
+    &log_location, RCUTILS_LOG_SEVERITY_FATAL, log_name, timestamp, format, "part1", "part2");
+}
+
+// There are no outputs of the handler function, and the only result are fprintf() calls.
+// This is just a smoke test to check that the code can handle bad inputs cleanly.
+TEST(TestLoggingConsoleOutputHandler, bad_inputs) {
```
hidmic, line 65: @brawner same as above.

brawner:
Testing once more before merging:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11049/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6373/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9000/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10972/)

test/test_logging_console_output_handler.cpp
brawner: I tried this out and couldn't get it to work. Assigning a new file pointer to stdout/stderr left me with blank files. Alternatively, using `freopen()` worked but it wouldn't let me reset stdout/stderr back in an easily portable way.

I like the idea of checking expected input on stdout/stderr, but I'll need a better solution, which may have to come later.

hidmic: Hmm, interesting. Oh well, it can be done in a follow-up PR.

PR #261, opened 2020-06-20
Add parenthesis around the argument in time conversion macros defined in time.h

meyerj:
... to protect against unexpected results of calls with sum arguments, or with other operators that have a lower precedence than the multiplication or division operators.

Related to https://github.com/ros2/rclcpp/pull/1188:

> There is another problem with the definition of time-related macros defined in [rcutils/time.h](https://github.com/ros2/rcutils/blob/8c1d190fc5e6425f83c3e8716a5ee861bbf82cf1/include/rcutils/time.h) which are also indirectly used for the `Duration` and `Time` implementations in `rclcpp`: They don't have parentheses around the macro argument:
> 
> ```cpp
> /// Convenience macro to convert seconds to nanoseconds.
> #define RCUTILS_S_TO_NS(seconds) (seconds * (1000LL * 1000LL * 1000LL))
> /// Convenience macro to convert milliseconds to nanoseconds.
> #define RCUTILS_MS_TO_NS(milliseconds) (milliseconds * (1000LL * 1000LL))
> /// Convenience macro to convert microseconds to nanoseconds.
> #define RCUTILS_US_TO_NS(microseconds) (microseconds * 1000LL)
> 
> /// Convenience macro to convert nanoseconds to seconds.
> #define RCUTILS_NS_TO_S(nanoseconds) (nanoseconds / (1000LL * 1000LL * 1000LL))
> /// Convenience macro to convert nanoseconds to milliseconds.
> #define RCUTILS_NS_TO_MS(nanoseconds) (nanoseconds / (1000LL * 1000LL))
> /// Convenience macro to convert nanoseconds to microseconds.
> #define RCUTILS_NS_TO_US(nanoseconds) (nanoseconds / 1000LL)
> ```
> 
> So `RCUTILS_S_TO_NS(1 + 1)` is not `2000000000LL` as one might expect, but `1000000001LL`, because of operator precedence rules. This case is not triggered in `rclcpp`. I will open a separate pull request in https://github.com/ros2/rcutils for that.

### DIFF
--- a/test/test_time.cpp
+++ b/test/test_time.cpp
@@ -52,40 +52,50 @@ public:
 
 // Tests the rcutils time unit conversion macros.
 TEST_F(TestTimeFixture, test_rcutils_time_conversion_macros) {
-  int64_t ns = 0;
-
   // Note: 9007199254740993 or higher cannot be represented anymore by intermediate type double
   // without a loss of precision.
 
   // seconds to nanoseconds
-  EXPECT_EQ(ns = RCUTILS_S_TO_NS(1), 1000000000ll);  // int
-  EXPECT_EQ(ns = RCUTILS_S_TO_NS(0.2), 200000000ll);  // double
-  EXPECT_EQ(ns = RCUTILS_S_TO_NS(1 + 1), 2000000000ll);  // sum of ints
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(1)), 1000000000ll);  // int
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(0.2)), 200000000ll);  // double
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(1 + 1)), 2000000000ll);  // sum of ints
   EXPECT_EQ(
-    ns = RCUTILS_S_TO_NS(9007199.254740992),
+    static_cast<int64_t>(RCUTILS_S_TO_NS(9007199.254740992)),
     9007199254740992ll);  // maximum precision double (53 bits)
-  EXPECT_NE(ns = RCUTILS_S_TO_NS(9007199.254740993), 9007199254740993ll);
-  EXPECT_EQ(ns = RCUTILS_S_TO_NS(9007199.254740993), 9007199254740992ll);  // value is truncated!
+  EXPECT_NE(
+    static_cast<int64_t>(RCUTILS_S_TO_NS(9007199.254740993)),
+    9007199254740993ll);
+  EXPECT_EQ(
+    static_cast<int64_t>(RCUTILS_S_TO_NS(9007199.254740993)),
+    9007199254740992ll);  // value is truncated!
 
   // milliseconds to nanoseconds
-  EXPECT_EQ(ns = RCUTILS_MS_TO_NS(1), 1000000ll);  // int
-  EXPECT_EQ(ns = RCUTILS_MS_TO_NS(0.2), 200000ll);  // double
-  EXPECT_EQ(ns = RCUTILS_MS_TO_NS(1 + 1), 2000000ll);  // sum of ints
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_MS_TO_NS(1)), 1000000ll);  // int
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_MS_TO_NS(0.2)), 200000ll);  // double
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_MS_TO_NS(1 + 1)), 2000000ll);  // sum of ints
   EXPECT_EQ(
-    ns = RCUTILS_MS_TO_NS(9007199254.740992),
+    static_cast<int64_t>(RCUTILS_MS_TO_NS(9007199254.740992)),
     9007199254740992ll);  // maximum precision double (53 bits)
-  EXPECT_NE(ns = RCUTILS_MS_TO_NS(9007199254.740993), 9007199254740993ll);
-  EXPECT_EQ(ns = RCUTILS_MS_TO_NS(9007199254.740993), 9007199254740994ll);  // value is truncated!
+  EXPECT_NE(
+    static_cast<int64_t>(RCUTILS_MS_TO_NS(9007199254.740993)),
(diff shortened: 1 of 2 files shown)


meyerj:
> Can you add a test for the problematic cases here?

Done in https://github.com/ros2/rcutils/pull/261/commits/ab897e4723cdacf4c258a8bf4531959effbb16a2.

The new unit test `test_rcutils_time_conversion_macros` covers the cases where the macro argument is a sum, and also highlights some other potentially problematic cases because of the limited precision of type `double` compared to `int64_t`.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11289/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6536/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11192/)

test/test_time.cpp
```diff
@@ -50,6 +50,74 @@ class TestTimeFixture : public ::testing::Test
...
+// Tests the rcutils time unit conversion macros.
+TEST_F(TestTimeFixture, test_rcutils_time_conversion_macros) {
+  int64_t ns = 0;
+
+  // Note: 9007199254740993 or higher cannot be represented anymore by intermediate type double
+  // without a loss of precision.
+
+  // seconds to nanoseconds
+  EXPECT_EQ(ns = RCUTILS_S_TO_NS(1), 1000000000ll);  // int
+  EXPECT_EQ(ns = RCUTILS_S_TO_NS(0.2), 200000000ll);  // double
```
clalancette, line 62: CI for windows is showing a lot of new compiler warnings (on this line and others).  This is because MSVC is very pedantic about types, so assigning a double to an int64_t is a possible loss of precision.  Can you fix these types up (you can see more in the test output at https://ci.ros2.org/job/ci_windows/11192/msbuild/new/)?

Also, out of curiousity, why are you doing the assignment to `ns` at all here?

meyerj: Exactly for that, to make the implicit conversion to `int64_t` explicit. Otherwise the result of `RCUTILS_S_TO_NS(0.2)` would be of type double and `EXPECT_EQ()` would probably also convert the right argument to double for the comparison.

So the "best" and most explicit variant would be
```suggestion
  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(0.2)), 200000000ll);  // double
```

test/test_time.cpp
```diff
@@ -50,6 +50,84 @@ class TestTimeFixture : public ::testing::Test
...
 };
 
+// Tests the rcutils time unit conversion macros.
+TEST_F(TestTimeFixture, test_rcutils_time_conversion_macros) {
+  // Note: 9007199254740993 or higher cannot be represented anymore by intermediate type double
+  // without a loss of precision.
+
+  // seconds to nanoseconds
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(1)), 1000000000ll);  // int
+  EXPECT_EQ(static_cast<int64_t>(RCUTILS_S_TO_NS(0.2)), 200000000ll);  // double
```
clalancette, line 60: For this (and all of the other double checks), it seems to me that instead of casting the result of the macro to int64_t, we should actually change the expected value to be a double.  After all, that would more closely test that the output of the macro returns the type we expect, no?

meyerj: That is clearly beyond scope of this pull request, I think. Why would you expect the macros to return seconds as `double`? They don't. They return the same type as their argument, and eventually promote the integer to the width of a `long long` because of the `ll` suffix of the constants.

But `RCUTILS_MS_TO_S(1)` for example returns `0`, even if you would cast the result to `double` afterwards. Calling the macros with a `uint64_t` triggers a sign-conversion warning.


If you expect certain fixed input and output types, then they should probably be replaced by inline functions or the macros need some additional explicit casts. But that would be a potentially breaking change.

meyerj: I am sorry, I associated your comment with the wrong line of code. Clearly, if the argument is of type `double`, the output can be expected of type `double`, too, and the cast to `int64_t` can be omitted.

clalancette: > I am sorry, I associated your comment with the wrong line of code. Clearly, if the argument is of type `double`, the output can be expected of type `double`, too, and the cast to `int64_t` can be omitted.

Yeah, that's what I was getting at.  Sorry if I wasn't clear.  Anyway, could you please make that change?  Thanks.

meyerj: My thought was that the macro might be used to assign to an `int64_t`, like
```
double seconds = 42.5;
int64_t nanoseconds = RCUTILS_S_TO_NS(seconds);
```

But I could not find any usage of these macros with a floating point argument and this code snippet triggers a `-Wconversion` compiler warning, so probably they were just never meant for that purpose.

clalancette:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11315/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6555/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9280/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11214/)

meyerj:
> New CI:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/11315/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6555/)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/9280/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/11214/)

Are the test failures on Windows related to this patch? I do not see how.

I am afraid that I cannot follow-up on this anymore. Creating and amending the unit tests for such a small and trivial patch as adding parentheses for macro arguments already cost too much time, and the multi-platform support makes it hard to test locally before applying changes (or I am not aware of the procedures yet).

clalancette:
> Are the test failures on Windows related to this patch? I do not see how.

Yeah, I don't see how either.  We have some known issues with services in ROS 2, so I'm going to chalk it up to that.  Thanks for iterating, I'm going to go ahead and merge.

PR #263, opened 2020-06-25
Add compiler option -Wconversion and add explicit casts for conversions that may alter the value or change the sign

meyerj:
Follow-up on https://github.com/ros2/rcutils/pull/261#discussion_r445743549.

I think most implicit conversions have been okay. But not checking the signed return value of `rcutils_vsnprintf()` in `format_string.c` could be considered a bug.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ find_package(ament_cmake_ros REQUIRED)
 ament_python_install_package(${PROJECT_NAME})
 
 if(NOT WIN32)
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic -Wconversion)
 endif()
 
 if(WIN32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,11 @@ find_package(ament_cmake_ros REQUIRED)
 ament_python_install_package(${PROJECT_NAME})
 
 if(NOT WIN32)
-  add_compile_options(-Wall -Wextra -Wpedantic -Wconversion)
+  # About -Wno-sign-conversion: With Clang, -Wconversion implies -Wsign-conversion. There are a number of
+  # implicit sign conversions in gtest.cc, see https://ci.ros2.org/job/ci_osx/9381/clang/.
+  # Hence disabling -Wsign-conversion for now until all those have eventually been fixed.
+  # (from https://github.com/ros2/rcutils/pull/263#issuecomment-663252537)
+  add_compile_options(-Wall -Wextra -Wconversion -Wno-sign-conversion -Wpedantic)
 endif()
 
 if(WIN32)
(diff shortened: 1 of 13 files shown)


jacobperron:
@meyerj Can you fix this additional warning on macOS: https://ci.ros2.org/job/ci_osx/9381/clang/fileName.-1409726512/ ?

For the the warnings coming from gtest, I guess we should do something similar to `rclcpp` and add `-Wno-sign-conversion` for Clang:

https://github.com/ros2/rclcpp/blob/7e68d3549cb4162630446a1cadf62a4b2f64cc82/rclcpp/CMakeLists.txt#L28-L32

meyerj:
> @meyerj Can you fix this additional warning on macOS: https://ci.ros2.org/job/ci_osx/9381/clang/fileName.-1409726512/ ?

Done in https://github.com/ros2/rcutils/pull/263/commits/7db43f8ce14aed9b5d72c6f60f18e11214bc10c5. By adding `-Wno-sign-conversion` in `CMakeLists.txt` the other fixes would not have been required, but I verified that all warnings other than from gtest/gmock sources and headers have been fixed even with `-Wsign-conversion` enabled (on Linux with gcc 9.3).

jacobperron:
> By adding -Wno-sign-conversion in CMakeLists.txt the other fixes would not have been required

What if we just add `-Wno-sign-conversion` for Clang? E.g.

```
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wno-sign-conversion)
endif()
```

jacobperron:
err, sorry. I think you meant that the additional fixes in 7db43f8 aren't necessary with the warning suppression?

meyerj:
> err, sorry. I think you meant that the additional fixes in [7db43f8](https://github.com/ros2/rcutils/commit/7db43f8ce14aed9b5d72c6f60f18e11214bc10c5) aren't necessary with the warning suppression?

They are not necessary to fix the build on macOS, but from https://github.com/ros2/rcutils/pull/263#issuecomment-663252537 I understood that you asked me to fix them anyway, such that the only _remaining_ sign conversion warnings would be produced by gtest and gmock.

I always considered `-Wno-sign-conversion` as a temporary workaround. Or do you think sign conversion are too pedantic and should remain ignored forever, such that the other changes are not required? In that case also the last sentence of the comments in [rclcpp/CMakeLists.txt](https://github.com/ros2/rclcpp/blob/7e68d3549cb4162630446a1cadf62a4b2f64cc82/rclcpp/CMakeLists.txt#L28-L32) and now added to [rcutils/CMakeLists.txt](https://github.com/ros2/rcutils/pull/263/files#diff-af3b638bc2a3e6c650974192a53c7291R22) is a bit misleading.

jacobperron:
> They are not necessary to fix the build on macOS, but from #263 (comment) I understood that you asked me to fix them anyway, such that the only remaining sign conversion warnings would be produced by gtest and gmock.

Understood, this is what I asked :+1: 

> I always considered -Wno-sign-conversion as a temporary workaround. 

I agree, it is just a workaround to deal with the gtest/gmock errors.

Sorry for the confusion. Thanks for iterating!

macOS: [Build Status](https://ci.ros2.org/job/ci_osx/9511/)

PR #264, opened 2020-07-01
Fault injection macros and functionality (plus example)

brawner:
This introduces a set of fault injection macros that enable selective functions to return failing error codes to help test upstream code. 

The macro `RCUTILS_MAYBE_RETURN_ERROR` is a compile flag enabled macro that selectively returns the error value that is passed to it. It allows for simple unit tests of upstream code that helps ensure the code under test is robust to all possible return failures regardless of how difficult they are to trigger. In order to enable `RCUTILS_MAYBE_RETURN_ERROR`, you must set the global counter.

The counter is set with `RCUTILS_SET_FAULT_INJECTION_COUNT`. The counter counts down for every invocation of `RCUTILS_MAYBE_RETURN_ERROR` and when it reaches 0 will cause `RCUTILS_MAYBE_RETURN_ERROR` to return the designated error. As this macro may be widely placed in the code base, it is necessary to run this in a for loop in the unit tests. For example

```c++
TEST(TestMyFunction, test_function) {
  constexpr int maximum_calls_to_maybe_return = 100;
  for (int i = 0; i < maximum_calls_to_maybe_return; ++i) {
    RCUTILS_SET_FAULT_INJECTION_COUNT(i);
    auto result = function_under_test();
    // unless cleanup needs to occur, result is likely to be ignored
  }
}
```

This sort of test doesn't help you assert that the behavior of the function is correct in face of downstream code failing, but it does let you ensure that your code returns safely and that resources are adequately cleaned up in all possible return paths (with the appropriate analysis tools). 

The high level macro `RCUTILS_CAN_RETURN_WITH_ERROR_OF` is a descriptive macro that includes a call to `RCUTILS_MAYBE_RETURN_ERROR`, with the `RCUTILS_ENABLE_FAULT_INJECTION` compiler flag enabled. It has the extra benefit of providing a quick glance at the expected failure return codes in the function.

Currently, the intention is to use `RCUTILS_CAN_RETURN_WITH_ERROR_OF` once for each error return value, but more fully featured macros could also be included. For example, variadic versions can be introduced like `RCUTILS_CAN_RETURN_WITH_ERRORS_OF`, `RCUTILS_CAN_RETURN_WITH_STRING_ERRORS_OF`, etc.

### Possible Extensions
**RCUTILS_CHECK_RETURN_OK**
If labeling the upstream code is troublesome, an alternative macro like `RCUTILS_CHECK_RETURN_OK` could be created that would replace simple if statements. This would be placed in the downstream code that checks the called functions for error values. That macro would include the call to `RCUTILS_MAYBE_FAIL`, which executes a statement on error (not just return an error value).

```c++
#define RCUTILS_CHECK_RETURN_OK(ret, error_message, error_statement) \
  RCUTILS_MAYBE_FAIL(error_statement); \
  if (ret != RCUTILS_RET_OK) { \
    RCUTILS_SET_ERROR_MESSAGE(error_message); \
    error_statement; \
  }
```

**Targeted fault injection macros**
While `RCUTILS_MAYBE_RETURN_ERROR` allows for robust coverage of numerous possible return paths with very easy unit tests, it doesn't allow for targeted failures. Another extension of this feature would be to give string names to each `RCUTILS_CAN_RETURN_WITH_ERROR_OF` macro that are registered with a global lookup. In the unit test, these macros could then be set to specifically fail and the expected failure behavior can then be asserted.

### DIFF
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -162,11 +162,12 @@ extern "C"
  * should go on a separate line.
  *
  * If in your function, there are expected effects on output parameters that occur during
- * the failure case, then it will introduce a discrepency between fault injection testing and
+ * the failure case, then it will introduce a discrepancy between fault injection testing and
  * production operation. This is because the fault injection will cause the function to return
  * where this macro is used, not at the location the error values are typically returned. To help
- * protect against this scenario you may consider adding unit tests that checks your function does
- * not modify output parameters when it actually returns a failing error code.
+ * protect against this scenario you may consider adding unit tests that check your function does
+ * not modify output parameters when it actually returns a failing error code if it's possible for
+ * your code.
  *
  * If your function is void, this macro can be used without parameters. However, for the above
  * reasoning, there should be no side effects on output parameters for all possible early returns.
@@ -177,8 +178,26 @@ extern "C"
 # define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value) \
   RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR(error_return_value);
 
+/**
+ * \def RCUTILS_CAN_FAIL_WITH
+ * Indicating macro similar to RCUTILS_CAN_RETURN_WITH_ERROR_OF but for use with more complicated
+ * statements.
+ *
+ * The `failure_code` will be executed inside a scoped if block, so any variables declared within
+ * will not be available outside of the macro.
+ *
+ * One example where you might need this version, is if a side-effect may occur within a function.
+ * For example, in snprintf, rcutils_snprintf needs to set both errno and return -1 on failure.
+ * This macro is used to capture both effects.
+ *
+ * \param failure_code Code that is representative of the failure case in this function.
+ */
+# define RCUTILS_CAN_FAIL_WITH(failure_code) \
+  RCUTILS_FAULT_INJECTION_MAYBE_FAIL(failure_code);
+
 #else
 # define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value)
+# define RCUTILS_CAN_FAIL_WITH(failure_code)
 #endif  // defined RCUTILS_ENABLE_FAULT_INJECTION
 
 #ifdef __cplusplus
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -131,6 +131,55 @@ extern "C"
 # define RCUTILS_UNLIKELY(x) (x)
 #endif  // _WIN32
 
+#if defined RCUTILS_ENABLE_FAULT_INJECTION
+#include "rcutils/testing/fault_injection.h"
+
+/**
+ * \def RCUTILS_CAN_RETURN_WITH_ERROR_OF
+ * Indicating macro that the function intends to return possible error value.
+ *
+ * Put this macro as the first line in the function. For example:
+ *
+ * int rcutils_function_that_can_fail() {
+ *   RCUTILS_CAN_RETURN_WITH_ERROR_OF(RCUTILS_RET_INVALID_ARGUMENT);
+ *   ...  // rest of function
+ * }
+ *
+ * For now, this macro just simply calls `RCUTILS_MAYBE_RETURN_ERROR` if fault injection is
+ * enabled. However, for source code, the macro annotation `RCUTILS_CAN_RETURN_WITH_ERROR_OF` helps
+ * clarify that a function may return a value signifying an error and what those are.
+ *
+ * In general, you should only include a return value that originates in the function you're
+ * annotating instead of one that is merely passed on from a called function already annotated with
+ *`RCUTILS_CAN_RETURN_WITH_ERROR_OF`. If you are passing on return values from a called function,
+ * but that function is not annotated with `RCUTILS_CAN_RETURN_WITH_ERROR_OF`, then you might
+ * consider annotating that function first. If for some reason that is not desired or possible,
+ * then annotate your function as if the return values you are passing on originated from your
+ * function.
+ *
+ * If the function can return multiple return values indicating separate failure types, each one
+ * should go on a separate line.
+ *
+ * If in your function, there are expected effects on output parameters that occur during
+ * the failure case, then it will introduce a discrepency between fault injection testing and
+ * production operation. This is because the fault injection will cause the function to return
+ * where this macro is used, not at the location the error values are typically returned. To help
+ * protect against this scenario you may consider adding unit tests that checks your function does
+ * not modify output parameters when it actually returns a failing error code.
+ *
+ * If your function is void, this macro can be used without parameters. However, for the above
+ * reasoning, there should be no side effects on output parameters for all possible early returns.
+ *
+ * \param error_return_value the value returned as a result of an error. It does not need to be
+ * a rcutils_ret_t type. It could also be NULL, -1, a string error message, etc
+ */
+# define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value) \
+  RCUTILS_MAYBE_RETURN_ERROR(error_return_value);
+
+#else
+# define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value)
+#endif  // defined RCUTILS_ENABLE_FAULT_INJECTION
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -145,9 +145,10 @@ extern "C"
  *   ...  // rest of function
  * }
  *
- * For now, this macro just simply calls `RCUTILS_MAYBE_RETURN_ERROR` if fault injection is
- * enabled. However, for source code, the macro annotation `RCUTILS_CAN_RETURN_WITH_ERROR_OF` helps
- * clarify that a function may return a value signifying an error and what those are.
+ * For now, this macro just simply calls `RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR` if fault
+ * injection is enabled. However, for source code, the macro annotation
+ * `RCUTILS_CAN_RETURN_WITH_ERROR_OF` helps clarify that a function may return a value signifying
+ * an error and what those are.
  *
  * In general, you should only include a return value that originates in the function you're
  * annotating instead of one that is merely passed on from a called function already annotated with
@@ -174,7 +175,7 @@ extern "C"
  * a rcutils_ret_t type. It could also be NULL, -1, a string error message, etc
  */
 # define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value) \
-  RCUTILS_MAYBE_RETURN_ERROR(error_return_value);
+  RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR(error_return_value);
 
 #else
 # define RCUTILS_CAN_RETURN_WITH_ERROR_OF(error_return_value)
--- a/include/rcutils/testing/fault_injection.h
+++ b/include/rcutils/testing/fault_injection.h
@@ -104,12 +104,6 @@ int _rcutils_fault_injection_maybe_fail();
 #define RCUTILS_FAULT_INJECTION_GET_COUNT() \
   _rcutils_fault_injection_get_count();
 
-#define RCUTILS_FAULT_INJECTION_INIT() \
-  RCUTILS_FAULT_INJECTION_SET_COUNT(RCUTILS_FAULT_INJECTION_NEVER_FAIL)
-
-#define RCUTILS_FAULT_INJECTION_FINI() \
-  RCUTILS_FAULT_INJECTION_SET_COUNT(RCUTILS_FAULT_INJECTION_NEVER_FAIL)
-
 #define RCUTILS_FAULT_INJECTION_TEST(code) \
   do { \
     int fault_injection_count = 0; \
@@ -117,6 +111,7 @@ int _rcutils_fault_injection_maybe_fail();
       RCUTILS_FAULT_INJECTION_SET_COUNT(fault_injection_count++); \
       code; \
     } while (!rcutils_fault_injection_is_test_complete()); \
+    RCUTILS_FAULT_INJECTION_SET_COUNT(RCUTILS_FAULT_INJECTION_NEVER_FAIL); \
   } while(0)
 
 #else  // RCUTILS_ENABLE_FAULT_INJECTION
@@ -128,11 +123,7 @@ int _rcutils_fault_injection_maybe_fail();
 
 #define RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR(msg, error_statement)
 
-#define RCUTILS_FAULT_INJECTION_INIT() \
-  printf("Fault injection is disabled, skipping\n"); \
-  return;
-
-#define RCUTILS_FAULT_INJECTION_FINI()
+#define RCUTILS_FAULT_INJECTION_TEST(code) return;
 
 #endif  // defined RCUTILS_ENABLE_FAULT_INJECTION
 
--- a/include/rcutils/testing/fault_injection.h
+++ b/include/rcutils/testing/fault_injection.h
@@ -1,0 +1,90 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCUTILS__TESTING_MACROS_H_
+#define RCUTILS__TESTING_MACROS_H_
+#include <stdio.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+void _rcutils_set_fault_injection_count(int count);
+
+int _rcutils_maybe_fail();
+
+#if defined RCUTILS_ENABLE_FAULT_INJECTION
+
+/**
+ * \def RCUTILS_MAYBE_RETURN_ERROR
+ * \brief This macro checks and decrements a static global variable atomic counter and returns
+ * `return_value_on_error` if 0.
+ *
+ * This macro is not a function itself, so when this macro returns it will cause the calling
+ * function to return with the return value.
+ *
+ * Set the counter with `RCUTILS_SET_FAULT_INJECTION_COUNT`. If the count is less than 0, then
+ * `RCUTILS_MAYBE_RETURN_ERROR` will not cause an early return.
+ *
+ * This macro is thread-safe, and ensures that at most one invocation results in a failure for each
+ * time the fault injection counter is set with `RCUTILS_SET_FAULT_INJECTION_COUNT`
+ *
+ * \param return_value_on_error the value to return in the case of fault injected failure.
+ */
+#define RCUTILS_MAYBE_RETURN_ERROR(return_value_on_error) \
+  if (0 == _rcutils_maybe_fail()) { \
+    return return_value_on_error; \
+  }
+
+/**
+ * \def RCUTILS_SET_FAULT_INJECTION_COUNT
+ * \brief Atomically set the fault injection counter.
+ *
+ * There will be at most one fault injected failure per call to RCUTILS_SET_FAULT_INJECTION_COUNT.
+ * To test all reachable fault injection locations, call this macro inside a for loop with
+ * sufficient iterations setting count to the loop iteration variable. For example:
+ *
+ *  for (int i = 0; i < SUFFICIENTLY_LARGE_ITERATION_COUNT; ++i) {
+ *    RCUTILS_SET_FAULT_INJECTION_COUNT(i);
+ *    ... // Call function under test
+ *  }
+ *
+ * Where SUFFICIENTLY_LARGE_ITERATION_COUNT is a value larger than the maximum expected calls to
+ * `RCUTILS_MAYBE_RETURN_ERROR`. In your fault injection unit test, it is recommended to run one
+ * last iteration with the fault injection counter set to this maximum value and validate that the
+ * results of the call to the function under test would result in the same thing as if no fault
+ * injection was used. This will help ensure that this maximum value is suitable and will call
+ * attention to maintainers if it needs to be increased because more instances of
+ * RCUTILS_MAYBE_RETURN_ERROR were introduced.
+ *
+ * \param count The count to set the fault injection counter to. If count is negative, then fault
+ * injection errors will be disabled. The counter is globally initialized to -1.
+ */
+#define RCUTILS_SET_FAULT_INJECTION_COUNT(count) \
+  _rcutils_set_fault_injection_count(count);
+
+#else  // RCUTILS_ENABLE_FAULT_INJECTION
+
+#define RCUTILS_SET_FAULT_INJECTION_COUNT(count)
+
+#define RCUTILS_MAYBE_RETURN_ERROR(msg, error_statement)
+
+#endif  // defined RCUTILS_ENABLE_FAULT_INJECTION
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCUTILS__TESTING_MACROS_H_
--- a/include/rcutils/testing/fault_injection.h
+++ b/include/rcutils/testing/fault_injection.h
@@ -16,12 +16,20 @@
 #define RCUTILS__TESTING_MACROS_H_
 #include <stdio.h>
 
+#include "rcutils/stdatomic_helper.h"
+
 #ifdef __cplusplus
 extern "C"
 {
 #endif
 
+#define RCUTILS_FAULT_INJECTION_NEVER_FAIL -1
+
+#define RCUTILS_FAULT_INJECTION_FAIL_NOW 0
+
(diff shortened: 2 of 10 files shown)


src/snprintf.c
```diff
@@ -39,6 +39,8 @@ rcutils_snprintf(char * buffer, size_t buffer_size, const char * format, ...)
 int
 rcutils_vsnprintf(char * buffer, size_t buffer_size, const char * format, va_list args)
 {
+  RCUTILS_CAN_RETURN_WITH_ERROR_OF(-1);
```
clalancette, line 42: The main concern here is that we are going to litter our code with these macros.  It's not the end of the world, but it is noisy and can make the code harder to read.

I have 2 separate suggestions:
1.  Before we start down this road, let's see what @hidmic's work on Mimick looks like.  If we can make that work, that will obviate the need for this, I think (I'm currently ignoring the fact that Mimick doesn't work for C++, while this one would).
1.  If we can't get Mimick to work for some reason, another alternative is to use the `LD_PRELOAD` mechanism to override the symbols as needed.  This is less desirable than Mimick since it is limited to Linux-only, but I think we can make the case (at least for now) that doing it on one platform is better than doing it on zero platforms.

The benefit to both solutions above is that there is no changes to the code, just changes to the tests.  But I'd also like to hear what others think.

brawner: I pushed this PR last night, and intended to give a more full writeup. Please take a look at the above PR description for some more information as well as possible extensions.

The approach taken in this PR allows for immediate cross-platform, cross-language utilization with simple unit tests with no external dependencies. You can write a unit test that easily verifies your code is hardened against failures in upstream libraries. See https://github.com/ros2/rcl_logging/pull/48 for an example.

Another approach was to add an `RCUTILS_CHECK_` macro that replaced simple if statements. Based on feedback from you and @wjwwood that also desired to see that we test the same code that we ship, I took this approach. The use of a macro like `RCUTILS_CAN_RETURN_WITH_ERROR_OF` allows for testing shipped production libraries against modified upstream libraries, which is not something our CI or buildfarm are setup for today but if desired could be feasibly updated in the future.

I took care in the design of `RCUTILS_CAN_RETURN_WITH_ERROR_OF` to ensure that it wasn't just litter. It actually serves an additional purpose of describing intended failure cases of the function. While docblocks typically try to capture the possible failures cases, this would be another approach that can also be integrated with unit testing. It may be difficult for a maintainer to see all the possible return values of a high level function that it passes on from lower level functions. However, with unit tests utilizing `RCUTILS_MAYBE_RETURN_ERROR`, you can verify that all possible return values are sufficiently understood.

This PR is intended as a starting point, and I want to make it as useful as possible. There are several possible feature additions that I think could fill it out, but I would like to get some more buy in as well.

brawner: Thank you, it does help clarify your thinking. My hope is that this sort of feature is used in conjunction with mocking like what @hidmic is investigating. That sort of testing would allow for more targeted failures, while this approach especially with the for loop example might be more like throwing a grenade at your code and seeing if it survives.

You're absolutely right that there are a large number of places where the `RCUTILS_CAN_RETURN_WITH_ERROR_OF` macro can be placed, and it would be overwhelming to put them in every single location at the outset, or even for every contribution. But I also think it's possible to add this macro in piecemeal, and we can start by only adding it in when it's needed for unit tests. For example, to push rcl_logging_spdlog to 100% coverage, you only need `rcutils_snprintf` to fail, but given the correct inputs that function probably never will on our supported platforms.

My suggestion with this sort of change is to start small and expand it as its role in this codebase is better understood. That's partially the reason why I only introduce one flavor of `RCUTILS_CAN_RETURN_WITH_ERROR_OF`, even though I also want a variadic version.

That macro is empty if fault injection is not enabled, which I think helps make it easier to remove in the future. The example unit test in rcl_logging_spdlog would pass if `RCUTILS_CAN_RETURN_WITH_ERROR_OF` did not exist at all. We could also properly deprecate `RCUTILS_SET_FAULT_INJECTION_COUNT` to help point external developers to the appropriate replacement.

I also agree that seeing this macro at the top of a function would be strange and unusual to many developers, particularly newer ROS 2 users who are unfamiliar with the code. While I tried to be careful about the macro name, I'm also very open to rewording `RCUTILS_CAN_RETURN_WITH_ERROR_OF`. I chose the wording so that it's hopefully clear to developers unfamiliar with it that doesn't change the nominal behavior of the function, while conveying more information about the intent of the function's behavior if you are familiar with it.

hidmic: > Before we start down this road, let's see what @hidmic's work on Mimick looks like. If we can make that work, that will obviate the need for this, I think (I'm currently ignoring the fact that Mimick doesn't work for C++, while this one would).

@clalancette I think this serves a different use case. Mocking is good for white-box unit testing, while this is performing black-box integration testing to see how the system, with all its side effects, behaves in an exceptional event (akin to exception safety verification). Could we use mocks to implement it? Yes, but we'd end up coupling a package to its dependencies' implementation and recreating side effects -- possible, but much harder.

IMHO, `RCUTILS_CAN_RETURN_WITH_ERROR_OF`, though atypical, isn't strictly worse than the prologue of `RCUTILS_CHECK_ARGUMENT_FOR_NULL` our C functions usually have.

hidmic:
> RCUTILS_CHECK_RETURN_OK

Hmm, if the call succeeded, with its side effects, cleanup code may not be enough when simulating an error. Also, if the callee does fault injection, there's no need for the caller to do the same. For those callees that don't, perhaps:

```c++
auto result = RCUTILS_MAY_FAIL_WITH(/* bad_result */, /* expression */);
```

can bypass the expression? It does clutter the code a bit though.

include/rcutils/testing/fault_injection.h
```diff
@@ -0,0 +1,90 @@
...
+void _rcutils_set_fault_injection_count(int count);
+
+int _rcutils_maybe_fail();
+
+#if defined RCUTILS_ENABLE_FAULT_INJECTION
+
+/**
+ * \def RCUTILS_MAYBE_RETURN_ERROR
+ * \brief This macro checks and decrements a static global variable atomic counter and returns
+ * `return_value_on_error` if 0.
```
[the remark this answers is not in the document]
brawner: There are some examples where a failure in one location is likely to result in a failure in another location. For example, allocations may continue to fail, or file system functions may continue to fail. For the sake of coverage, there are locations where it would take two failures to reach some lines of code. Your example of cleanup is a good one.

I think there are a few of potential solutions:
1. Setting potential alternate behaviors when the counter is negative, which is what you are suggesting. I think the idea works great, it would just require writing multiple unit tests for each behavior.

2. Instead of leaving it as -1, the `RCUTILS_SET_FAULT_COUNT` macro could have a variadic version which takes as parameters counts that it's reset to after it reaches 0. 
```C
RCUTILS_SET_FAULT_COUNT(0, 0)
```
For example, would trip a second failure immediately after its first failure. This would let you test more types of failures, and then we could use this type of macro to check that code is not only one-fault tolerant, but two-fault tolerant, three-fault tolerant etc.

3. For instances where testing two particular faults is important, then we could introduce a named fault injection. Instead of `RCUTILS_CAN_RETURN_WITH_ERROR`, there would be a `RCUTILS_CAN_RETURN_WITH_ERROR_NAMED`. Then in the unit test, we could target those locations to fail explicitly in a unit test.
```C
RCUTILS_SET_FAULT_INJECTION_NAMED(...);
RCUTILS_SET_FAULT_INJECTION_NAMED(...);
... // Run test code
```
If we do introduce this one at a later date, I could see it replacing the unnamed version macro entirely. This idea requires multiple unit tests for each targeted failed injection, but is also the only option that allows for targeted injections in the first place. 

Do you have any thoughts about the other options suitability for the situation you are describing?

include/rcutils/testing/fault_injection.h
```diff
@@ -0,0 +1,90 @@
...
+ *    ... // Call function under test
+ *  }
+ *
+ * Where SUFFICIENTLY_LARGE_ITERATION_COUNT is a value larger than the maximum expected calls to
+ * `RCUTILS_MAYBE_RETURN_ERROR`. In your fault injection unit test, it is recommended to run one
+ * last iteration with the fault injection counter set to this maximum value and validate that the
+ * results of the call to the function under test would result in the same thing as if no fault
+ * injection was used. This will help ensure that this maximum value is suitable and will call
+ * attention to maintainers if it needs to be increased because more instances of
+ * RCUTILS_MAYBE_RETURN_ERROR were introduced.
```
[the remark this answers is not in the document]
brawner: I like the suggestion of switching the for loop to a do while loop and adding a check that the global counter is > 0. But I'm not sure I follow your code example. I might just recommend leaving it up to the unit test writer. But I will definitely add a getter now that there is some good reasoning for it.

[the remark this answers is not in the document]
brawner: Added in https://github.com/ros2/rcutils/pull/264/commits/97dc3694370de9b6e30e7f05bde3f4764855cb4b

include/rcutils/macros.h
```diff
@@ -131,6 +131,55 @@ extern "C"
...
+ *
+ * If the function can return multiple return values indicating separate failure types, each one
+ * should go on a separate line.
+ *
+ * If in your function, there are expected effects on output parameters that occur during
+ * the failure case, then it will introduce a discrepency between fault injection testing and
+ * production operation. This is because the fault injection will cause the function to return
+ * where this macro is used, not at the location the error values are typically returned. To help
+ * protect against this scenario you may consider adding unit tests that checks your function does
+ * not modify output parameters when it actually returns a failing error code.
```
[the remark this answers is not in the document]
hidmic: Fair enough. I'd like to point out though that while:

> While many cases might have side effects after returning through an error case, my limited experience with this code base suggests those typically result in an undefined state more than an expected partially completed state.

often holds true, it's bad practice. Really bad practice. If your program is left in an undefined state after returning an error, you may as well terminate it.

include/rcutils/testing/fault_injection.h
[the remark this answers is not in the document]
hidmic: Basically, rinse and repeat until you've inject a fault in every possible location, testing the outcome of your statement under test on success and on failure (as I mentioned above, outcome should be defined on error too).

include/rcutils/testing/fault_injection.h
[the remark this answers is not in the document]
hidmic: I think solution (2.) would be best. We don't really know what we're hitting, so N-fault tolerant sounds as good as it gets.

include/rcutils/testing/fault_injection.h
```diff
@@ -51,6 +54,8 @@ int _rcutils_maybe_fail();
  */
 #define RCUTILS_MAYBE_RETURN_ERROR(return_value_on_error) \
   if (RCUTILS_FAULT_INJECTION_FAIL_NOW == _rcutils_maybe_fail()) { \
+    printf( \
```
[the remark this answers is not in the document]
hidmic: I don't see why not having an `RCUTILS_LOG_DEBUG` call here.

include/rcutils/testing/fault_injection.h
```diff
@@ -0,0 +1,134 @@
...
+ */
+#define RCUTILS_FAULT_INJECTION_GET_COUNT() \
+  _rcutils_fault_injection_get_count();
+
+#define RCUTILS_FAULT_INJECTION_TEST(code) \
+  do { \
+    int fault_injection_count = 0; \
+    do { \
+      RCUTILS_FAULT_INJECTION_SET_COUNT(fault_injection_count++); \
+      code; \
```
hidmic, line 112: @brawner nice! I still think we need a way to disable fault injection temporarily within the `code` statement, otherwise you may hit a fault while handling or checking the outcome of the piece of functionality that is actually under test.

include/rcutils/testing/fault_injection.h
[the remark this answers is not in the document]
brawner: RCUTILS_LOG_DEBUG is not defined inside the rcutils library unfortunately.

include/rcutils/testing/fault_injection.h
brawner: That can be done by way of: 
```
int current_count = RCUTILS_FAULT_INJECTION_GET_COUNT();
RCUTILS_FAULT_INJECTION_SET_COUNT(RCUTILS_FAULT_INJECTION_NEVER_FAIL);
...
RCUTILS_FAULT_INJECTION_SET_COUNT(current_count);
```

Do you think we need something more like:

```
RCUTILS_FAULT_INJECTION_PAUSE_TEST();
...
RCUTILS_FAULT_INJECTION_RESUME_TEST();
```

brawner:
Builds and tests of the following packages `rcutils`, `rcl_logging_spdlog`, `rosidl_typesupport_c`, `rosidl_typesupport_cpp`, `rcl`, `rcl_action`, `rcl_lifecycle`, `rosidl_runtime_c`, `rmw_dds_common`, `rmw`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11916/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9709/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11850/)

brawner:
Rebuilding
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11917/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6962/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9710/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11852/)

brawner:
Today Stephen learned about ATOMIC_VAR_INIT. I also removed the get/set macros and just made the functions public.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11918/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6963/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9711/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11853/)

brawner:
Rebasing rosidl_typesupport_c onto new bug fix!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11938/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6981/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9728/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11877/)

brawner:
rosidl_typesupport_c fix https://github.com/ros2/rosidl_typesupport/pull/82

brawner:
@hidmic @clalancette I think this is ready for a re-review. The above jobs are now working well, with the recent bug fix and a new push to rcl_action, which should resolve the windows build warning.

include/rcutils/macros.h
```diff
@@ -131,6 +131,56 @@ extern "C"
...
+ * but that function is not annotated with `RCUTILS_CAN_RETURN_WITH_ERROR_OF`, then you might
+ * consider annotating that function first. If for some reason that is not desired or possible,
+ * then annotate your function as if the return values you are passing on originated from your
+ * function.
+ *
+ * If the function can return multiple return values indicating separate failure types, each one
+ * should go on a separate line.
+ *
+ * If in your function, there are expected effects on output parameters that occur during
+ * the failure case, then it will introduce a discrepency between fault injection testing and
```
hidmic, line 165: @brawner typo:
```suggestion
 * the failure case, then it will introduce a discrepancy between fault injection testing and
```

include/rcutils/macros.h
```diff
@@ -131,6 +131,56 @@ extern "C"
...
+ * production operation. This is because the fault injection will cause the function to return
+ * where this macro is used, not at the location the error values are typically returned. To help
+ * protect against this scenario you may consider adding unit tests that checks your function does
```
hidmic, line 168: @brawner nit: 
```suggestion
 * protect against this scenario you may consider adding unit tests that check your function does
```

include/rcutils/macros.h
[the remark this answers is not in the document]
hidmic: > (does) not modify output parameters when it actually returns a failing error code.

Meta: that's not always true or possible, finalization functions being a good counterexample. I don't think the limitation should be addressed here though (or ever if mocks help us cover those cases).

include/rcutils/testing/fault_injection.h
```diff
@@ -0,0 +1,128 @@
...
+ * to an incrementing count variable.
+ *
+ *  for (int i = 0; i < SUFFICIENTLY_LARGE_ITERATION_COUNT; ++i) {
+ *    RCUTILS_FAULT_INJECTION_SET_COUNT(i);
+ *    ... // Call function under test
+ *  }
+ * ASSERT_LT(RCUTILS_FAULT_INJECTION_NEVER_FAIL, RCUTILS_FAULT_INJECTION_GET_COUNT());
+ *
+ * Where SUFFICIENTLY_LARGE_ITERATION_COUNT is a value larger than the maximum expected calls to
+ * `RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR`. This last assertion just insures that your choice for
```
hidmic, line 53: @brawner not a native speaker here (so take it with a grain of salt), but
```suggestion
 * `RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR`. This last assertion just ensures that your choice for
```
`insures` sounds like insurance.

include/rcutils/testing/fault_injection.h
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * This macro is thread-safe, and ensures that at most one invocation results in a failure for each
+ * time the fault injection counter is set with `RCUTILS_FAULT_INJECTION_SET_COUNT`
+ *
+ * \param return_value_on_error the value to return in the case of fault injected failure.
+ */
+#define RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR(return_value_on_error) \
+  if (RCUTILS_FAULT_INJECTION_FAIL_NOW == _rcutils_fault_injection_maybe_fail()) { \
+    printf( \
+      "%s:%d Injecting fault and returning " #return_value_on_error "\n", __FILE__, __LINE__); \
```
hidmic, line 110: @brawner why not using debug `logs` or safe fwrites to `stderr`?

include/rcutils/testing/fault_injection.h
hidmic: Yeah, some sugar would be nice. I'd personally like to have something along the lines of: 

```c++
RCUTILS_FAULT_INJECTION_NO_INJECT({
  // ...
});
```

src/snprintf.c
```diff
@@ -39,6 +39,8 @@ rcutils_snprintf(char * buffer, size_t buffer_size, const char * format, ...)
...
+
   if (NULL == format) {
     errno = EINVAL;
```
hidmic, line 45: @brawner a no-op for now, but here's another side-effect that we're missing.

include/rcutils/macros.h
```diff
@@ -131,6 +131,56 @@ extern "C"
...
+ *
+ * int rcutils_function_that_can_fail() {
+ *   RCUTILS_CAN_RETURN_WITH_ERROR_OF(RCUTILS_RET_INVALID_ARGUMENT);
+ *   ...  // rest of function
+ * }
+ *
+ * For now, this macro just simply calls `RCUTILS_FAULT_INJECTION_MAYBE_RETURN_ERROR` if fault
+ * injection is enabled. However, for source code, the macro annotation
+ * `RCUTILS_CAN_RETURN_WITH_ERROR_OF` helps clarify that a function may return a value signifying
+ * an error and what those are.
```
hidmic, line 151: @brawner meta, thinking out loud: a string literal that describes the situation in which said error can be returned would be valuable for documentation and could even be printed and/or logged e.g.:

```
RCUTILS_CAN_RETURN_WITH_ERROR_OF(nullptr, "on failed publisher name allocation");
RCUTILS_CAN_RETURN_WITH_ERROR_OF(nullptr, "on failed rmw publisher initialization");
```

It also sets the scene for future handling of side-effects.

include/rcutils/testing/fault_injection.h
hidmic: > RCUTILS_LOG_DEBUG is not defined inside the rcutils library unfortunately.

Just saw this. That's unfortunate, yeah.

hidmic: Hmm wait a sec, `logging_macros.h` is available at build time (see [CMakeLists.txt](https://github.com/ros2/rcutils/blob/ed1f549b02a2054597aa339cbf529ab1e2151baf/CMakeLists.txt#L111-L120)).

src/snprintf.c
brawner: I've added a new macro, `RCUTILS_CAN_FAIL_WITH` and `RCUTILS_FAULT_INJECTION_MAYBE_FAIL` for this condition.

include/rcutils/testing/fault_injection.h
brawner: You are correct

include/rcutils/macros.h
brawner: Done

include/rcutils/macros.h
brawner: Done

include/rcutils/testing/fault_injection.h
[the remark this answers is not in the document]
brawner: This can be added in a followup PR when the functionality becomes necessary for better coverage.

brawner:
More testing.

Standard builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11962/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6998/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11900/)

Debug builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6999/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9749/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11901/)

Linux only, rmw specific builds:
* Connext [Build Status](https://ci.ros2.org/job/ci_linux/11964/)
* CycloneDDS [Build Status](https://ci.ros2.org/job/ci_linux/11965/)

brawner:
Testing this patch of rcutils against current ros2 branches with `--packages-above rcutils` to ensure I'm not breaking anything.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11966/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7000/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11902/)

brawner:
All the unstable builds for the wider array of packages were caused by a cpplint issue with rcl_action. Addressed that and pushed to that respective branch. Failures for builds against current branches of ros 2 with this PR were due to unrelated qt_gui_cpp issue.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11969/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7003/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9753/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11907/)

brawner:
Coverage results with current fault injection branches:

* [rosidl_typesupport_c](https://github.com/ros2/rosidl_typesupport/pull/80) -- [100%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rosidl_typesupport_rosidl_typesupport_c_src/)
* [rosidl_typesupport_cpp](https://github.com/ros2/rosidl_typesupport/pull/80) -- [100%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rosidl_typesupport_rosidl_typesupport_cpp_src/)
* [rosidl_runtime_c](https://github.com/ros2/rosidl/pull/509) -- [97%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rosidl_rosidl_runtime_c_src/)
* [rcl_logging_spdlog](https://github.com/ros2/rcl_logging/pull/49) - [99%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rcl_logging_rcl_logging_spdlog_src/)
* [rcl_action](https://github.com/ros2/rcl/pull/730) -- [95%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rcl_rcl_action_src_rcl_action/)
* [rcl_lifecycle](https://github.com/ros2/rcl/pull/731) -- [97%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rcl_rcl_lifecycle_src/)
* [rmw_dds_common](https://github.com/ros2/rmw_dds_common/pull/27) -- [99%](https://ci.ros2.org/job/ci_linux/11962/cobertura/src_ros2_rmw_dds_common_rmw_dds_common_src/)

include/rcutils/testing/fault_injection.h
hidmic: I'd bet that's because the library doesn't list the custom command that generates the header file as a dependency. At any rate, it can be solved in a follow-up PR.

brawner:
Testing this patch against current sources one more time after yesterdays commits.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12000/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7027/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9778/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11935/)

brawner:
All tests pass, except `qt_gui_cpp` which fails due to the same gcov issue. That package and packages that depend on it are excluded from the nightly coverage jobs (I'm assuming for this reason).

PR #265, opened 2020-07-03
Add mock test for rcutils/strerror

Blast545:
This PR includes a test for the `rcutils_strerror` function that required making a mock of the `strerror_r / strerror_s`.

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -24,6 +24,7 @@
 #include <string>
 
 #include "rcutils/strerror.h"
+#include "mimick_vendor/mimic.h"
 
 TEST(test_strerror, get_error) {
   // cleaning possible errors
@@ -53,4 +54,35 @@ TEST(test_strerror, get_error) {
     "Calling rcutils_strerror with an errno of '" << errno <<
     "' did not cause the expected error message.";
 #endif
+}
+
+/*
+   Define the blueprint of a mock identified by `strerror_r_proto`
+   strerror_r signature:
+   int strerror_r(int errnum, char *buf, size_t buflen);
+*/
+mmk_mock_define (strerror_r_mock, int, char *, size_t);
+
+/* Mocking test example */
+TEST(test_strerror, get_error) {
+  /* Mock the strerror_r function in the current module using
+     the `strerror_r_mock` blueprint. */
+  mmk_mock("strerror_r@self", strerror_r_mock);
+
+  /* Tell the mock to return NULL and set errno to ENOMEM
+     whatever the given parameter is. */
+  mmk_when(strerror_r(mmk_any(int, char *, size_t)),
+	   .then_return = EINVAL);
+
+  // Now normal usage of the function returning unexpected EINVAL
+  // error for the internal strerror_r
+  // This works only for POSIX
+
+  // Set the error "No such file or directory"
+  errno = 2;
+  char error_string[1024];
+  rcutils_strerror(error_string, sizeof(error_string));
+  ASSERT_STREQ(error_string, "Failed to get error");
+
+  mmk_reset(strerror_r);
 }
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -76,12 +76,12 @@ errno_t mocked_windows_strerror(char * buf, rsize_t bufsz, errno_t errnum)
   return errnum;
 }
 
-/* Mocking test example */
+// Mocking test example
 TEST(test_strerror, test_mock) {
-  /* Mock the strerror_s function in the current module using
-     the `strerror_s_mock` blueprint. */
+  // Mock the strerror_s function in the current module using
+  // the `strerror_s_mock` blueprint.
   mmk_mock(RCUTILS_STRINGIFY(strerror_s) "@lib:rcutils", strerror_s_mock);
-  /* Tell the mock to call mocked_windows_strerror instead*/
+  // Tell the mock to call mocked_windows_strerror instead
   mmk_when(
     strerror_s(mmk_any(errno_t), mmk_any(char *), mmk_any(rsize_t), mmk_any(errno_t)),
     .then_call = (mmk_fn) mocked_windows_strerror);
@@ -104,12 +104,12 @@ char * mocked_gnu_strerror(int errnum, char * buf, size_t buflen)
   return buf;
 }
 
-/* Mocking test example */
+// Mocking test example
 TEST(test_strerror, test_mock) {
-  /* Mock the strerror_r function in the current module using
-     the `strerror_r_mock` blueprint. */
+  // Mock the strerror_r function in the current module using
+  // the `strerror_r_mock` blueprint.
   mmk_mock(RCUTILS_STRINGIFY(strerror_r) "@lib:rcutils", strerror_r_mock);
-  /* Tell the mock to call mocked_gnu_strerror instead */
+  // Tell the mock to call mocked_gnu_strerror instead
   mmk_when(
     strerror_r(mmk_any(int), mmk_any(char *), mmk_any(size_t) ),
     .then_call = (mmk_fn) mocked_gnu_strerror);
@@ -124,13 +124,14 @@ TEST(test_strerror, test_mock) {
 
 #else
 mmk_mock_define(strerror_r_mock, int, int, char *, size_t);
-/* Mocking test example */
+
+// Mocking test example
 TEST(test_strerror, test_mock) {
-  /* Mock the strerror_r function in the current module using
-     the `strerror_r_mock` blueprint. */
+  // Mock the strerror_r function in the current module using
+  // the `strerror_r_mock` blueprint.
   mmk_mock(RCUTILS_STRINGIFY(strerror_r) "@lib:rcutils", strerror_r_mock);
-  /* Tell the mock to return NULL and set errno to EINVAL
-     whatever the given parameter is. */
+  // Tell the mock to return NULL and set errno to EINVAL
+  // whatever the given parameter is.
   mmk_when(
     strerror_r(mmk_any(int), mmk_any(char *), mmk_any(size_t) ),
     .then_return = mmk_val(int, EINVAL));
@@ -142,5 +143,4 @@ TEST(test_strerror, test_mock) {
   ASSERT_STREQ(error_string, "Failed to get error");
   mmk_reset(strerror_r);
 }
-
 #endif
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -91,7 +91,7 @@ TEST(test_strerror, test_mock) {
   char error_string[1024];
   rcutils_strerror(error_string, sizeof(error_string));
   ASSERT_STREQ(error_string, "Failed to get error");
-  mmk_reset(strerror_s);
+  mmk_reset(reinterpret_cast<mmk_fn>strerror_s);
 }
 
 #elif defined(_GNU_SOURCE) && (!defined(ANDROID) || __ANDROID_API__ >= 23)
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -91,7 +91,7 @@ TEST(test_strerror, test_mock) {
   char error_string[1024];
   rcutils_strerror(error_string, sizeof(error_string));
   ASSERT_STREQ(error_string, "Failed to get error");
-  mmk_reset(reinterpret_cast<mmk_fn>(strerror_s));
+  mmk_reset(strerror_s_mock);
 }
 
 #elif defined(_GNU_SOURCE) && (!defined(ANDROID) || __ANDROID_API__ >= 23)
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -80,7 +80,7 @@ errno_t mocked_windows_strerror(char * buf, rsize_t bufsz, errno_t errnum)
 TEST(test_strerror, test_mock) {
   // Mock the strerror_s function in the current module using
   // the `strerror_s_mock` blueprint.
-  mmk_mock(RCUTILS_STRINGIFY(strerror_s) "@lib:rcutils", strerror_s_mock);
+  malloc_mock mock = mmk_mock(RCUTILS_STRINGIFY(strerror_s) "@lib:rcutils", strerror_s_mock);
   // Tell the mock to call mocked_windows_strerror instead
   mmk_when(
     strerror_s(mmk_any(char *), mmk_any(rsize_t), mmk_any(errno_t)),
@@ -91,7 +91,7 @@ TEST(test_strerror, test_mock) {
   char error_string[1024];
   rcutils_strerror(error_string, sizeof(error_string));
   ASSERT_STREQ(error_string, "Failed to get error");
-  mmk_reset(strerror_s_mock);
+  mmk_reset(mock);
 }
 
 #elif defined(_GNU_SOURCE) && (!defined(ANDROID) || __ANDROID_API__ >= 23)
--- a/test/test_strerror.cpp
+++ b/test/test_strerror.cpp
@@ -12,13 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifdef _GNU_SOURCE
-#undef _GNU_SOURCE
-#include <string>
-#else
-#include <string>
-#endif
-
 #include <gtest/gtest.h>
 
 #include <sys/types.h>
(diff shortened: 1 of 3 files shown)


test/test_strerror.cpp
```diff
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifdef _GNU_SOURCE
+#undef _GNU_SOURCE
+#include <string>
+#else
+#include <string>
+#endif
```
hidmic, line 20: @brawner @wjwwood @clalancette this is a non-portable workaround for an issue @Blast545 and I stumbled upon. Turns out `librcutls.so` and this test's executable where linking against different `strerror_r` versions. Different `libc` versions, and different symbol names (`strerror_r` is actually a macro for the `libc` version `librcutils.so` links to on Ubuntu 20.04).

This seems like a significant pain point whenever mocking system libraries if it turns out such practices are widespread (and I'd think so). We need a way to ensure both test and library use and link against the same third party OR use a wrapper with a stable ABI instead of system libraries directly.

clalancette: > @brawner @wjwwood @clalancette this is a non-portable workaround for an issue @Blast545 and I stumbled upon. Turns out `librcutls.so` and this test's executable where linking against different `strerror_r` versions.

Huh, weird.  Can you explain a little more of the details?  I'm just trying to get a handle on how common this will be.

hidmic: _holiday mode OFF_ 

> Can you explain a little more of the details? I'm just trying to get a handle on how common this will be.

Maybe ELF dumps help.

Output fragment of `readelf --relocs path/to/librcutils.so`:
```
Relocation section '.rela.plt' at offset 0x3480 contains 107 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
...
0000000180c8  000e00000007 R_X86_64_JUMP_SLO 0000000000000000 __xpg_strerror_r@GLIBC_2.3.4 + 0
000000018358  007000000007 R_X86_64_JUMP_SLO 000000000000e250 rcutils_strerror + 0
...
```

Output fragment of `readelf --relocs path/to/test_strerror`:
```
Relocation section '.rela.plt' at offset 0x6780 contains 206 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
...
0000000799e8  001700000007 R_X86_64_JUMP_SLO 0000000000000000 strerror@GLIBC_2.2.5 + 0
000000079bd8  005700000007 R_X86_64_JUMP_SLO 0000000000000000 rcutils_strerror + 0
...
```
Relevant `glibc` [`string.h`](https://github.molgen.mpg.de/git-mirror/glibc/blob/20003c49884422da7ffbc459cdeee768a6fee07b/string/string.h#L406-L441) header section.

I don't know exactly why this happens, though the C++ runtime may have something do with it (`librcutils.so` is C only).

_holiday mode on_

test/test_strerror.cpp
```diff
@@ -54,3 +61,38 @@ TEST(test_strerror, get_error) {
...
+     the `strerror_r_mock` blueprint. */
+  // mmk_mock("strerror_r@rcutils", strerror_r_mock);
+  // mmk_mock("__xpg_strerror_r@self", strerror_r_mock);
+  // mmk_mock("strerror_r@self", strerror_r_mock);
+  mmk_mock("__xpg_strerror_r@lib:rcutils", strerror_r_mock);
+
+  /* Tell the mock to return NULL and set errno to ENOMEM
+     whatever the given parameter is. */
+  mmk_when(
+    strerror_r(mmk_any(int), mmk_any(char *), mmk_any(size_t) ),
```
brawner, line 84: This sets the global behavior of the mocked function? Are there reasonable scenarios where you might want the default behavior of the mocked function generally, but have the mocked version for one specific call? How might you accomplish that?

test/test_strerror.cpp
```diff
@@ -54,3 +61,38 @@ TEST(test_strerror, get_error) {
...
+
+  // Now normal usage of the function returning unexpected EINVAL
+  // error for the internal strerror_r
+  // This works only for POSIX
+
+  // Set the error "No such file or directory"
+  errno = 2;
+  char error_string[1024];
+  rcutils_strerror(error_string, sizeof(error_string));
+  ASSERT_STREQ(error_string, "Failed to get error");
```
brawner, line 95: I'm sure this is a case-by-case matter, but what risks are there for maintainers adding calls to the mocked function that sits between the unit test and the targeted invocation? It seems like unit tests that call high-level functions, but mock low-level functions would be pretty brittle, and you would need to mock functions whose number of uses in the code under test is unlikely to change. `rcutils_strerror` is a good example where the number of uses of `strerror_r` are unlikely to change. Whereas mocking something like `rcutils_strndup` during complex code like `rclcpp::init` would result in behaviors that are difficult to assert.

test/test_strerror.cpp
hidmic: Yeah, that's something I thought about, but unfortunately `Mimick` mocks do not explicitly support that feature. I only know of `gMock` for a framework that can do that though.

test/test_strerror.cpp
hidmic: I don't see a fundamental risk, besides the maintenance burden that comes from white-box testing and the coupling between test and implementation it entails. It's important to bear in mind that mocking is not a tool for fault injection in a running system, but a way to unit test the interactions between the entity under test and its direct dependencies. In an `rclcpp::init` test, one would mock `rcutils_strndup` along with all other direct dependencies to ensure correct behavior.

test/test_strerror.cpp
```diff
@@ -54,3 +61,38 @@ TEST(test_strerror, get_error) {
...
+*/
+mmk_mock_define(strerror_r_mock, int, char *, size_t);
+
+/* Mocking test example */
+TEST(test_strerror, test_mock) {
+  /* Mock the strerror_r function in the current module using
+     the `strerror_r_mock` blueprint. */
+  // mmk_mock("strerror_r@rcutils", strerror_r_mock);
+  // mmk_mock("__xpg_strerror_r@self", strerror_r_mock);
+  // mmk_mock("strerror_r@self", strerror_r_mock);
```
hidmic, line 78: @Blast545 nit: drop code that's commented out.

test/test_strerror.cpp
```diff
@@ -54,3 +61,38 @@ TEST(test_strerror, get_error) {
...
+  mmk_mock("__xpg_strerror_r@lib:rcutils", strerror_r_mock);
```
hidmic, line 79: @Blast545 hmm the symbol is non-portable. Maybe something like:
```suggestion
  mmk_mock(RCUTILS_STRINGIFY(strerror_r) "@lib:rcutils", strerror_r_mock);
```
can do.

test/test_strerror.cpp
hidmic: @Blast545 we need a solution for this. Playing with system library defines is brittle and non-portable. 

Digging a bit, it turns out that it's `libstdc++` that **requires** `_GNU_SOURCE` to be set (see [here](https://gcc.gnu.org/onlinedocs/libstdc++/faq.html#faq.predefined)). It's admittedly a bad workaround. We could set `_GNU_SOURCE` when building `rcutils` on Linux. @clalancette @dirk-thomas @wjwwood for  feedback.

test/test_strerror.cpp
Blast545: You mean that we could define another macro similar to RCUTILS_STRINGIFY? I don't understand this suggestion

hidmic: I mean to let the compiler find out what `strerror_r` actual symbol is. If `strerror_r` is a macro that expands to something else, that'll get resolved before `RCUTILS_STRINGIFY` turns it into a string literal. Then we rely on C to concatenate string literals (in C, "foo" " bar" is the same as writing "foobar") and _voilà_! We get to write it in a portable way.

Blast545: Mmm, any ideas on how to do that? that would be great, not sure if the compiler is able to find that actual symbol name.

Without anything else, it resolves to "strerror_r@lib:rcutils". We could write a macro that returns the hardcoded generated name in each platform if we find how that name is generated.

test/test_strerror.cpp
wjwwood: After talking to @hidmic about this offline, I offered this opinion:

I think so long as a) we do not require these GNU specific functions (graceful degradation to POSIX or Windows is possible) and b) they (the GNU functions) are available, then it is fine to use them or force it to use them.

I don't have a sense for any intrinsic drawback or advantage to using the GNU ones over the POSIX ones.
Like performance or security benefits, but they could convince me otherwise, if they exist.

hidmic: PTAL https://github.com/ros2/rcutils/pull/267. Enabling GNU extensions has two potential drawbacks: (1) it makes it possible for non-portable C APIs to be used, and (2) it seems GNU folks departed from the POSIX standard for some functions. We can consciously avoid (1), but I have yet to find an exhaustive list of such departures.

hidmic: Ultimately, the problem's in `g++`, but the bug's been standing for 18 years and is tagged as `wontfix`. It also creates some fun booby traps in C++ code I wasn't aware off e.g. you can't use the stl and name a free function `minor` headers, it'll clash with some GNU macro.

clalancette: For the most part, I agree with @wjwwood 's assessment.  My only concern is about booby-traps like @hidmic laid out in the last comment.

There are specific areas I'm worried about:
1.  The cross-platform area.  If you use a non-portable GNU API by accident, will it obviously show up (fail to compile or link) on the other platforms?
1.  The name clash areas.  As @hidmic found out, it looks like the GNU API is "wider" than the POSIX one.  Is it obvious when you run into one of these clashes?  Have we tried running a full CI run to see what happens?

Assuming we are happy with the answers to those questions, then yeah, I'm fine with going ahead with the GNU API on Linux.

hidmic: > The cross-platform area. If you use a non-portable GNU API by accident, will it obviously show up (fail to compile or link) on the other platforms?

In general, yes. But there are [other cases](https://www.gnu.org/prep/standards/html_node/Non_002dGNU-Standards.html), more subtle, that we have to bear in mind.

> The name clash areas. As @hidmic found out, it looks like the GNU API is "wider" than the POSIX one. Is it obvious when you run into one of these clashes? Have we tried running a full CI run to see what happens?

So far, `rcutils` builds and tests succeed locally. That does not guarantee we won't ever come across a name clash (with a macro, other redefinitions will be caught by the compiler). 

---

Nevermind the choice we make here, I'd want to stress the fact that, because of what `g++` does when it finds `glibc` in the system, in general we cannot ignore GNU API peculiarities. A C++ program using `getopt` is potentially broken already if these are not taken into account.

test/test_strerror.cpp
hidmic: Nevermind what it expands too, I think it's more portable to let the compiler figure out what the symbol is instead of hardcoding it.

test/test_strerror.cpp
clalancette: > Nevermind the choice we make here, I'd want to stress the fact that, because of what `g++` does when it finds `glibc` in the system, in general we cannot ignore GNU API peculiarities. A C++ program using `getopt` is potentially broken already if these are not taken into account.

Actually, that brings up another point to consider here.  I don't remember the exact way to do it in cmake/ament, but I think we need to "export" `-D_GNU_SOURCE` to upstream packages so that they also compile properly.  Does that sound right to you?

hidmic: That's probably the safest bet, but I fear we're spreading the disease. Should a downstream package that uses `rcl` be exposed to GNU extensions?

Blast545:
@hidmic I rewrote some of the tests to use the "then_return" functionality, I was worried about the cases that don't return any useful, but modify one of the passed arguments. Seems to be working really well!

test/test_strerror.cpp
```diff
@@ -21,10 +21,11 @@
 #include <dirent.h>
 #endif
 
-#include <string>
-
+#include "rcutils/macros.h"
 #include "rcutils/strerror.h"
 
+#include "./mimick.h"
```
hidmic, line 27: @Blast545 nit: 
```suggestion
#include "mimick.h"
```

test/test_strerror.cpp
hidmic: Uh, turns out it has the feature and @Blast545 found it :tada:

test/test_strerror.cpp
```diff
@@ -54,3 +55,90 @@ TEST(test_strerror, get_error) {
...
+// Function to be called for (Case 3)
+errno_t mocked_windows_strerror(char * buf, rsize_t bufsz, errno_t errnum)
+{
+  (void) errnum;
+  unsigned char index_err = 0;
+
+  while (buf && bufsz--) {
+    buf[index_err] = expected_error_msg[index_err];
+    index_err++;
+  }
```
hidmic, line 87: @Blast545 here and elsewhere, why not using `strncpy` or even `snprintf` (as it appears GNU `strerror_r` returns `Unknown error nnn`)?

test/test_strerror.cpp
```diff
@@ -54,3 +55,90 @@ TEST(test_strerror, get_error) {
...
+  while (buf && buflen--) {
+    buf[index_err] = error_msg[index_err];
+    index_err++;
+  }
+  return buf;
+}
+#endif
+
+/* Mocking test example */
+TEST(test_strerror, test_mock) {
```
hidmic, line 106: @Blast545 I wonder, would it be easier to read if this was split into three separate tests? There's little reuse beyond a few lines.

hidmic:
BTW this PR needs a new name.

test/test_strerror.cpp
Blast545: I don't like it either, but `cpplint` complains if I don't include the folder path for `.h` files.

test/test_strerror.cpp
Blast545: I didn't use `strncpy` or `snprintf` to avoid using the different implementations required for each host platform. I can replace them if separating the tests completely.

As for the returned `Unknown error nnn`, I made the code to assert for the same result string at the end, the code is taking the returned value of the mock.

test/test_strerror.cpp
Blast545: Didn't consider it yesterday, but I like the idea of separating the tests for readability, I'll add separate them.

Blast545:
@hidmic I addressed your suggestions, let me know what you think

test/test_strerror.cpp
```diff
@@ -67,78 +68,79 @@ TEST(test_strerror, get_error) {
...
+
 #else
+mmk_mock_define(strerror_r_mock, int, int, char *, size_t);
+/* Mocking test example */
+TEST(test_strerror, test_mock) {
+  /* Mock the strerror_r function in the current module using
+     the `strerror_r_mock` blueprint. */
+  mmk_mock(RCUTILS_STRINGIFY(strerror_r) "@lib:rcutils", strerror_r_mock);
+  /* Tell the mock to return NULL and set errno to EINVAL
+     whatever the given parameter is. */
```
hidmic, line 133: @Blast545 nit^N: consider using `//` instead of `/*`.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11610/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6746/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9490/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11571/)

Blast545:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11608/)

Blast545:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9516/)

Blast545:
Re triggering CI after change in `mimick_vendor`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11660/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6784/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9535/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11632/)

Blast545:

test/test_strerror.cpp
hidmic: I see. This is wrong, and it's because of `mimick` includes' layout. See https://github.com/ros2/Mimick/pull/7.

Blast545:
Tier 1 platforms green, PR job green + 1 reviewer, merging :tada:

PR #267, opened 2020-07-16
Force _GNU_SOURCE if glibc is used.

hidmic:
Precisely what the title says. Connected to #265.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ ament_python_install_package(${PROJECT_NAME})
 if(UNIX)
   include(cmake/rcutils_check_c_compiler_uses_glibc.cmake)
   rcutils_check_c_compiler_uses_glibc(USES_GLIBC)
-  if (USES_GLIBC)
+  if(USES_GLIBC)
     # Ensure GNU extended libc API is used, as C++ test code will.
     # See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=2082.
     add_definitions(-D_GNU_SOURCE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ find_package(ament_cmake_ros REQUIRED)
 
 ament_python_install_package(${PROJECT_NAME})
 
-if(UNIX)
+if(UNIX AND NOT APPLE)
   include(cmake/check_c_compiler_uses_glibc.cmake)
   check_c_compiler_uses_glibc(USES_GLIBC)
   if(USES_GLIBC)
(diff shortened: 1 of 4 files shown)


src/process.c
```diff
@@ -17,7 +17,9 @@ extern "C"
 {
 #endif
 
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
```
clalancette, line 22: This concerns me.  In particular, I'm worried that we now may start using incompatible symbols *within* rcutils, if we forget to define this everywhere it is needed.  If we are switching to the GNU API, then I'd suggest we just go all the way and do `add_compile_definitions(-D_GNU_SOURCE)` unconditionally for all files in here.  We may also consider the same for other C libraries, though we can deal with them on a case-by-case basis.  Thoughts?

clalancette: Sorry, I totally misread this patch.  Above you are doing exactly what I expect; `add_definitions(-D _GNU_SOURCE)`.  With that being the case, I would say we can just remove this `#ifndef` block completely.  It will even be "more correct" without this patch on platforms other than Linux, as we'll now only set `_GNU_SOURCE` on Linux.

CMakeLists.txt
```diff
@@ -18,6 +18,16 @@ find_package(ament_cmake_ros REQUIRED)
 
 ament_python_install_package(${PROJECT_NAME})
 
+if(UNIX)
+  include(cmake/rcutils_check_c_compiler_uses_glibc.cmake)
+  rcutils_check_c_compiler_uses_glibc(USES_GLIBC)
```
clalancette, line 23: One minor suggestion here: rename this to `check_c_compiler_uses_glibc()` (and the filename too).  I suspect that we'll be moving this bit of CMake functionality to a more generic place in the near future, so we may as well make it so we don't have to "break the API" here.

hidmic: I shied away from using generic naming, but I like it the most.

hidmic:
Running CI up to `rclcpp` and `rclpy`:

* Linux using `gcc` [Build Status](http://ci.ros2.org/job/ci_linux/11533/)
* Linux using `clang` [Build Status](http://ci.ros2.org/job/ci_linux/11534/)

hidmic:
Ok, all's green. @clalancette @wjwwood anything else?

clalancette:
Looks good to me.

hidmic:
Ok, cool. Going in!

allsey87:
This change broke compatibility with [musl libc](https://www.musl-libc.org/). I am seeing that `_GNU_SOURCE` is not enabled in `errno.h` which means that `program_invocation_name` is never defined.

PR #277, opened 2020-08-21
Add fault injection hooks to default allocator

brawner:
This adds the fault injection hooks to the default allocator. This will enable similar tests to the `time_bomb_allocator` but will inject failures into the three allocating functions (not deallocate) whereas the `time_bomb_allocator` allows for testing each one individually (as well as deallocate).

I'm leaving it out of deallocate so that memory can be freed as well as possible during fault_injection tests. Likewise, unless there is a bug in the program, I don't think `free()` can really fail.

### DIFF
--- a/test/test_allocator.cpp
+++ b/test/test_allocator.cpp
@@ -16,8 +16,10 @@
 
 #include "./allocator_testing_utils.h"
 #include "rcutils/allocator.h"
+#include "rcutils/testing/fault_injection.h"
 
 #include "osrf_testing_tools_cpp/memory_tools/memory_tools.hpp"
+#include "osrf_testing_tools_cpp/scope_exit.hpp"
 
 #ifdef RMW_IMPLEMENTATION
 # define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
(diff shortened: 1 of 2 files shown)


test/test_allocator.cpp
```diff
@@ -110,3 +112,29 @@ TEST(test_allocator, realloc_failing_allocators) {
   auto failing_allocator = get_failing_allocator();
   EXPECT_EQ(nullptr, rcutils_reallocf(allocated_memory, 1024, &failing_allocator));
 }
+
+TEST(test_allocator, default_allocator_maybe_fail) {
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+
+  // Check allocate
+  rcutils_fault_injection_set_count(0);
```
Blast545, line 120: Could this be: 
```suggestion
  rcutils_fault_injection_set_count(RCUTILS_FAULT_INJECTION_FAIL_NOW);
```

test/test_allocator.cpp
```diff
@@ -110,3 +112,29 @@ TEST(test_allocator, realloc_failing_allocators) {
...
+  EXPECT_EQ(nullptr, allocator.allocate(1u, allocator.state));
+  EXPECT_EQ(RCUTILS_FAULT_INJECTION_NEVER_FAIL, rcutils_fault_injection_get_count());
```
Blast545, line 122: The usage of `RCUTILS_FAULT_INJECTION_NEVER_FAIL` is unclear to me in this scenario, `rcutils_fault_injection_get_count` is not supposed to return `0` once the code injected failed? and staying with that value until reset?

brawner: It decrements one more time after hitting 0 to -1 and then stays there.

test/test_allocator.cpp
brawner: They are equivalent, so if you think the suggestion works better then I can go with it.  I think I just went with 0 in the odd event that this test has to be updated and someone needs to add a `set_count(1)` (or change the current case to `set_count(1)`) to handle another case.

test/test_allocator.cpp
Blast545: It's ok then, feel free to mark this comment as resolved

test/test_allocator.cpp
Blast545: In that case, it will be better to have a complete `RCUTILS_FAULT_INJECTION_TEST` test, right?

I thought it may look better with the macro, so it is clear that 0 means to fail now. Feel free to leave as you think is more convenient.

brawner: > In that case, it will be better to have a complete RCUTILS_FAULT_INJECTION_TEST test, right?

`RCUTILS_FAULT_INJECTION_TEST` always runs one more time than necessary in order to check the nominal case. If I used `RCUTILS_FAULT_INJECTION_TEST`, then the unit test would just have to get more complicated I think. I updated it to use `FAIL_NOW` as I think if code changes and this unit test fails, then updating this unit test should require a bit more review.

That all being said, I doubt default allocator will change often so this unit test should be pretty stable.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12052/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7075/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11985/)

brawner:
Good call @Blast545!

Issue #279, opened 2020-08-23
Case insensitive string compare function.


fujitatomoya:
There is no standard case insensitive string compare in C, requesting to introduce `rcutils_strcasecmp`. see https://github.com/ros2/rcl/pull/741#discussion_r474432820.

This is needed to support https://github.com/ros2/rcl/pull/741.

PR #280, opened 2020-08-23
introduce rcutils_strcasecmp, case insensitive string compare.

fujitatomoya:
close https://github.com/ros2/rcutils/issues/279

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>


fujitatomoya:
@iuhilnehc-ynos @Ada-King
CC: @clalancette 

either of you, could you take a look at this? it seems that https://github.com/ros2/rcl/pull/741 does not require this anymore, but i guess that this would be useful.

CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set(rcutils_sources
   src/repl_str.c
   src/shared_library.c
   src/snprintf.c
+  src/strcasecmp.c
```
iuhilnehc-ynos, line 67: What about moving it after `src/split.c`?

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef RCUTILS__STRCASECMP_H_
+#define RCUTILS__STRCASECMP_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <string.h>
```
iuhilnehc-ynos, line 23: Is it good to put this line here? I'd like to move it to '.c'

src/strcasecmp.c
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
```
iuhilnehc-ynos, line 18: I think we don't expect c++ compiler to compile it. ( I know some files also added these lines, but others did not. )

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,49 @@
...
+ * \param[in] s2 String with null terminated to compare.
+ * \return An integer less than, equal to, or greater than zero if s1 is, after
+ *   ignoring case, found to be less than, to match, or be greater than s2,
+ *   respectively.
+ */
+RCUTILS_PUBLIC
+RCUTILS_WARN_UNUSED
+int
+rcutils_strcasecmp(const char * s1, const char * s2);
+
```
iuhilnehc-ynos, line 44: I think it's better to add `strncasecmp(const char *s1, const char * s2, size_t n)` more.

test/test_strcasecmp.cpp
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "rcutils/strcasecmp.h"
+
+// Tests the rcutils_strcasecmp() function.
+TEST(TestStrcasecmp, test_strcasecmp) {
+  EXPECT_EQ(0, rcutils_strcasecmp("", ""));
```
iuhilnehc-ynos, line 21: Do we need to add some test cases if arguments are NULL.

iuhilnehc-ynos (commented):
@fujitatomoya 

nitpick, just some minor comments.

src/strcasecmp.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include "rcutils/strcasecmp.h"
+
+int
+rcutils_strcasecmp(const char * s1, const char * s2)
+{
```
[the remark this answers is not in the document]
iuhilnehc-ynos: And we can follow the style by
https://github.com/ros2/rcutils/blob/a287e74c700deb5816946349fa35aae31b1476f6/src/snprintf.c#L42-L47

include/rcutils/strcasecmp.h
fujitatomoya: right, `.c` should be the right place.

src/strcasecmp.c
fujitatomoya: that is what i though too, but I'd like to keep it just in case.

include/rcutils/strcasecmp.h
fujitatomoya: sgtm, will add `rcutils_strncasecmp` based on `strncasecmp` and `strnicmp`

test/test_strcasecmp.cpp
fujitatomoya: will do

src/strcasecmp.c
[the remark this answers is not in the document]
fujitatomoya: yes, this is needed. will patch as you suggest.

fujitatomoya:
@iuhilnehc-ynos @Ada-King 

comments are addressed, could you check when you got time?

src/strcasecmp.c
```diff
@@ -0,0 +1,59 @@
...
+  RCUTILS_CAN_FAIL_WITH({errno = EINVAL; return -1;});
+
+  if (s1 == NULL || s2 == NULL) {
+    errno = EINVAL;
+    return -1;
```
iuhilnehc-ynos, line 32: I think we need to comment `errno` for this API, because the result '-1' might be a case that "s1" is less than "s2".

- The test case below expecting '-1' might be not good.

- About the test case, do you think we need to expect the return value in the cases is (-1, 0, 1), rather than 'NE 0'.

iuhilnehc-ynos: Or maybe using LT GT is better.

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This function compares two strings ignoring case which just wraps strcasecmp()
+ * or stricmp() in a portable way. This performs a byte-by-byte comparison of the
+ * strings s1 and s2, ignoring the case of the characters.
+ *
+ * \param[in] s1 String with null terminated to compare.
+ * \param[in] s2 String with null terminated to compare.
+ * \return An integer less than, equal to, or greater than zero if s1 is, after
+ *   ignoring case, found to be less than, to match, or be greater than s2,
+ *   respectively.
```
clalancette, line 36: I'll suggest we get rid of the discussion of the implementation.  It's a detail that the user of the API doesn't need to know.  How about:


```suggestion
 * This function compares two strings ignoring case in a portable way.
 * This performs a byte-by-byte comparison of the strings s1 and s2, ignoring the case of the characters.
 *
 * \param[in] s1 Null terminated string to compare.
 * \param[in] s2 Null terminated string to compare.
 * \return An integer less than, equal to, or greater than zero if s1 is, after
 *   ignoring case, found to be less than, to match, or be greater than s2,
 *   respectively.
```

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,66 @@
...
+ * or strnicmp() in a portable way. This performs a byte-by-byte comparison of the
+ * strings s1 and s2 up to count characters of s1 and s2, ignoring the case of the
+ * characters.
+ *
+ * \param[in] s1 String with null terminated to compare.
+ * \param[in] s2 String with null terminated to compare.
+ * \param[in] n Count characters to compare.
+ * \return An integer less than, equal to, or greater than zero if s1 is, after
+ *   ignoring case, found to be less than, to match, or be greater than s2,
+ *   respectively.
```
clalancette, line 55:
```suggestion
/// Case insensitive string compare up to count characters.
/**
 * This function compares two strings ignoring case in a portable way.
 * This performs a byte-by-byte comparison of the strings s1 and s2 up to count
 * characters of s1 and s2, ignoring the case of the
 * characters.
 *
 * \param[in] s1 First string to compare.
 * \param[in] s2 Second string to compare.
 * \param[in] n Count of characters to compare.
 * \return An integer less than, equal to, or greater than zero if s1 is, after
 *   ignoring case, found to be less than, to match, or be greater than s2,
 *   respectively.
```

It may also be worthwhile to talk about what happens if s1 is shorter than s2 and vice-versa.

src/strcasecmp.c
clalancette: Yeah, this is definitely a problem.  While there are some APIs that rely on errno for this (like `strtod`), it is hard to use them properly.  My suggestion here is to change it so that the return value is either 0 for success or -1 for failure.  Then we add in a pointer out argument which is the comparison value.  That way it is easy for the caller to tell whether it failed or not, and we disambiguate.

test/test_strcasecmp.cpp
```diff
@@ -0,0 +1,110 @@
...
+  EXPECT_NE(0, rcutils_strcasecmp("abc1", "abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("abc", "abc1"));
+  EXPECT_NE(0, rcutils_strcasecmp("ABCd", "abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("abc", "ABCd"));
+  EXPECT_NE(0, rcutils_strcasecmp("1Abc", "abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("abc", "1Abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("a1Bc", "abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("abc", "a1Bc"));
+  EXPECT_NE(0, rcutils_strcasecmp("ab1C", "abc"));
+  EXPECT_NE(0, rcutils_strcasecmp("abc", "ab1C"));
```
clalancette, line 61: These tests should be more specific; instead of just `NE`, they should check an exact value:

```
EXPECT_EQ(-1, rcutils_strcasecmp("a", "b");
```

Or similar.

src/strcasecmp.c
fujitatomoya: @clalancette 

that makes sense to me.

fujitatomoya:
@clalancette @iuhilnehc-ynos @Ada-King 

could you do review again?

iuhilnehc-ynos (approved):

Ada-King (approved):

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,76 @@
...
+
+/// Case insensitive string compare.
+/**
+ * This function compares two strings ignoring case in a portable way.
+ * This performs a byte-by-byte comparison of the strings s1 and s2,
+ * ignoring the case of the characters.
+ *
+ * \param[in] s1 Null terminated string to compare.
+ * \param[in] s2 Null terminated string to compare.
+ * \param[out] value Pointer to camparison result.
```
clalancette, line 34:
```suggestion
 * \param[out] value Pointer to comparison result.
```

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,76 @@
...
+ *   An integer less than, equal to, or greater than zero if s1 is, after
+ *   ignoring case, found to be less than, to match, or be greater than s2,
+ *   respectively.
+ * \return 0 if s1 and s2 match, -1 if don't.
```
clalancette, line 38: I don't think this is what we want.  The result of the comparison is already stored in `*value`; this should just be a return of 0 if the method succeeded and -1 if it failed.

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,76 @@
...
+/// Case insensitive string compare up to count characters.
+/**
+ * This function compares two strings ignoring case in a portable way.
+ * This performs a byte-by-byte comparison of the strings s1 and s2 up to count
+ * characters of s1 and s2, ignoring the case of the characters.
+ *
+ * \param[in] s1 First string to compare.
+ * \param[in] s2 Second string to compare.
+ * \param[in] n Count of characters to compare.
+ * \param[out] value Pointer to camparison result.
```
clalancette, line 57:
```suggestion
 * \param[out] value Pointer to comparison result.
```

include/rcutils/strcasecmp.h
```diff
@@ -0,0 +1,76 @@
...
+ *   An integer less than, equal to, or greater than zero if s1 is, after
+ *   ignoring case, found to be less than, to match, or be greater than s2,
+ *   respectively.
+ * \return 0 if s1 and s2 match, -1 if don't.
```
clalancette, line 61: I don't think this is what we want. The result of the comparison is already stored in *value; this should just be a return of 0 if the method succeeded and -1 if it failed.

src/strcasecmp.c
```diff
@@ -0,0 +1,70 @@
...
+#ifndef _WIN32
+  *value = strcasecmp(s1, s2);
+#else
+  *value = stricmp(s1, s2);
+#endif
+  if (*value == 0) {
+    return 0;
+  } else {
+    return -1;
+  }
```
clalancette, line 43:
```suggestion
  return 0;
```

src/strcasecmp.c
```diff
@@ -0,0 +1,70 @@
...
+#ifndef _WIN32
+  *value = strncasecmp(s1, s2, n);
+#else
+  *value = strnicmp(s1, s2, n);
+#endif
+  if (*value == 0) {
+    return 0;
+  } else {
+    return -1;
+  }
```
clalancette, line 65:
```suggestion
  return 0;
```

include/rcutils/strcasecmp.h
fujitatomoya: my bad. thanks

include/rcutils/strcasecmp.h
fujitatomoya: actually i was wondering that. At the moment i was thinking that maybe it is simpler to return 0 if strings are the same ignoring cases. (it is likely to use this way, i guess) but i understand that this changes behavior of `strcasecmp` or `stricmp`. After all, taking your suggestion.

fujitatomoya:
@clalancette 

thanks for the suggestion, i changed it back along with `strcasecmp` behavior. could you check when you got time?

clalancette (approved):

clalancette:
CI (only up to rcutils since this is a new API):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12206/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7189/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9932/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12095/)

clalancette:
@fujitatomoya It looks like Windows has some warnings with this change.  Can you look into it?  Thanks.

fujitatomoya:
@clalancette 

got it, i will address them in today. will get back to you.

fujitatomoya:
@clalancette 

i believe windows is now comfortable with https://github.com/ros2/rcutils/pull/280/commits/9b586f246c59a111ac238ec664babec9506fb7ff. Could you run CI again? thank!

clalancette:
Thanks.  Here's another CI run:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12220/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7203/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9946/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12115/)

clalancette:
Looks good now, thanks for iterating @fujitatomoya !

Issue #278, opened 2020-08-21
Rationale about what to test for performance


Blast545:
Took some time to check the `rcutils` submodules, here's a list of them and what I consider that should be tested for each of those in terms of performance:

├── `allocator.h`
I/O, allocation, runtime critical
 * Allocation time
 * Memory leaks
 * Asymptotic complexity

├── `cmdline_parser.h`
STABLE, not often used
* Not required

├── `env.h`
STABLE, I/O
* Not required, maybe check platform runtime differences

├── `error_handling.h`
STABLE, I/O, Allocation, runtime critical
 * Allocation time
 * Memory leaks
 * Asymptotic complexity

├── `filesystem.h`
STABLE, I/O
* Not required, maybe check platform runtime differences

├── `find.h`
STABLE, Algorithm
* Asymptotic complexity

├── `format_string.h`
Allocation, Critical runtime
* IMO This could be tested through logging or a higher level module

├── `get_env.h`
STABLE, I/O, conf code
* Not required

├── `isalnum_no_locale.h`
STABLE
* Not required

├── `logging.h`
Allocation, Critical runtime
 * Allocation time
 * Memory leaks
 * Asymptotic complexity

├── `process.h`
STABLE, IO, Allocation, not often used
* Not required

├── `qsort.h`
Algorithm
* Asymptotic complexity

├── `repl_str.h`
STABLE, Algorithm
* Asymptotic complexity

├── `shared_library.h`
IO, Conf code or critical runtime (?
 * Allocation time
 * Memory leaks
 * Asymptotic complexity

├── `snprintf.h`
STABLE, IO
* IMO This could be tested through logging or a higher level module. Or simple tests to check platform differences

├── `split.h`
STABLE, Algorithm
 * Asymptotic complexity

├── `stdatomic_helper.h`
STABLE, IO, Critical runtime
* Can those be tested for performance?

├── `strdup.h`
Allocation
 * Memory leaks

├── `strerror.h`
STABLE, IO, Critical runtime
 * Allocation time
 * Memory leaks
 * Asymptotic complexity

├── `time.h`
IO
* Runtime platform differences, maybe test through higher level modules

Custom defined primitives:
│   ├── `array_list.h`
│   ├── `char_array.h`
│   ├── `hash_map.h`
│   ├── `string_array.h`
│   ├── `string_map.h`
│   └── `uint8_array.h`
* Test basic operations runtime performance + Asymptotic complexity
* Memory leaks

[...]

PR #288, opened 2020-09-01
Performance tests demo

Blast545:
Testing performance test usage for `rcutils/error_handling.h` and `rcutils/logging.h`. 

IMO it would be a good idea to combine coverage measurements to see which parts of the code are run with the performance tests, investigation pending.


fujitatomoya (commented):
It would be nice to integrate performance benchmark test. but for doing that, we should be able to have some criteria or threshold to decide if performance is not good enough? I think that most likely this performance check is gonna be useful to make sure if the staging PR does not affect the performance aspect. Could you share your thoughts about this kind of plan?

Blast545:
@fujitatomoya There is a pending discussion on #278 for this package, we will be discussing what's reasonable to test for each package in terms of performance and include checks on CI to be able of checking when a PR affects considerably performance to be considered before merging.

CMakeLists.txt
```diff
@@ -145,6 +145,12 @@ if(BUILD_TESTING)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
   endif()
 
+  # BENCHMARK_F seems to break cppcheck 2.1
+  # set(ament_cmake_cppcheck_FOUND TRUE)
```
wjwwood, line 149: Is this just disabling cppcheck? I don't think that's an ok solution. If it is only disabling it for the tests somehow that would be ok I think. You can also disable cppcheck locally, e.g.:

https://github.com/ros2/rclcpp/blob/c214ddc57b1f2e75076aa087e77c9a06b769008d/rclcpp/include/rclcpp/node.hpp#L195

wjwwood (approved):

CMakeLists.txt
```diff
@@ -505,6 +511,15 @@ if(BUILD_TESTING)
...
     test/test_macros.cpp
   )
+
+add_performance_test(benchmark_test_logging test/test_benchmark_logging.cpp)
+target_link_libraries(benchmark_test_logging ${PROJECT_NAME})
+target_include_directories(benchmark_test_logging PUBLIC include)
+
+add_performance_test(benchmark_test_err_handle test/test_benchmark_error_handling.cpp)
+target_link_libraries(benchmark_test_err_handle ${PROJECT_NAME})
+target_include_directories(benchmark_test_err_handle PUBLIC include)
```
ahcorde, line 521:
```suggestion
  add_performance_test(benchmark_test_logging test/test_benchmark_logging.cpp)
  target_link_libraries(benchmark_test_logging ${PROJECT_NAME})
  target_include_directories(benchmark_test_logging PUBLIC include)

  add_performance_test(benchmark_test_err_handle test/test_benchmark_error_handling.cpp)
  target_link_libraries(benchmark_test_err_handle ${PROJECT_NAME})
  target_include_directories(benchmark_test_err_handle PUBLIC include)
```

test/test_benchmark_logging.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+  const rcutils_log_location_t * location;
+  int level;
+  std::string name;
+  rcutils_time_point_value_t timestamp;
+  std::string message;
+};
+LogEvent g_last_log_event;
+
+static void simple_benchmark(benchmark::State & state) {
```
ahcorde, line 43: rename test

```suggestion
static void benchmark_logging(benchmark::State & state) {
```

test/test_benchmark_logging.cpp
```diff
@@ -0,0 +1,88 @@
...
+    rcutils_log(NULL, RCUTILS_LOG_SEVERITY_ERROR, "", "%s", "");
+    rcutils_log(NULL, RCUTILS_LOG_SEVERITY_FATAL, NULL, "%s", "");
+
+    // restore original state
+    rcutils_logging_set_default_logger_level(original_level);
+    rcutils_logging_set_output_handler(original_function);
+  }
+}
+
+BENCHMARK(simple_benchmark);
```
ahcorde, line 88:
```suggestion
BENCHMARK(benchmark_logging);
```

ahcorde (dismissed):
Does it make sense to add a new folder for the benchmark tests?

test/test_benchmark_logging.cpp
Blast545: See 99818c0

test/test_benchmark_logging.cpp
Blast545: See 99818c0

CMakeLists.txt
Blast545: See 87e22e0

CMakeLists.txt
Blast545: This was to disable cppcheck, I copied it from the performance fixture test demo. Although I enabled it locally and it seems to be working fine. I'll run CI and check if `cppcheck` causes problems with this PR.

Blast545:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12307/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7278/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10022/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12204/)

ahcorde (commented):
I have two suggestions: 
 - call the target `benchmark_logging` instead of `benchmark_test_logging`. I think is redundant somehow. 
 - remove the word `test` from the filename.  (test_benchmark_logging.cpp -> benchmark_logging.cpp)

Blast545:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12376/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7345/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10088/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12288/)

clalancette:
@Blast545 What's the status of this one?  Should we merge it?

ahcorde:
@clalancette I believe that rhe idea it's to create end-to-end performance tests instead of microbenchmarks.

Maybe @chapulina can add more context here.

chapulina:
Yes, the new plan is to create end-to-end tests. But it looks like this PR is already approved and close to getting in, so it would be a shame to throw it away. What's missing to merge it?

clalancette:
> Yes, the new plan is to create end-to-end tests. But it looks like this PR is already approved and close to getting in, so it would be a shame to throw it away. What's missing to merge it?

Yeah, agreed.  If this is almost ready to go, then I would say we get this in, but don't build any *new* micro-benchmarks.

CMakeLists.txt
```diff
@@ -145,6 +145,9 @@ if(BUILD_TESTING)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
   endif()
 
+  find_package(ament_cmake_google_benchmark REQUIRED)
```
cottsay, line 148: You should be able to drop this, as well as the reference in the `package.xml`. The `performance_test_fixture` dependency should be sufficient for what you're doing here.

CMakeLists.txt
```diff
@@ -505,6 +508,15 @@ if(BUILD_TESTING)
   rcutils_custom_add_gtest(test_macros
     test/test_macros.cpp
   )
+
+  add_performance_test(benchmark_logging test/benchmark/benchmark_logging.cpp)
+  target_link_libraries(benchmark_logging ${PROJECT_NAME})
+  target_include_directories(benchmark_logging PUBLIC include)
```
cottsay, line 514: These should get the same `if(TARGET ...)` treatment that the gtests are getting above. If `benchmark` isn't available, the tests won't be built, and these lines will fail (the same as gtest).

cottsay (changes requested):

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13427/)

CMakeLists.txt
Blast545: Sure, see b830dba

CMakeLists.txt
Blast545: Thanks for the advice, addressed with b830dba

clalancette:
Here is full CI on this to make sure everything is happy.  If it is, I'll go ahead and merge it:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13893/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8744/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11602/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13984/)

clalancette (approved):

Issue #285, opened 2020-08-27
Variable name in macro shadows other variable


felixendres:
Hi,

The logging code [here](https://github.com/ros2/rcutils/blob/master/include/rcutils/logging.h#L520-L521) triggers a lot of warnings on building. It is because a macro uses the very simple variable name _ret_, which is also used by the rclcpp function _borrow_serialized_message_ (on my system in /opt/ros/dashing/include/rclcpp/message_memory_strategy.hpp L103) which uses that macro.

I would suggest to skip the assignment to the variable, as it is only used once. Alternatively, use a more unique name that is not likely to be used outside of the macro.

The full warning in question: 

```
/opt/ros/dashing/include/rclcpp/message_memory_strategy.hpp:100:16: warning: declaration of ‘ret’ shadows a previous local [-Wshadow]
           auto ret = rmw_serialized_message_fini(msg);
                ^~~
/opt/ros/dashing/include/rclcpp/message_memory_strategy.hpp:93:10: note: shadowed declaration is here
     auto ret = rmw_serialized_message_init(msg, capacity, &rcutils_allocator_);
          ^~~
In file included from /opt/ros/dashing/include/rmw/types.h:28:0,
                 from /opt/ros/dashing/include/rcl/types.h:18,
                 from /opt/ros/dashing/include/rcl/arguments.h:20,
                 from /opt/ros/dashing/include/rcl/context.h:26,
                 from /opt/ros/dashing/include/rcl/guard_condition.h:24,
                 from /opt/ros/dashing/include/rclcpp/executor.hpp:29,
                 from /opt/ros/dashing/include/rclcpp/executors/multi_threaded_executor.hpp:24,
                 from /opt/ros/dashing/include/rclcpp/executors.hpp:21,
                 from /opt/ros/dashing/include/rclcpp/rclcpp.hpp:144,
                 from ../../../../../xxx.h:12,
                 from ../../../../../xxx.cpp:5:
/opt/ros/dashing/include/rcutils/logging.h:543:19: warning: declaration of ‘ret’ shadows a previous local [-Wshadow]
     rcutils_ret_t ret = rcutils_logging_initialize(); \
                   ^
/opt/ros/dashing/include/rcutils/logging_macros.h:68:5: note: in expansion of macro ‘RCUTILS_LOGGING_AUTOINIT’
     RCUTILS_LOGGING_AUTOINIT \
     ^~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #290, opened 2020-09-11
Update logging.h

felixendres:
Fix issue #285:
Remove the temporary variable in a macro, because the name of the variable may shadow a variable of the macro user.
As suggested by @clalancette, also wrapped it into a do-while wrapping.

Signed-off-by: Felix Endres

### DIFF
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -516,17 +516,18 @@ void rcutils_logging_console_output_handler(
  * All logging macros ensure that this has been called once.
  */
 #define RCUTILS_LOGGING_AUTOINIT \
-  if (RCUTILS_UNLIKELY(!g_rcutils_logging_initialized)) { \
-    rcutils_ret_t ret = rcutils_logging_initialize(); \
-    if (ret != RCUTILS_RET_OK) { \
-      RCUTILS_SAFE_FWRITE_TO_STDERR( \
-        "[rcutils|" __FILE__ ":" RCUTILS_STRINGIFY(__LINE__) \
-        "] error initializing logging: "); \
-      RCUTILS_SAFE_FWRITE_TO_STDERR(rcutils_get_error_string().str); \
-      RCUTILS_SAFE_FWRITE_TO_STDERR("\n"); \
-      rcutils_reset_error(); \
+  do { \
+    if (RCUTILS_UNLIKELY(!g_rcutils_logging_initialized)) { \
+      if (rcutils_logging_initialize() != RCUTILS_RET_OK) { \
+        RCUTILS_SAFE_FWRITE_TO_STDERR( \
+          "[rcutils|" __FILE__ ":" RCUTILS_STRINGIFY(__LINE__) \
+          "] error initializing logging: "); \
+        RCUTILS_SAFE_FWRITE_TO_STDERR(rcutils_get_error_string().str); \
+        RCUTILS_SAFE_FWRITE_TO_STDERR("\n"); \
+        rcutils_reset_error(); \
+      } \
     } \
-  }
+  } while (0);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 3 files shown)


include/rcutils/logging.h
```diff
@@ -516,17 +516,18 @@ void rcutils_logging_console_output_handler(
...
+        RCUTILS_SAFE_FWRITE_TO_STDERR( \
+          "[rcutils|" __FILE__ ":" RCUTILS_STRINGIFY(__LINE__) \
+          "] error initializing logging: "); \
+        RCUTILS_SAFE_FWRITE_TO_STDERR(rcutils_get_error_string().str); \
+        RCUTILS_SAFE_FWRITE_TO_STDERR("\n"); \
+        rcutils_reset_error(); \
+      } \
     } \
-  }
+  } while (0);
```
clalancette, line 530: This looks pretty good to me.  One thing that I'd prefer to do here is to remove the semicolon from the end of this line, and make the callers of this macro have the semicolon.  That will play more nicely with text editors, and also look more like statements inline.

To do this, we'd have to update all of the callers of `RCUTILS_LOGGING_AUTOINIT` to have a semicolon on the end.  Most of the callers are in this repository, but there is also a caller in [rcl](https://github.com/ros2/rcl/blob/f3396a2a7dcf061d92642a8dd2af0b1673d36057/rcl/src/rcl/logging.c#L65) and [rclpy](https://github.com/ros2/rclpy/blob/a37c01511324c1e3e57bc47a7e1286bf79514c60/rclpy/src/rclpy/_rclpy_logging.c#L173).  We'd need to have separate PRs for that.

@felixendres Are you willing to make those updates?  If so, please go ahead and fix up rcutils as suggested, and open PRs to rcl and rclpy as well.  If not, that's OK, but please open an issue so we can do this follow-up in the future.

fujitatomoya: I totally agree on @clalancette suggestion, in addition to

- No need to prevent uncrustify from making unnecessary indents.
- It is unlikely that user application calls `RCUTILS_LOGGING_AUTOINIT` directly.

fujitatomoya (changes requested):

felixendres:
I'd prefer to not make the semicolon changes. This issue is actually an extremely minor issue, so let's not sink too much time into it.

Apart from the uncrustify comment (which I don't understand, so you be the judge) the current change improves the status quo. The semicolon issue is the same as before (putting one after the macro is superfluous before and after).

So I suggest to merge it as is, unless that uncrustify thing is a problem.

clalancette:
I've gone ahead and rebased this, and also changed to remove the colon.  Besides the changes in this PR, this required two other PRs, one in rclpy and one in rcl.  Here's CI for the whole lot:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12578/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7546/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10289/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12503/)

clalancette:
All right.  The Windows failures are happening on nightlies as well, so this is ready to go.  Since I made changes here, I'm going to wait for another review from @ros2/team .

dirk-thomas:
@felixendres it would good to update the ticket title to express the content of the patch since it will become the squashed commit message.

ivanpauno (approved):

felixendres:
Sorry, for the title-change back and forth. I could swear I still saw the old title "Update logging.h", when I changed it...

clalancette:
> Sorry, for the title-change back and forth. I could swear I still saw the old title "Update logging.h", when I changed it...

No worries, it was probably just cached in your browser.

PR #201, opened 2020-01-17
increase max length of env var value on Windows to 32767

dirk-thomas:
Trying to use `rcutils_get_env()` in ros2/rmw_implementation#57 fails since that package e.g. tries to get the environment variable `PATH` which exceeds the current limit of 2048 characters on the Jenkins CI nodes.

Therefore this PR raises the limit to the maximum size of the sum of all environment variables (see https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083).

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8999/)

PR #291, opened 2020-09-18
env.h and get_env.h docblock fixes

ivanpauno:
Update docblock after changes in:
* https://github.com/ros2/rcutils/pull/237
* https://github.com/ros2/rcutils/pull/201

### DIFF
--- a/include/rcutils/get_env.h
+++ b/include/rcutils/get_env.h
@@ -47,9 +47,8 @@ extern "C"
  * printf("Valued of 'SOME_ENV_VAR': %s\n", env_value);
  * ```
  *
- * Environment variables will be truncated at 2048 characters on Windows.
- *
- * This function is not thread-safe.
+ * This function cannot be thread-safely called together with rcutils_set_env
+ * (or any platform specific equivalent), but multiple calls to this function are thread safe.
  *
  * \param[in] env_name the name of the environment variable
  * \param[out] env_value pointer to the value cstring, or "" if unset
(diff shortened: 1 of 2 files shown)


include/rcutils/get_env.h
```diff
@@ -47,9 +47,8 @@ extern "C"
  * printf("Valued of 'SOME_ENV_VAR': %s\n", env_value);
  * ```
  *
- * Environment variables will be truncated at 2048 characters on Windows.
- *
- * This function is not thread-safe.
+ * This function cannot be thread-safely called together with rcutils_set_env
+ * (or any platform specific equivalent), but multiple calls to this function are thread safe.
```
clalancette, line 51:
```suggestion
 * This function cannot be concurrently called together with rcutils_set_env (or any platform specific equivalent)
 * on different threads, but multiple concurrent calls to this function are thread safe.
```

Also, while we are in here, can you please update the documentation for `rcutils_get_home_dir`?  It has the same constraints as this call (also, it no longer truncates to 2048 characters).

clalancette (changes requested):

include/rcutils/get_env.h
ivanpauno: > Also, while we are in here, can you please update the documentation for rcutils_get_home_dir? It has the same constraints as this call (also, it no longer truncates to 2048 characters).

Ah I didn't see that one, I will fix it.

ivanpauno: Fixed in https://github.com/ros2/rcutils/pull/291/commits/967e833daa8f62803b42fca01498e8484b464088.

I also corrected a bit `set_env` docs in https://github.com/ros2/rcutils/pull/291/commits/56ce9ef02e859d4ad30789e20b9aefdbc9296620.
`set_env` isn't thread safe, either if you're using `get_env` or not.

clalancette (approved):
Looks good, thanks for iterating.

PR #294, opened 2020-09-30
Move RCUTILS_PATH_DELIMITER to header

christophebedard:
This way it can be used by other packages.

### DIFF
--- a/include/rcutils/filesystem.h
+++ b/include/rcutils/filesystem.h
@@ -27,6 +27,12 @@ extern "C"
 #include "rcutils/macros.h"
 #include "rcutils/visibility_control.h"
 
+#ifdef _WIN32
+# define RCUTILS_PATH_DELIMITER "\\"
+#else
+# define RCUTILS_PATH_DELIMITER "/"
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):

clalancette:
@christophebedard Is this still required?  Sorry, its unclear to me from the context of ros2/rcl_logging#53

christophebedard:
> @christophebedard Is this still required? Sorry, its unclear to me from the context of [ros2/rcl_logging#53](https://github.com/ros2/rcl_logging/pull/53)

I ended up not using it for ros2/rcl_logging#53, but maybe it could still be useful? I'd leave that up to you.

clalancette:
> I ended up not using it for [ros2/rcl_logging#53](https://github.com/ros2/rcl_logging/pull/53), but maybe it could still be useful? I'd leave that up to you.

If we don't need it right now, I'm going to say we shouldn't make the change.  It is simple enough, so we can always do it in the future if need be.

So I'm going to close this out for now.  Thanks!

Issue #304, opened 2020-10-27
Calculate the size of the specified directory (include subdirectory)


Barry-Xu-2018:
rcutils_calculate_directory_size() only support calculating the size of directory without recursive.  
Why rcutils_calculate_directory_size() doesn't support recursive ? 
Is there any cause ?

fujitatomoya:
I don't think of anything but i believe that it would be nice if it works recursively.

`rcutils_calculate_directory_size` is currently used to calculate [rosbag2 metadata](https://github.com/ros2/rosbag2/blob/700817c42f1bbc26ced79aca2186094350c8d1be/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp#L235). since all files in the directory, it does not have to be recursive for right now.

Barry-Xu-2018:
> rcutils_calculate_directory_size is currently used to calculate rosbag2 metadata. since all files in the directory, it does not have to be recursive for right now.

Yes. Now only `metadata_io.cpp` in rosbag2 call this function.

fujitatomoya:
it explicitly describes that it does not support recursive search. do we need it to be recursive with actual use cases?

Barry-Xu-2018:
> it explicitly describes that it does not support recursive search. do we need it to be recursive with actual use cases?

Yes.  Current implementation in rosbag2 as below

https://github.com/ros2/rosbag2/blob/700817c42f1bbc26ced79aca2186094350c8d1be/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp#L229-L240

It only calculates total of file size in specified path. For sqlite3 database (Now rosbag2 only provide this as storage back-end), it is okay since sqlite3 database is one file. But for others (such as leveldb), it use a directory to save data. Above codes cannot get correct size of database.  
I want to update this function to support recursive.  
This modification should not introduce the problem for current codes.

clalancette:
> I want to update this function to support recursive.
> This modification should not introduce the problem for current codes.

Yeah, I don't see a problem with enhancing it to do that.  We may want to introduce a `recursive` boolean parameter to the function in case the caller really doesn't want that, but otherwise it sounds good to me.

fujitatomoya:
@clalancette 
CC: @Barry-Xu-2018 

thanks for the comment!
actually we've been trying to use [leveldb](https://github.com/google/leveldb) as rosbag2 backend database. this enhancement would be necessary to do that.

Barry-Xu-2018:
[...]

PR #306, opened 2020-10-29
Update rcutils_calculate_directory_size() to support recursion

Barry-Xu-2018:


Barry-Xu-2018:
Related to #304

fujitatomoya:
@clalancette @Karsten1987 friendly ping. i will be looking into this today.

test/test_filesystem.cpp
```diff
@@ -419,9 +419,9 @@ TEST_F(TestFilesystemFixture, calculate_directory_size) {
 #ifdef WIN32
```
fujitatomoya, line 419: I would keep the non-recursive test case like following.

- non-recursive test case: rcutils_calculate_directory_size(test/dummy_folder/dummy-subfolder/dummy-subfolder)
- recursive test case: rcutils_calculate_directory_size(test/dummy_folder)

Barry-Xu-2018: This modification is to support recursion.  So not support non-recursion.   
If there is non-recursive test case, do you mean 2 API  --- one for non-recursion and the other for recursion ?

fujitatomoya: my bad, i wasn't clear enough. what i mean is not about the function but test cases,

- test only files in a single directory. (non-recursive search `test` case)
- test directories and files. (recursive search `test` case)

fujitatomoya (approved):
lgtm

include/rcutils/filesystem.h
```diff
@@ -187,7 +187,8 @@ rcutils_mkdir(const char * abs_path);
 /// Calculate the size of the specified directory.
 /*
  * Calculates the size of a directory by summarizing the file size of all files.
- * \note This operation is not recursive.
+ * \note This operation is recursive.
```
iuhilnehc-ynos, line 190: It seems that the recursive operation does not contain any subdirectory symbolic links. Do you think we need to update the note with this kind of message?

fujitatomoya: Nice catch! i do not really see the use cases for plugin creates the symlink. but it should be clearly defined and described in the header.

src/filesystem.c
```diff
@@ -263,7 +263,11 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
       // Skip over local folder handle (`.`) and parent folder (`..`)
       if (strcmp(data.cFileName, ".") != 0 && strcmp(data.cFileName, "..") != 0) {
         char * file_path = rcutils_join_path(directory_path, data.cFileName, allocator);
-        dir_size += rcutils_get_file_size(file_path);
+        if (rcutils_is_directory(file_path)) {
+          dir_size += rcutils_calculate_directory_size(file_path, allocator);
```
clalancette, line 267: I'm somewhat concerned about this being a recursive operation.  If you have a very deeply nested directory tree, this could use up quite a lot of stack space (or even hit the default limit).

One way to get around this would be to make this iterative rather than recursive.  We would still need to allocate memory to store which items to visit next, but then it would be from the heap.

What do people think about my concern?

clalancette (commented):
I still have the overall concern about this, which is whether this change makes a difference to rosbag2.  I'm still waiting to hear back from @Karsten1987 about that.  Other than that, I have one more concern inline.

src/filesystem.c
fujitatomoya: I do agree with you. recursive operation could be concern. We might want to get directory recursive iterators at 1st and call `rcutils_calculate_directory_size` for each. this can avoid calling function recursively.

include/rcutils/filesystem.h
```diff
@@ -187,7 +187,8 @@ rcutils_mkdir(const char * abs_path);
 /// Calculate the size of the specified directory.
 /*
  * Calculates the size of a directory by summarizing the file size of all files.
- * \note This operation is not recursive.
+ * \note This operation is recursive. Not follow the symbolic link that points to directory.
```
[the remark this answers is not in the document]
Barry-Xu-2018: Thanks. I will update description.

src/filesystem.c
Barry-Xu-2018: > I'm somewhat concerned about this being a recursive operation. If you have a very deeply nested directory tree, this could use up quite a lot of stack space (or even hit the default limit).
>
> One way to get around this would be to make this iterative rather than recursive. We would still need to allocate memory to store which items to visit next, but then it would be from the heap.

Completely agree.  

Maybe we can provide another API function to provide recursive operation support. 
rcutils_calculate_directory_size() API isn't changed.   
Add new API rcutils_calculate_directory_size_with_recursion().  
Real implementation are moved to rcutils_calculate_directory_size_impl().
```c
rcutils_calculate_directory_size(...) {
  rcutils_calculate_directory_size_impl(..., is_recursive = 0);
}

rcutils_calculate_directory_size_with_recursion (...) {
  rcutils_calculate_directory_size_impl(..., is_recursive = 1);
}
```

What do you think ? 

About whether rcutils_calculate_directory_size_with_recursion() is used to replace rcutils_calculate_directory_size() in rosbag2, I will discuss with @Karsten1987.

Barry-Xu-2018: > Recursive operation could be concern. We might want to get directory recursive iterators at 1st and call rcutils_calculate_directory_size for each. this can avoid calling function recursively.

Yes. But it hard to make sure whether the depth of the directory tree need to be added in the future (For other database).   
Maybe we can consider there is a reasonable maximum depth limitation in API function. Such as 30.

fujitatomoya: if i wasn't mistaken, that is not exactly pointed here. I didn't mean the concern is about depth of the tree (size of memory), but location of memory. (not stack but heap) calling the caller function in the callee, this could hit the limit of stack size. So i was thinking that instead of calling the function recursively, get the directory iterators at 1st, then get all file size in `a single` function.

clalancette: > if i wasn't mistaken, that is not exactly pointed here. I didn't mean the concern is about depth of the tree (size of memory), but location of memory. (not stack but heap) calling the caller function in the callee, this could hit the limit of stack size. So i was thinking that instead of calling the function recursively, get the directory iterators at 1st, then get all file size in `a single` function.

Yes, this is exactly the kind of thing I was thinking about.  If we do this iteratively, instead of recursively, then we can both limit the total amount of memory needed and make sure that memory is allocated from the heap rather than from the stack (no matter how deep the structure is).  Something like:

```
std::vector<std::string> directory_names{"/foo"};
int size = 0;
do {
   diriter = get_directory_iter(directory_names[0]);
   for (filename in diriter) {
      if (is_file(filename)) {
        size += filesize(filename);
      } else if (is_dir(filename)) {
        directory_names.push_back(filename);
      }
   }
   directory_names.pop();
} while (directory_names.size() > 0);
```

Separately, we might *also* considering allowing the API user to limit the recursion depth (mostly to limit the time it takes for this operation to complete).

Barry-Xu-2018: @clalancette @fujitatomoya 

Thanks for further explanation.  
I understand.  
I will update codes.

Barry-Xu-2018:
@clalancette @fujitatomoya

I updated codes to implement recursion without consuming memory from stack.   
A new API `rcutils_calculate_directory_size_with_recursion()` is added.  For the original API `rcutils_calculate_directory_size()`, the behavior isn't changed.

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
   return dir_size;
 #else
-  DIR * dir = opendir(directory_path);
-  if (NULL == dir) {
-    RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING(
-      "Can't open directory %s. Error code: %d\n", directory_path, errno);
-    return dir_size;
+  DIR * dir = NULL;
+
+  dir_list->next = NULL;
```
clalancette, line 359: If you use `zero_allocate` above, you don't need this line.

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+        "Can't open directory %s. Error code: %d\n", dir_list->path, errno);
+      goto fail;
     }
+
+    // Scan in specified path
+    // If found directory, add to dir_list
+    // If found file, calculate file size
+    while (NULL != (entry = readdir(dir))) {
+      // Skip over local folder handle (`.`) and parent folder (`..`)
+      if (strcmp(entry->d_name, ".") != 0 && strcmp(entry->d_name, "..") != 0) {
```
clalancette, line 382: I think this loop can be simplified a bit if this does a "continue" if it is `.` or `..`.  Like:

```
while (NULL != (entry = readdir(dir))) {
    if (name == "." || name == "..") {
      continue;
    }
    ...
}
```

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+        if (NULL == file_path) {
+          RCUTILS_SAFE_FWRITE_TO_STDERR("rcutils_join_path return NULL !\n");
+          goto fail;
+        }
+
+        if (rcutils_is_directory(file_path)) {
+          if ((max_deepth == 0) || ((dir_list->depth + 1) <= max_deepth)) {
+            // Add new directory to dir_list
+            dir_list_t * found_new_dir =
+              (dir_list_t *) allocator.allocate(sizeof(dir_list_t), allocator.state);
```
clalancette, line 393:
```suggestion
              allocator.allocate(sizeof(dir_list_t), allocator.state);
```

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+            if (NULL == found_new_dir) {
+              RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory !\n");
```
clalancette, line 395: I think we are leaking `file_path` on error.

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+    handle = FindFirstFile(dir_list->path, &data);
+    if (INVALID_HANDLE_VALUE == handle) {
+      RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING(
+        "Can't open directory %s. Error code: %lu\n", dir_list->path, GetLastError());
+      goto fail;
+    }
+
     do {
       // Skip over local folder handle (`.`) and parent folder (`..`)
       if (strcmp(data.cFileName, ".") != 0 && strcmp(data.cFileName, "..") != 0) {
```
clalancette, line 310: I think the loop would be a little easier to read if we `continue` early on `.` and `..`.  Something like:

```
  do {
    if (cFilename == "." || cFilename == "..") {
      continue;
    }
    ...
  } while (FindNextFile(handle, &data));
```

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+        if (NULL == file_path) {
+          RCUTILS_SAFE_FWRITE_TO_STDERR("rcutils_join_path return NULL !\n");
+          goto fail;
+        }
+
+        if (rcutils_is_directory(file_path)) {
+          if ((max_deepth == 0) || ((dir_list->depth + 1) <= max_deepth)) {
+            // Add new directory to dir_list
+            dir_list_t * found_new_dir =
+              (dir_list_t *) allocator.allocate(sizeof(dir_list_t), allocator.state);
```
clalancette, line 321:
```suggestion
              allocator.allocate(sizeof(dir_list_t), allocator.state);
```

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+            if (NULL == found_new_dir) {
+              RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory !\n");
```
clalancette, line 323: I think we are leaking `file_path` on error.

clalancette (changes requested):
I've got a bunch of stuff to fix in here.

One other thing; the inner loops for both the Windows and Unix sides are very similar.  Do you think it is possible to use a common function for that?

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
+  const uint32_t max_deepth,
+  rcutils_allocator_t allocator)
+{
+  dir_list_t * dir_list = NULL;
+  size_t dir_size = 0;
+
+  dir_list = (dir_list_t *) allocator.allocate(sizeof(dir_list_t), allocator.state);
+  if (NULL == dir_list) {
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to allocate memory !\n");
+    return 0;
```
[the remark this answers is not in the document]
Barry-Xu-2018: I considered this.  For compatibility, I keep return 0 while the problem occurs.   
This is the behavior of  current rcutils_calculate_directory_size() implementation.

In rosbag2, it calls this function as below.   

https://github.com/ros2/rosbag2/blob/4bcb631e674d7c0e3e7c55eb6b4915ff74f9cd8c/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp#L235-L236

As your comment, it is not good. After updating return value, I also need to update above codes.

src/filesystem.c
```diff
@@ -248,45 +248,182 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
...
   }
+
+  return rcutils_calculate_directory_size_with_recursion(directory_path, 1, allocator);
+}
+
+typedef struct dir_list_t
+{
+  char * path;
+  uint32_t depth;  // Compare with base path
+  struct dir_list_t * next;
```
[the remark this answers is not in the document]
Barry-Xu-2018: > I think link list is really simple and faster in this case, but we would want to use rcutils_array_list_t? i believe that this is the case that is for.

I check the implementation of rcutils_array_list_t.  
But I think it is unsuitable for our case.   
After scan one directory, we will remove this directory in directory list. So there are many remove operations in this case.  
The remove operation in rcutils_array_list_t as below:

https://github.com/ros2/rcutils/blob/master/src/array_list.c#L164-L179

It introduces shift cost.

Barry-Xu-2018:
> if we cannot change the rosbag2 code, i think we should stick with current rcutils_calculate_directory_size that means it returns 0 even with error. but as interface, i think current interface is the one which needs to be changed. since user application cannot tell error or actual size is 0. that would lead the another problem.

Agree. So I try to change the interface to return -1 for error.   
Now only one project  (rosbag2) use this rcutils_calculate_directory_size().   
So the sooner we update this interface, the better.

src/filesystem.c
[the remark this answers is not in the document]
fujitatomoya: okay i see. creating and deleting lists via `rcutils_array_list_t` would be unreasonable, thanks.

Barry-Xu-2018:
@clalancette @fujitatomoya 

I updated codes based on your comments.   

> Change the return type from a size_t to an ssize_t. Then we can return -1 to indicate error.

I adopt this way to show error while calling rcutils_calculate_directory_size() or rcutils_calculate_directory_size_with_recursion().

src/filesystem.c
```diff
@@ -238,56 +238,185 @@ rcutils_mkdir(const char * abs_path)
...
+} dir_list_t;
+
+static inline void free_dir_list(dir_list_t * dir_list, rcutils_allocator_t allocator)
```
clalancette, line 254: Let's remove the `inline` here (and below).  It is only a hint to the compiler anyway, and it can figure out whether it wants to inline it or not.

```suggestion
static void free_dir_list(dir_list_t * dir_list, rcutils_allocator_t allocator)
```

include/rcutils/filesystem.h
```diff
@@ -185,19 +192,46 @@ bool
...
+ * If necessary, you can specify the maximum directory depth to recurse into.
+ * Depth definition as below.
+ * \code
+ * directory_path  <= depth 1
+ *    |- subdirectory <= depth 2
+ *            |- subdirectory <= depth 3
+ *                    ...
+ * \endcode
+ *
+ * \note This operation doesn't follow the symbolic link that points to directory.
```
clalancette, line 219: Can you clarify the situation with symlinks a bit, and maybe add tests for it?  In particular:

1.  What happens if there is a symlink to a file?  Does the size of that file get added to the overall returned size?
1.  What happens if there is a symlink to a directory?  Do we recurse into that directory and add the sizes?

include/rcutils/filesystem.h
```diff
@@ -185,19 +192,46 @@ bool
...
+ * \param[in] allocator Allocator being used for internal file path composition.
+ * \return The size of the directory in bytes on success.
+ * \return -1 on failure.
+ */
+RCUTILS_PUBLIC
+ssize_t
+rcutils_calculate_directory_size_with_recursion(
+  const char * directory_path,
+  const size_t max_depth,
+  rcutils_allocator_t allocator);
```
clalancette, line 231: I'm very sorry about this, but I realize I made a mistake suggesting an `ssize_t` return value here.

On 64-bit systems (amd64, Aarch64), this will work fine.  We won't be seeing file sizes greater 2^63 - 1.

But on 32-bit systems (x86, arm32), the maximum size we'll be able to return is 2^31 - 1 .  That's just over 2 GB, and it seems entirely plausible that we'll see bags (and other things) larger than that.  While 32-bit systems aren't Tier 1 support, they are something that we try to keep working.  So my suggestion is that we change this signature to:

```
rcutils_ret_t
rcutils_calculate_directory_size_with_recursion(
  const char * directory_path,
  const size_t max_depth,
  uint64_t * size,
```

What do you think?

clalancette (changes requested):
A few more things to think about in here.  This one is tricky to get right.

include/rcutils/filesystem.h
Barry-Xu-2018: Yeah. This is a problem.   
I will update codes.

Barry-Xu-2018:
@clalancette

Sorry to late update.  
Please help review again.  
For rcutils_calculate_directory_size(), I don't keep the same interface as rcutils_calculate_directory_size_with_recursion().  
It is because we should update code of rosbag once rcutils_calculate_directory_size() interface is changed.  
I cannot make sure 2 repositories have been update at the same time for CI system. (Otherwise, CI system always reports build failure).

include/rcutils/filesystem.h
```diff
@@ -185,19 +192,49 @@ bool
...
+ * \return `RCUTILS_RET_INVALID_ARGUMENT` for invalid arguments, or
+ * \return `RCUTILS_RET_BAD_ALLOC` if memory allocation fails
+ * \return `RCUTILS_RET_ERROR` if other error occurs
+ */
+RCUTILS_PUBLIC
+rcutils_ret_t
+rcutils_calculate_directory_size_with_recursion(
+  const char * directory_path,
+  const size_t max_depth,
+  size_t * size,
```
clalancette, line 233: I don't think this works on 32-bit; as far as I know, this is defined as `unsigned int`, and hence will still be 32-bit there.  I think we should explicitly use `uint64_t` here.

clalancette (changes requested):
One major thing to change.  We are getting closer, thanks for all of the revisions.

include/rcutils/filesystem.h
Barry-Xu-2018: > I don't think this works on 32-bit; as far as I know, this is defined as unsigned int, and hence will still be 32-bit there. I think we should explicitly use uint64_t here.

Yeah.  It's my fault.  
I understand why your suggestion use uint64_t instead of size_t.

Barry-Xu-2018:
@clalancette 

Updated. Please review again.

include/rcutils/filesystem.h
clalancette: The declaration here still needs to be changes to `uint64_t`.

src/filesystem.c
```diff
@@ -241,53 +241,207 @@ rcutils_mkdir(const char * abs_path)
 size_t
 rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_t allocator)
```
clalancette, line 242: I understand that it is hard to make API changes across repositories, but I do think we should do that here.  In point of fact, this API has the same problems as the new `rcutils_calculate_directory_size_with_recursion` had, namely:

1.  You can't tell if the return value is 0 because the call failed, or because there were just no files in the directory.
1.  On 32-bit, you can only return a size up to 4GB, which is pretty small.

So I think we should change this signature to:

```
rcutils_ret_t
rcutils_calculate_directory_size(const char *directory_path, uint64_t *size, rcutils_allocator_t allocator)
```
And also update rosbag2 to use that signature as well.  It won't pass the PR job until we release both of them, but that's OK.  We'll run normal CI (that's why we have it), and then I can merge both and do a release of both at the same time.

include/rcutils/filesystem.h
```diff
@@ -22,6 +22,13 @@ extern "C"
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
+#ifndef _WIN32
+#include <sys/types.h>
+#else
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
```
clalancette, line 31: Since we aren't using `ssize_t` anymore, do we still need this block?

clalancette (changes requested):

include/rcutils/filesystem.h
Barry-Xu-2018: Yeah. Above codes should be removed.

src/filesystem.c
Barry-Xu-2018: Okay. I will update rcutils_calculate_directory_size().  
And update rosbag2' codes since rcutils_calculate_directory_size() is changed.   

>   We'll run normal CI (that's why we have it)

I don't know this `normal CI`. What is the difference with the current system ？

Barry-Xu-2018:
@clalancette 

Codes have been updated. Please help to review.  
Besides, I have created PR (ros2/rosbag2#567) for rosbag2.

src/filesystem.c
clalancette: > I don't know this `normal CI`. What is the difference with the current system ？

Sorry, I should have been more clear.

We actually have two separate types of CI.  When you open a PR (and anytime you update it), the "Rpr" job runs on https://build.ros2.org.  You can see the output of that down below in the "Checks" section of this PR.  The PR job takes all of the current Rolling packages that this package depends on, installs them, checks this branch out, and then does a build-and-test.  This is great as it is automatic, and gives us fairly quick feedback on small issues like compiler warnings, linter warnings, unit tests, etc.  But it does have some problems:

1.  It won't test packages that depend on this package.  So in this case, the PR job succeeded, but if we merge this and then try to build rosbag2, it would fail because of the changed API.
1.  If there are API changes in packages that this package depends on, it will just fail.  There is no way to do a coordinated test across several repositories.
1.  It only tests Linux, not the other Tier-1 platforms.

For all of those reasons, we also have https://ci.ros2.org, where we can test all of the things above.

Therefore, for this PR, once we are happy with the code and it is approved, I can run a full test on ci.ros2.org to get cross-platform tests going, and with all packages that depend on this one.

Does that all make some sense?

include/rcutils/filesystem.h
```diff
@@ -185,19 +186,56 @@ bool
...
+ * If necessary, you can specify the maximum directory depth to recurse into.
+ * Depth definition as below.
+ * \code
+ * directory_path  <= depth 1
+ *    |- subdirectory <= depth 2
+ *            |- subdirectory <= depth 3
+ *                    ...
+ * \endcode
+ *
+ * \note Not calculate the size of the symlink which points to file or directory.
```
clalancette, line 219:
```suggestion
 * \note This API does not follow symlinks to files or directories.
```

include/rcutils/filesystem.h
```diff
@@ -185,19 +186,56 @@ bool
...
+ * \param[in] directory_path The directory path to calculate the size of.
+ * \param[in] max_depth The maximum depth of subdirectory. 0 means no limitation.
+ * \param[out] size The size of the directory in bytes on success
```
clalancette, line 222:
```suggestion
 * \param[out] size The size of the directory in bytes on success.
```

include/rcutils/filesystem.h
```diff
@@ -185,19 +186,56 @@ bool
 rcutils_mkdir(const char * abs_path);
 
 /// Calculate the size of the specified directory.
-/*
+/**
  * Calculates the size of a directory by summarizing the file size of all files.
  * \note This operation is not recursive.
  * \param[in] directory_path The directory path to calculate the size of.
+ * \param[out] size The size of the directory in bytes on success
```
clalancette, line 193:
```suggestion
 * \param[out] size The size of the directory in bytes on success.
```

clalancette (changes requested):
This is looking great.  I've got a few more minor things to fix up inline; once that is done, I'll go ahead and run CI.  Thanks for all of the iterations here.

src/filesystem.c
Barry-Xu-2018: Thank you so much for your detailed explanation.  
I get it.  
I try to login https://ci.ros2.org and I think normal user have no permission to execute it.   
So you will help to execute it. Right ?

src/filesystem.c
```diff
@@ -238,56 +238,210 @@ rcutils_mkdir(const char * abs_path)
...
+rcutils_ret_t
+rcutils_calculate_directory_size_with_recursion(
+  const char * directory_path,
+  const size_t max_depth,
+  uint64_t * size,
+  rcutils_allocator_t allocator)
 {
-  size_t dir_size = 0;
+  dir_list_t * dir_list = NULL;
+  uint64_t * dir_size = size;
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yeah.

include/rcutils/filesystem.h
Barry-Xu-2018: Okay. Updated

src/filesystem.c
clalancette: > So you will help to execute it. Right ?

Yes, exactly.  Once I approve, I'll go ahead and run the job on https://ci.ros2.org.

clalancette:
I did some minor cleanups in the test code just to avoid memory leaks if an assertion fails.  You can see that work in 6b16702.  With that, I'm happy with this code, so I'm going to approve and run CI.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8126/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10911/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13209/)

clalancette:
So the Windows CMake warnings and test warnings can be ignored; those are in the nightlies (and at least some of them should be fixed for tomorrow).

The MSBuild ones I *believe* are fixed on master.  So I just rebased this whole thing onto master and force-pushed it.  I'll run another Windows CI to check:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13220/)

clalancette:
All right, Windows is now happier.  One more run of the rest of the platforms, and I'll merge tomorrow:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13207/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8138/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10923/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13220/)

clalancette:
All right, going ahead and merging along with https://github.com/ros2/rosbag2/pull/567.  Thanks for the contribution @Barry-Xu-2018 .

PR #311, opened 2020-11-11
Disable a Windows platform warning.

clalancette:

This should get rid of most of the warnings we are currently getting from the Windows CI jobs, as seen in https://ci.ros2.org/view/nightly/job/nightly_win_deb/1804/msbuild .  See the diff for a better explanation of why we have to do this.

### DIFF
--- a/include/rcutils/stdatomic_helper/win32/stdatomic.h
+++ b/include/rcutils/stdatomic_helper/win32/stdatomic.h
@@ -63,7 +63,19 @@
 #ifndef RCUTILS__STDATOMIC_HELPER__WIN32__STDATOMIC_H_
 #define RCUTILS__STDATOMIC_HELPER__WIN32__STDATOMIC_H_
 
+// When building with MSVC 19.28.29333.0 on Windows 10 (as of 2020-11-11),
+// there appears to be a problem with winbase.h (which is included by
+// Windows.h).  In particular, warnings of the form:
+//
+// warning C5105: macro expansion producing 'defined' has undefined behavior
+//
+// See https://developercommunity.visualstudio.com/content/problem/695656/wdk-and-sdk-are-not-compatible-with-experimentalpr.html
+// for more information.  For now disable that warning when including windows.h
+
+#pragma warning(push)
+#pragma warning(disable : 5105)
 #include <Windows.h>
+#pragma warning(pop)
 
 #include <stddef.h>
 #include <stdint.h>
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -27,7 +27,18 @@ extern "C"
 #include <dirent.h>
 #include <unistd.h>
 #else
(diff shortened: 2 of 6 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13020/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7970/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10747/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13003/)

clalancette:
Looks like I missed one in time_win32.c; that should now be fixed.  Otherwise, initial CI looks pretty good.  I'm going to run a full CI just to make sure that everything still works for all downstreams as well.  I expect that to still have one warning on Windows, which needs a separate patch for RViz.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13022/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7972/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10749/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13005/)

ivanpauno (approved):
LGTM!

I'm surprised I couldn't find any issue saying this can happen without using `/experimental:preprocessor`, but maybe Windows introduced a bug in MSVC yesterday (?).

The other option is that we're enabling that compiler flag accidentally.
It would be good to double check that.

wjwwood (approved):

clalancette:
> I'm surprised I couldn't find any issue saying this can happen without using `/experimental:preprocessor`, but maybe Windows introduced a bug in MSVC yesterday (?).

So we don't have `/experimental:preprocessor` on, as far as I can tell.  We *do* turn on `/Zc:preprocessor` (the successor to `/experimental:preprocessor`) in one place: https://github.com/ros2/rosidl/blob/e22083c970f6235ef4266bab80947f9dbc316c37/rosidl_generator_cpp/CMakeLists.txt#L108 .  But as far as I can tell, that should only be for that particular test, so it wouldn't be a global thing.

I'm going to go ahead and merge this as-is to reduce our warnings for now.  If Windows/MSVC gets fixed, or we figure out that we are accidentally turning on flags, we can always revert this.

clalancette:
(oh, and all of the failures on Windows are there on the nightlies as well)

ivanpauno:
> /Zc:preprocessor (the successor to /experimental:preprocessor) in one place: https://github.com/ros2/rosidl/blob/e22083c970f6235ef4266bab80947f9dbc316c37/rosidl_generator_cpp/CMakeLists.txt#L108 . 

Yeah, I saw that one.

> But as far as I can tell, that should only be for that particular test, so it wouldn't be a global thing.

Agreed, I don't think that will affect other places.

PR #315, opened 2020-11-12
use a better way to check whether string is empty

iuhilnehc-ynos:
call `strlen` twice at https://github.com/ros2/rcutils/blob/4c534768869bdec1bf380832729e1eb2f2bed896/src/split.c#L41-L47, I believe L41 can be updated from `strlen` to `'\0' == *str`

Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/src/split.c
+++ b/src/split.c
@@ -38,7 +38,7 @@ rcutils_split(
     RCUTILS_SET_ERROR_MSG("string_array is null");
     return RCUTILS_RET_INVALID_ARGUMENT;
   }
-  if (NULL == str || strlen(str) == 0) {
+  if (NULL == str || '\0' == *str) {
     *string_array = rcutils_get_zero_initialized_string_array();
     return RCUTILS_RET_OK;
   }


clalancette (approved):
In the case where this is an empty string, this may be *slightly* faster, since a function call doesn't have to be made.

But in the case where it is a non-empty string, this will be substantially faster; it is constant time rather than the length of the string.

Works for me.  I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13034/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7983/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10759/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13015/)

fujitatomoya (approved):

clalancette:
The macOS failure happened in the nightly as well: https://ci.ros2.org/view/nightly/job/nightly_osx_release/1869/

Windows also had failures (its a bit random on Windows, but in the last week or so all of the failures also failed): https://ci.ros2.org/view/nightly/job/nightly_win_rel/1750/

With that, I'm going to merge this.  Thanks for the contribution.

PR #316, opened 2020-11-19
Address unused return values found in scan-build

brawner:
These return values were not being checked. After addressing these scan-build passes for `rcutils`.

### DIFF
--- a/test/test_array_list.cpp
+++ b/test/test_array_list.cpp
@@ -237,6 +237,7 @@ TEST_F(ArrayListPreInitTest, remove_success_removes_from_list) {
   EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
 
   ret = rcutils_array_list_get_size(&list, &size);
+  EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
   EXPECT_EQ(size, (size_t)1);
 
   ret = rcutils_array_list_remove(&list, index);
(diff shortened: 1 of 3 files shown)


test/test_array_list.cpp
```diff
@@ -237,6 +237,7 @@ TEST_F(ArrayListPreInitTest, remove_success_removes_from_list) {
   EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
 
   ret = rcutils_array_list_get_size(&list, &size);
+  EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
```
clalancette, line 240: I think this one (and the ones below) need to be `ASSERT_EQ`; otherwise, the following line will be testing an invalid value (in point of fact, if this were the first test in this function, it would probably flag a warning that this was accessing an uninitialized value).

clalancette (changes requested):

test/test_array_list.cpp
brawner: What is uninitialized here? `size` is initialized in line 233 and `list` is initialized in `ArrayListPreInitTest::SetUp`.

Switching to ASSERT_EQ though, the rest of the test does depend on success after each function call here though.

brawner:
Testing `--packages-select rcutils`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13129/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8060/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10838/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13123/)

fujitatomoya (approved):
lgtm

test/test_array_list.cpp
clalancette: If we assume for a second that `rcutils_array_list_get_size` failed, then it may not have set the value of `size` at all.  In that case, the check would be testing against a previously set value of `size`.  That's just incorrect here, but if this were the top of the function, then it could potentially access an uninitialized variable.  That is, the following code is wrong:

```
TEST(foo, bar)
{
  size_t size;
  rcutils_ret_t ret = rcutils_array_list_get_size(&list, &size);
  EXPECT_EQ(ret, RCUTILS_RET_OK);
  EXPECT_EQ(size, 0);
}
```

Still assuming that `rcutils_array_list_get_size` fails, then this code is also wrong:

```
TEST(foo, bar)
{
  size_t size = 0;
  rcutils_ret_t ret = rcutils_array_list_get_size(&list, &size);
  EXPECT_EQ(ret, RCUTILS_RET_OK);
  EXPECT_EQ(size, 0);
}
```

(that is, we're testing the 0 that was set as part of the initialization, not from the return value of `rcutils_array_list_get_size`, which may have failed)

In any case, you've changed these to `ASSERT_EQ` now, so I'm happy :).

clalancette (approved):

Issue #143, opened 2019-03-25
find_library: Prefer currently loaded libraries over searching `LD_LIBRARY_PATH`


EricCousineau-TRI:
At present, `find_library` drops straight into searching libraries along `LD_LIBRARY_PATH`, rather than first querying loaded libraries. One of the 3 current implementations:
https://github.com/ros2/rmw_implementation/blob/32c3de1/rmw_implementation/src/functions.cpp#L77

It'd be nice if this code instead preferred currently loaded libraries, to support application code that *don't* want the ability to switch DDS implementations (e.g. simplifying repro cases, etc.) at runtime, and instead RPATH link the appropriate libraries.

EricCousineau-TRI:
Here is a working prototype on Linux:
https://github.com/EricCousineau-TRI/rcpputils/commit/b66a3d9e02ace44e6c3b6022eeb3447b121d3311

Cribbing from Drake source:
https://github.com/RobotLocomotion/drake/blob/4c6246197/common/find_loaded_library.cc

EricCousineau-TRI:
At this commit for this Bazel repro project, I can now either specify the RMW implementation at linking time, or defer to environment variables:
https://github.com/EricCousineau-TRI/repro/commit/cea11026722b340580257564dc49cb0f59b55178

@dirk-thomas I would find this *extremely* useful. Can I ask what (if any) evidence might sway you to consider this? :D

dirk-thomas:
> Can I ask what (if any) evidence might sway you to consider this?

Since the goal of having a distro is to ensure stability backports are for bug fixes only. Enhancements should target the next distro.

The first test packages of Dashing are targeted to be available in the first week of April.

EricCousineau-TRI:
Aye, sounds good. Will await the merge of https://github.com/ros2/rcpputils/issues/3, and then file this against the dev branch. Thanks!

wieset:
We have a ros2 node that needs capabilities set via `setcap` which leads to `LD_LIBRARY_PATH` being omitted during execution. So we are setting `RPATH` in the binary to find shared libraries. However, as `find_library` relies on `LD_LIBRARY_PATH` the node fails with
```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
[...]

PR #320, opened 2020-12-11
Improve shared library relative paths handling

hidmic:
Connected to #143. This patch ensures relative library paths are properly handled.

CI up to `test_communication`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13267/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8198/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10987/) (known test failure)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13310/)

### DIFF
--- a/src/shared_library.c
+++ b/src/shared_library.c
@@ -105,7 +105,7 @@ rcutils_load_shared_library(
       image_name = candidate_name;
     }
     if (dlclose(handle) != 0) {
-      RCUTILS_SET_ERROR_MSG("dlclose error: %s", dlerror());
+      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("dlclose error: %s", dlerror());
       ret = RCUTILS_RET_ERROR;
       goto fail;
     }
--- a/src/shared_library.c
+++ b/src/shared_library.c
@@ -20,6 +20,11 @@ extern "C"
 #include <stdlib.h>
 
 #ifndef _WIN32
+#if defined(__APPLE__)
+#include <mach-o/dyld.h>
+#elif defined (_GNU_SOURCE)
+#include <link.h>
+#endif
 #include <dlfcn.h>
 #else
 // When building with MSVC 19.28.29333.0 on Windows 10 (as of 2020-11-11),
@@ -35,6 +40,10 @@ extern "C"
 #include <windows.h>
 #pragma warning(pop)
 C_ASSERT(sizeof(void *) == sizeof(HINSTANCE));
+#ifdef UNICODE
+#error "rcutils does not support Unicode paths"
+#endif
+C_ASSERT(sizeof(char) == sizeof(TCHAR));
 #endif  // _WIN32
 
 #include "rcutils/error_handling.h"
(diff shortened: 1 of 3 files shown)


src/shared_library.c
```diff
@@ -65,33 +74,116 @@ rcutils_load_shared_library(
...
+      RCUTILS_SET_ERROR_MSG("dyld image index out of range");
+      ret = RCUTILS_RET_ERROR;
+      goto fail;
+    }
+    void * handle = dlopen(candidate_name, RTLD_LAZY | RTLD_NOLOAD);
+    if (handle == lib->lib_pointer) {
+      image_name = candidate_name;
+    }
+    if (dlclose(handle) != 0) {
+      RCUTILS_SET_ERROR_MSG("dlclose error: %s", dlerror());
```
ahcorde, line 108: RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING

ahcorde (changes requested):

src/shared_library.c
hidmic: See https://github.com/ros2/rcutils/pull/320/commits/5915051a0b7544e834ac7625e87c0b97952d1b2b.

ahcorde (approved):
LGTM

src/shared_library.c
```diff
@@ -65,33 +74,116 @@ rcutils_load_shared_library(
...
+  lib->lib_pointer = dlopen(library_path, RTLD_LAZY);
+  if (NULL == lib->lib_pointer) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("dlopen error: %s", dlerror());
+    return RCUTILS_RET_ERROR;
+  }
+
+#if defined(__APPLE__)
+  const char * image_name = NULL;
+  uint32_t image_count = _dyld_image_count();
+  for (uint32_t i = 0; NULL == image_name && i < image_count; ++i) {
```
azeey, line 95: I feel like I'm missing something. Is my understanding correct that what ends up in `lib->library_path` is the absolute path of `library_path`? If so, can't we just call `realpath`? Or is it possible `library_path` and `lib->library_path` will point to two different files?

hidmic: > If so, can't we just call `realpath`?

The thing is, `library_path` might simply be the library name e.g. `libfoo.so`. And that doesn't mean it should be searched for in the current working directory. It's only after `dlopen` performs the search over RPATHs, (DY)LD_LIBRARY_PATHs, and RUNPATHs and actually finds it that you can get the full path to it.

azeey: Ah, that makes sense. Thanks! I don't know what the comment etiquette is for this package, but I think it would be great to have that paragraph as a comment just to explain what the goal of this section of code is.

azeey (approved):
LGTM

src/shared_library.c
hidmic: Sure thing. See 1d4276a.

hidmic:
Ok, I finally managed to write some tests! It stumbled with `CMake` and `dlopen` quirks, but it should be fine now.

hidmic:
Re-running CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13315/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8245/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11045/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13355/)

hidmic:
Alright, all tests passing! Thanks for the reviews.

cottsay:
@hidmic, this change broke clang nightly builds on macOS. Please submit a fix.
https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/

PR #331, opened 2021-03-10
Update the documentation to be more consistent.

clalancette:
1. Switch \returns to \return
2. Document things that were missing
3. Switch enum linkage to use #RCUTILS_RET_OK so linking works
4. Add @file to all files we want to document


ahcorde (approved):

clalancette:
Thanks for the review.  Given that this is all changes to the documentation, I'm going to take the PR job as good enough and merge this without additional CI.

PR #333, opened 2021-03-12
Stop double-defining structs.

clalancette:
Currently we use something like:

```c
typedef struct foo_t
{
} foo_t;
```

But that actually double-defines the `foo_t` symbol
(at least, according to Doxygen).  Instead, use the
more traditional:

```c
typedef struct foo
{
} foo_t;
```

This is slightly controversial in that it is an API break;
any code that was using `struct foo_t` before will now fail
to compile (anything that was just using `foo_t` will be fine).

However, this should mostly affect low-level libraries
like client libraries and the such, so the scale of breakage
here should be low.

### DIFF
--- a/include/rcutils/allocator.h
+++ b/include/rcutils/allocator.h
@@ -44,7 +44,7 @@ extern "C"
  * Developers should note that, while the fields of a const-qualified allocator
  * struct cannot be modified, the state of the allocator can be modified.
  */
-typedef struct rcutils_allocator_t
+typedef struct rcutils_allocator_s
 {
   /// Allocate memory, given a size and the `state` pointer.
   /** An error should be indicated by returning `NULL`. */
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -102,14 +102,14 @@ extern "C"
     1)
 
 /// Struct wrapping a fixed-size c string used for returning the formatted error string.
-typedef struct rcutils_error_string_t
+typedef struct rcutils_error_string_s
 {
   /// The fixed-size C string used for returning the formatted error string.
   char str[RCUTILS_ERROR_MESSAGE_MAX_LENGTH];
 } rcutils_error_string_t;
 
 /// Struct which encapsulates the error state set by RCUTILS_SET_ERROR_MSG().
-typedef struct rcutils_error_state_t
+typedef struct rcutils_error_state_s
 {
   /// User message storage, limited to RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH characters.
   char message[RCUTILS_ERROR_STATE_MESSAGE_MAX_LENGTH];
--- a/include/rcutils/filesystem.h
+++ b/include/rcutils/filesystem.h
@@ -246,7 +246,7 @@ size_t
 rcutils_get_file_size(const char * file_path);
 
 /// An iterator used for enumerating directory contents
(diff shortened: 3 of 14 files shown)


clalancette:
CI (everything, just to ensure there is no higher-level breakage):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13926/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8769/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11630/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14012/)

ahcorde (approved):

fujitatomoya (approved):
lgtm

wjwwood (changes requested):
I would actually prefer to see this pattern:

```c
typedef struct rcutils_foo_s {
  // ...
} rcutils_foo_t;
```

My reasons for this:

- we already have code like `rcutils_foo_t rcutils_foo = ...;`
- other users (outside our code base) could be using `rcutils_foo` (without the `_t`) causing a collision, but I think it is much less likely anyone is using `rcutils_foo_s`
- having some suffix helps denote that it may be a complex type
- though it is discouraged to use `_t` in your projects, the reasoning is that `POSIX` may arbitrarily add new `<type>_t` types, but since we namespace 100% of our C symbols with `<package_name>_`, that's not a problem
- when using auto-complete seeing `foo_t` and `foo_s` indicates to me (without knowing the code base) which is the struct and which is the typedef, i.e. with `foo` and `foo_t` it could just as easily be `struct foo_t` as `struct foo` in my opinion

Thoughts?

---

If others agree, then I can update this pull request and add some documentation to our contributing/style guides.

wjwwood:
> we already have code like `rcutils_foo_t rcutils_foo = ...;`

I should clarify I think this is the case if we expand this pattern to other packages like `rmw` and `rcl`, though I don't think we have any instances of it for `rcutils`.

clalancette:
> Thoughts?

I honestly don't have a strong opinion either way, so I'll go ahead and rebase this and change it to `_s` everywhere.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14986/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9727/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12672/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15150/)

wjwwood (approved):

wjwwood:
Rerunning CI since https://github.com/ros2/system_tests/pull/481 and https://github.com/ros2/rcl_interfaces/pull/125 were merged:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15003/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9742/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12690/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15165/)

clalancette:
The failures in Windows were also in the nightlies.  I'm going to go ahead and merge this, thanks for the reviews!

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

PR #334, opened 2021-03-15
Quiet down a warning in release mode.

clalancette:
Since assert gets compiled away in release mode, there's a
warning thrown that ret is unused.  Just quiet it down here
since it is in benchmark code.

### DIFF
--- a/test/benchmark/benchmark_error_handling.cpp
+++ b/test/benchmark/benchmark_error_handling.cpp
@@ -24,6 +24,7 @@ static void benchmark_err_handling(benchmark::State & state)
     rcutils_ret_t ret =
       rcutils_initialize_error_handling_thread_local_storage(rcutils_get_default_allocator());
     assert(ret == RCUTILS_RET_OK);
+    (void)ret;
     rcutils_reset_error();
     const char * test_message = "test message";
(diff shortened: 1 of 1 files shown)


cottsay:
Which platform is complaining? I thought that in release mode, `assert(foo)` evaluated to `(void)foo`.

cottsay (approved):
Ah, I'm mistaken. It appears that the expression is discarded if `NDEBUG` is specified: https://sourceware.org/git/?p=glibc.git;a=blob;f=assert/assert.h;h=2894908559560747920961a38de210a16318fe75;hb=HEAD#l50

That's unfortunate.

clalancette:
CI (just up to rcutils):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13942/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8782/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11646/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14029/)
* Linux Release [Build Status](http://ci.ros2.org/job/ci_linux/13943/)

PR #340, opened 2021-04-30
Deprecate get_env.h and move content to env.{h,c}

christophebedard:
Resolves these `TODO`s:

https://github.com/ros2/rcutils/blob/ca42bc9f588e11ad40f5879dcb9a364b435d2cba/include/rcutils/env.h#L30

https://github.com/ros2/rcutils/blob/ca42bc9f588e11ad40f5879dcb9a364b435d2cba/src/env.c#L26

I think the intention was originally to deprecate in Galactic and remove in H-turtle (https://github.com/ros2/rcutils/pull/250#issuecomment-628785186), but it missed the :boat: so it'll have to be deprecated in H-turtle and removed in I-turtle. I can create an issue for the removal after this is merged.


include/rcutils/env.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Open Source Robotics Foundation, Inc.
+// Copyright 2017-2021 Open Source Robotics Foundation, Inc.
```
clalancette, line 1: Let's not change this; we haven't been updating it anywhere else.  Same goes for all of the copyright dates below.

include/rcutils/env.h
```diff
@@ -58,6 +55,67 @@ RCUTILS_WARN_UNUSED
...
+ * The c-string which is returned is only valid until the next time this
+ * function is called, because it is a direct pointer to the static storage.
+ * Also note that the string returned here should *not* be freed.
+ *
+ * The function first tries to get the HOME environment variable.
+ * If that variable exists and is non-empty, that will be returned.
+ * Otherwise, the function tries to get the USERPROFILE environment variable.
+ * If that variable exists and is non-empty, that will be returned.
+ * If neither exists, NULL will be returned.
+ * The above algorithm is portable across both Unix and Windows.
```
clalancette, line 106: I know that you are just moving this, but this documentation is actually incorrect.  On Unix, `rcutils_get_home_dir` will try to get the HOME environment variable only.  If that fails, it returns NULL.  On Windows, `rcutils_get_home_dir` will try to get the HOME environment variable.  If that fails, it tries to get the USERPROFILE environment variable.  If that fails, it returns NULL.

include/rcutils/get_env.h
```diff
@@ -17,77 +17,14 @@
...
-const char *
-rcutils_get_home_dir(void);
-
-#ifdef __cplusplus
-}
+#ifdef _MSC_VER
+#pragma message ("rcutils/get_env.h has been deprecated, please include rcutils/env.h instead")
+#else
+#warning rcutils/get_env.h has been deprecated, please include rcutils/env.h instead
 #endif
```
clalancette, line 26: Do we really need a separate `#pragma` on Windows?  Looking around the codebase, there are a couple of other instances that just unconditionally use `#warning`, which leads me to believe it is supported by MSVC.  But I'm not sure.

clalancette (changes requested):
A few minor things to fix up, but this is a good cleanup.  Thanks!

include/rcutils/get_env.h
christophebedard: Yeah I was wondering too. `#warning` is a `gcc` extension, and apparently to get an actual/proper warning on MSVC we have to use `#pragma message`.

In any case, I can just use `#warning` if you prefer.

include/rcutils/env.h
christophebedard: Alright, I reverted 4eb6f857ba63e3e26b8f2384c10031cff7f3a399: fed8ca8fdf7094adb99ce2e936162c64f78dfffa

include/rcutils/env.h
christophebedard: Fixed in 7eb69e04cecd44bd3bf0a47c38bae5b05a873ba3

fujitatomoya (approved):
looks good to me

include/rcutils/get_env.h
clalancette: > Yeah I was wondering too. `#warning` is a `gcc` extension, and apparently to get an actual/proper warning on MSVC we have to use `#pragma message`.
> 
> In any case, I can just use `#warning` if you prefer.

Ah, I didn't realize.  OK, in that case we can leave this in place, I was just wondering why it was different.

clalancette (approved):
Looks good to me with green CI.

christophebedard:
Should I do a CI run for this PR + the other related PRs all at once?

clalancette:
> Should I do a CI run for this PR + the other related PRs all at once?

Yeah, that would be my suggestion.  And in this case, since we are touching so many different packages, I'll just suggest doing one run where we build and test everything.

christophebedard:
Test failures on macOS look unrelated.

clalancette:
> Test failures on macOS look unrelated.

Yeah, agreed.  Annoying, because it looks new, but not related to this PR.

PR #178, opened 2019-08-30
check and link against libatomic (#172)

sloretz:
Cherry-pick of #172 after reverting in #177. This fixes #171, but needs more work because it causes a CMake warning about `atomic` not being found in downstream libraries.

@christianrauch FYI

Issue #171, opened 2019-07-22
Linking error (test_atomics_executable) on Raspbian/armv6l


christianrauch:
I am trying to compile ROS dashing from source on Raspbian Buster (`gcc (Raspbian 8.3.0-6+rpi1) 8.3.0`) running on armv6l and encounter a linking error for target `test_atomics_executable`:
```
/usr/bin/ld: CMakeFiles/test_atomics_executable.dir/test/test_atomics.c.o: in function `main':
test_atomics.c:(.text+0x11a8): undefined reference to `__atomic_store_8'
/usr/bin/ld: test_atomics.c:(.text+0x11c0): undefined reference to `__atomic_load_8'
/usr/bin/ld: test_atomics.c:(.text+0x1268): undefined reference to `__atomic_store_8'
/usr/bin/ld: test_atomics.c:(.text+0x12b8): undefined reference to `__atomic_exchange_8'
[...]
```

I am using the standard repo at https://raw.githubusercontent.com/ros2/ros2/release-latest/ros2.repos, excluding `rviz2` and `rqt`.

Edit: The problem can be resolved by manually adding `-latomic`, i.e. `target_link_libraries(test_atomics_executable ${PROJECT_NAME} -latomic)`.

clalancette:
> Edit: The problem can be resolved by manually adding `-latomic`, i.e. `target_link_libraries(test_atomics_executable ${PROJECT_NAME} -latomic)`.

Please open a PR with that change; we'll gladly review it.  Thanks.

christianrauch:
I discovered that I get the same linking issues in a couple of other packages (`tlsf_cpp`, `rcl`, `rclcpp`, `demo_nodes_cpp`, ...) which can all be resolved by linking against `libatomic`.
Before sending PRs for all of them, adding `-latomic`, I would like to hear back if there might be a better solution. After all, this seems to happen either with never GCC (>8) or specifically on Raspbian.

sloretz:
[...]

PR #341, opened 2021-06-12
fix linking on raspbian buster

BrettRD:
Raspbian buster still doesn't have atomics in the standard symbol table, this is a continuation of https://github.com/ros2/rcutils/pull/178

discussion: https://github.com/ros2/rcutils/issues/171#issuecomment-860006445

This pull request only adds libatomic where it was necessary to build foxy on a pi4 using `colcon build  --symlink-install --packages-select rcutils`


BrettRD:
This change seems a bit too brittle and verbose, pulling foxy latest to merge it turned up a bunch of new linker errors.

Could this be converted into flag that is conditionally added to a linker list and passed to all targets?

BrettRD:
This is the tip of the iceberg, every package that links against rcutils needs extra flags.  
After patching both `rcutils` and `rcpputils` (in the same way), the build stops at  `5 packages failed: class_loader fastrtps libyaml_vendor rcl_logging_spdlog rosidl_runtime_c`, all missing the same linker flags, I can only assume there are more in the 122 packages colcon couldn't reach.

I don't have this issue on Armbian, just Raspi OS.  
It would be nice to have clean builds on the raspberry pi, but until either it gets fixed upstream (Raspi OS) or the raspi link step goes into ROS2 testing, this patch is going to stop working for pi users very quickly.

I'm going to leave this patch here and pass flags to colcon manually.

BrettRD:
This patch is an ineffective bandaid for a problem with the Raspbian build system.
Closing

see: https://github.com/ros2/rcutils/issues/171#issuecomment-1123243641 for workarounds (and then run a clean build)

PR #345, opened 2021-08-05
Use FindPython3 explicitly instead of FindPythonInterp implicitly

sloretz:
Rcutils implicitly depends on FindPythonInterp being called by `ament_cmake_core`. I noticed it will fail to build while testing a branch to use `FindPython3` in `ament_cmake`. This PR makes the python interpreter dependency explicit, and uses `FindPython3` instead of `FindPythonInterp`

Related to ros2/python_cmake_module#6
Blocks ament/ament_cmake#355

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.12)
 
 project(rcutils)
 
@@ -15,6 +15,8 @@ include(CheckLibraryExists)
 
 find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
+
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
 
 ament_python_install_package(${PROJECT_NAME})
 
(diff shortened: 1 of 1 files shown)


sloretz:
CI (build: `--packages-up-to rcutils` test: ` --packages-select rcutils`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14990/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9730/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12676/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15153/)

aprotyas (approved):

CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ include(CheckLibraryExists)
 find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
 
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
clalancette, line 19: So this is great that we are making this an explicit dependency in CMakeLists.txt.

I wonder if we should also make `python3` an explicit `build_depend`/`buildtool_depend` in the package.xml.  It will always implicitly get pulled in because of the `buildtool_depend` on `python3-empy`, but it is a little weird to `find_package` something that doesn't have an entry in the package.xml.  Thoughts?

sloretz: I think it's reasonable to add `python3` as the `buildtool_depend`. This is getting that for free from `ament_cmake`, so adding it to the `package.xml` fits the theme of being explicit. It seems unlikely to have an impact in the future as I imagine `ament_cmake` will always depend on Python.

> it is a little weird to find_package something that doesn't have an entry in the package.xml

The `FindPython3` module is provided by CMake, so this is `find_packaging()` is a module from CMake itself.

clalancette: OK, that's a fair point.  In that case, I don't feel strongly about whether we have a package.xml depends on `python` or not.  I'm going to go ahead and approve this, and let you decide whether to add the depends.

clalancette (approved):

CMakeLists.txt
sloretz: I'll leave the `package.xml` as is for expediency.

PR #346, opened 2021-08-10
Add qsort to the excluded patterns.

clalancette:
This gets rid of the last rosdoc2 warning on the buildfarm.

The last warning can be seen at the end of the log here: https://build.ros2.org/view/Rdoc/job/Rdoc__rcutils__ubuntu_focal_amd64/4/console .

It's not ideal to be disabling the generation of the docs for this, but I think it is more important to get some green Rdoc builds going than fixing this one lightly-used API.  We can always try to make it work later.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -6,10 +6,9 @@ PROJECT_BRIEF          = "C API providing common utilities and data structures."
 
 # Use these lines to include the generated logging_macro.h (update install path if needed)
 #INPUT                  = README.md ../../../install_isolated/rcutils/include
-#STRIP_FROM_PATH        = /Users/william/ros2_ws/install_isolated/rcutils/include
 # Otherwise just generate for the local (non-generated header files)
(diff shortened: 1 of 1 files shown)


wjwwood:
I think this is the issue @aprotyas originally found related to a issue with newer versions of breathe. 

This might remove the last warning from here but it's a system problem that we'll probably run into over and over unless we get a fix in breathe or pin the version back.

aprotyas:
Yes, this is a known issue https://github.com/michaeljones/breathe/issues/722 and it looks like there is a fix in the process.

Till then, either the problematic signatures can be excluded - but this would fix the problem only for this repo - or a specific version of breathe (**<=4.26.0**) can be pinned back in the Rdoc jobs.

clalancette:
With the latest release of breathe (4.31.0), this is no longer a problem.  So closing this PR.

PR #347, opened 2021-09-23
Add option to set logger severity level using a config file

JafarAbdi:
Add ability to specify the logger severity using a config file similar to ROS1's `ROSCONSOLE_CONFIG_FILE` env-variable, this can be done by `export RCUTILS_LOGGING_CONFIG_FILE=path_to_config_file`

This's a demo:

https://user-images.githubusercontent.com/16278108/134510231-f88ec412-ae1d-44b0-a832-d48c12cfa3f4.mp4


src/logging.c
```diff
@@ -293,6 +293,44 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
     }
 
     g_rcutils_logging_initialized = true;
+
+    // We load the logging configs after setting g_rcutils_logging_initialized
+    // to true otherwise rcutils_logging_set_logger_level will cause recursive
+    // call to this function due to RCUTILS_LOGGING_AUTOINIT Check for the
+    // Check for the environment variable for selecting logging level
+    const char * logging_config_filename;
```
facontidavide, line 301: always good to initialize the variable to NULL

src/logging.c
```diff
@@ -293,6 +293,44 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+        if (RCUTILS_RET_OK != rcutils_logging_severity_level_from_string(
+            severity, g_rcutils_logging_allocator, &severity_level))
+        {
+          RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING(
+            "Logger has an invalid severity level: %s\n", severity);
+          return RCUTILS_RET_ERROR;
+        }
+        if (RCUTILS_RET_OK != rcutils_logging_set_logger_level(logger_name, severity_level)) {
+          RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING(
+            "Failed to set severity level: %s for logger '%s'\n", severity, logger_name);
```
facontidavide, line 328: Don't you want to `return RCUTILS_RET_ERROR;` here?

src/logging.c
```diff
@@ -293,6 +293,44 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
     }
 
     g_rcutils_logging_initialized = true;
+
+    // We load the logging configs after setting g_rcutils_logging_initialized
+    // to true otherwise rcutils_logging_set_logger_level will cause recursive
+    // call to this function due to RCUTILS_LOGGING_AUTOINIT Check for the
+    // Check for the environment variable for selecting logging level
+    const char * logging_config_filename;
+    ret_str = rcutils_get_env("RCUTILS_LOGGING_CONFIG_FILE", &logging_config_filename);
```
fujitatomoya, line 302: how about define for `RCUTILS_LOGGING_CONFIG_FILE`, so that we can refer to it.

src/logging.c
```diff
@@ -293,6 +293,44 @@ rcutils_ret_t rcutils_logging_initialize_with_allocator(rcutils_allocator_t allo
...
+    if (NULL == ret_str && strcmp(logging_config_filename, "") != 0) {
+      FILE * logging_config_file = fopen(logging_config_filename, "r");
+      if (NULL == logging_config_file) {
+        RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+          "Failed to open logging config file `[%s]`.",
+          logging_config_filename);
+        return RCUTILS_RET_ERROR;
+      }
+      char logger_name[50];
+      char severity[10];   // fatal error debug info warn case insensitive
```
fujitatomoya, line 312: how about allocate memory dynamically? I think that how it is done when parsing the arguments logging level.

fujitatomoya (commented):
some questions and comments,

- environment variable prevails logging level arguments? I am not sure which one prevails to the other.
- it should be able to set `default` level as well in the configuration file?
- probably it would be nice to follow up tutorials how to configure logging level via `RCUTILS_LOGGING_CONFIG_FILE`?

destogl (commented):
This would be very, very useful!

aprotyas:
>     * environment variable prevails logging level arguments? I am not sure which one prevails to the other.

@fujitatomoya is there such a facility to set logging levels through environment variables? I don't think so, in which case this question would be moot for now.

fujitatomoya:
> is there such a facility to set logging levels through environment variables?

No, i do not think so. not via environmental variable, but we can change it with ros arguments. so i just thought that it should be discussed as design before implementation?

src/logging.c
JafarAbdi: Good point, I did some googling, looks like having scanf allocate memory is a [posix extension](https://stackoverflow.com/questions/11926883/scanf-dynamic-allocation) which mean that it may not work on windows (I'm not sure)

JafarAbdi:
> environment variable prevails logging level arguments?

No, logging level arguments prevail the environment variable ones, so if we have logger1 set to debug in the `RCUTILS_LOGGING_CONFIG_FILE` and we set it to error in the argument (`ros2 run pkg node --ros-args --log-level logger1:=error`) it will be set to error

> it should be able to set default level as well in the configuration file?

Done in the latest commit

> probably it would be nice to follow up tutorials how to configure logging level via RCUTILS_LOGGING_CONFIG_FILE?

I'm working on it now

PR #349, opened 2022-01-07
Use static_cast instead of C-style cast

jacobperron:
Fixes cpplint error.

Relates to https://github.com/ament/ament_lint/pull/324


fujitatomoya (approved):

clalancette (approved):
Looks good with green CI.

PR #355, opened 2022-04-11
Support ROS2 on Hololens 2

ooeygui:
Hololens 2 uses a version of Windows which specifically prohibits certain APIs from being used. In this repository, prohibited APIs are used - `FindFirstFile`/`FindNextFile`/`FindClose` and `LoadLibrary`. On Universal Windows Platform (UWP), filesystem enumeration is not required, so simply removed. `LoadLibrary` is replaced with `LoadPackagedLibrary`, which loads from the UWP package.

These changes are in the `WIN32` block, using the family partition, so do not affect other build types.

### DIFF
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -233,6 +233,7 @@ rcutils_calculate_directory_size(const char * directory_path, rcutils_allocator_
     return dir_size;
   }
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


ooeygui:
I will apply a version of this change to master and backport.

PR #366, opened 2022-06-11
strdup.c: fix arbitrary length overread

guijan:
rcutils_strndup() would unconditionally read as many bytes as its 2nd argument said regardless of the source string's length, but strndup() is defined to read the minimum between the 2nd argument and the source string's length.

Signed-off-by: Guilherme Janczak <guilherme.janczak@yandex.com>

Hi, I'm back with the related but separate bug in strndup(). It should be easier to understand this time.

### DIFF
--- a/src/strdup.c
+++ b/src/strdup.c
@@ -19,6 +19,7 @@ extern "C"
 
 #include "rcutils/strdup.h"
 
+#include <limits.h>
 #include <stddef.h>
 #include <string.h>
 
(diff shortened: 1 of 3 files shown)


guijan:
My C was invalid C++ again, so here's a quick amend.

src/strdup.c
```diff
@@ -31,20 +32,19 @@ rcutils_strdup(const char * str, rcutils_allocator_t allocator)
 {
   RCUTILS_CAN_RETURN_WITH_ERROR_OF(NULL);
 
-  if (NULL == str) {
-    return NULL;
-  }
-  return rcutils_strndup(str, strlen(str), allocator);
+  return rcutils_strndup(str, SIZE_MAX, allocator);
```
fujitatomoya, line 35: Does this have to be changed into `SIZE_MAX`? Could you give us the problem case?

guijan: Using `strlen()` iterates over the string unnecessarily.

fujitatomoya (approved):
lgtm, thanks for iterating.

clalancette (approved):
This looks like a reasonable fix to me, thanks!  Will run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16972/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11535/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17373/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/17374/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/240/)

PR #367, opened 2022-07-13
Optimize the implementation of rcutils_char_array_strncpy.

clalancette:
It turns out that this function is used a lot in the rcutils
logging code, so it is on a hot-path.  Thus it is worth optimizing.

The optimization itself consists of recognizing that the current
implementation of rcutils_char_array_strncpy is calling strlen
twice; once explicitly at the beginning of the function, and then
again implicitly while calling strncat.  However, it turns out
that we don't need to recalculate the length either time; we
already know that information based on the buffer_length.  Thus
we use the information we already have and speed up the calculation,
particularly for long strings.

In my local testing, I found (using `perf`) that this patch significantly cuts down on the number of strlen invocations.  For very long strings, this can be up to 40% faster (for shorter strings, the speed increase is more modest, around 10%).

I'll note that this also fixes a subtle bug in the interaction of `rcutils_char_array_memcpy` and `rcutils_char_array_strncat`.  Previously, if you ended up calling `rcutils_char_array_memcpy` and did *not* copy a '\0' in, then a following call to `rcutils_char_array_strncat` would run off into arbitrary memory looking for a '\0' during the strlen call.  While this sequence of calls would be a strange usage, it would cause UB before.  Now it would work properly.

### DIFF
--- a/src/char_array.c
+++ b/src/char_array.c
@@ -213,22 +213,23 @@ rcutils_char_array_strcpy(rcutils_char_array_t * char_array, const char * src)
 rcutils_ret_t
 rcutils_char_array_strncat(rcutils_char_array_t * char_array, const char * src, size_t n)
 {
-  size_t current_strlen = strlen(char_array->buffer);
+  size_t current_strlen;
+  if (char_array->buffer_length == 0) {
+    current_strlen = 0;
+  } else {
+    // The buffer length always contains the trailing \0, so the strlen is one less than that.
+    current_strlen = char_array->buffer_length - 1;
+  }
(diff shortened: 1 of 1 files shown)


audrow (approved):
Looks good to me!

clalancette:
Just because this is touching such a low-level piece of code, here is CI on RHEL and Windows Debug as well:

* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/239/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/17372/)

Issue #364, opened 2022-05-24
High logging overhead


almaslov:
Hello!
I was profiling some nodes in my project recently and noticed an issue in case of high incoming messages rate(~2-3K hz). In this case `rcl_wait` consumes significant amount of CPU. About 30-50% of `rcl_wait` execution time is `rcutils_logging_logger_is_enabled_for`:
<img width="1191" alt="perf_flamegraph_" src="https://user-images.githubusercontent.com/10155461/169996508-f2c977c0-7919-4b06-a8da-8a2cedb9828f.png">
It looks like most of overhead can be eliminated by removing `rcutils_logging_severity_level_from_string` and switching to integral `log_level` as mentioned in comment [here](https://github.com/ros2/rcutils/blob/master/src/logging.c#L410) and [here](https://github.com/ros2/rcutils/blob/master/src/logging.c#L481)

almaslov:
I would be glad to make a PR, but I'm a little confused with idea of implementing string-to-int map type which is almost complete copy-paste of string-to-string map. Could you suggest, which solution did you have in mind?

clalancette:
I took a quick look at this as well.  I think there are several problems in here:

1.  `rcutils_logging_severity_level_from_string` is doing a string copy and toupper on the string every time it is invoked (https://github.com/ros2/rcutils/blob/8da2aace17bcecad3bdf0bebafe23298602f50c0/src/logging.c#L320-L355).  While I guess that is nice, I would argue that we can just make this case-sensitive and save a lot of work here.  Since this is changing a semantic, we'd want to some research on how widely used it is, but I think that should be straightforward.
2. As @fujitatomoya pointed out, we are doing a lot of work in `rcutils_logging_get_logger_effective_level` to try to suss out the parts of the logger that we want.  We may be able to optimize this by memoizing that result, though we would still incur overhead in checking on the memoization.  We could also reconsider whether we really need that functionality.  That is more controversial, since I can definitely see uses for the current system, but logging really should be as simple as possible.

I'm not sure I understand your latest proposal.  If we had some kind of global logging update timestamp, wouldn't we just be throttling messages?  While that certainly would help, we already have the `THROTTLE` versions of the messages.

almaslov:
@clalancette Thanks for reply.
[...]

PR #369, opened 2022-07-29
Optimize calls via the RCUTILS_LOG macros.

clalancette:
It turns out that when we call RCUTILS_LOG_DEBUG (and friends),
we were calling rcutils_logging_logger_is_enabled_for() twice;
once in the macro itself (where we want to avoid evaluating
conditions if the logger is disabled), and once in the rcutils_log()
call.  Since rcutils_logging_logger_is_enabled_for() is a pretty
expensive function, this means we were doing unnecessary work.

Improve this by splitting rcutils_log() into rcutils_log() and
rcutils_log_internal(), the latter of which does not do the
enabled check.  Then change the macros to call the internal
version of the API, which avoids the double check while still
maintaining the safety guarantees we want if a user calls
rcutils_log() directly.

### DIFF
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -476,6 +476,7 @@ int rcutils_logging_get_logger_effective_level(const char * name);
  * \param[in] format The format string
  * \param[in] ... The variable arguments
  */
+RCUTILS_PUBLIC
 void rcutils_log_internal(
   const rcutils_log_location_t * location,
   int severity,
--- a/include/rcutils/logging.h
+++ b/include/rcutils/logging.h
@@ -451,10 +451,45 @@ RCUTILS_PUBLIC
 RCUTILS_WARN_UNUSED
 int rcutils_logging_get_logger_effective_level(const char * name);
 
+/// Internal call to log a message.
+/**
+ * Unconditionally log a message.
+ * This is an internal function, and assumes that the caller has already called
+ * rcutils_logging_logger_is_enabled_for().
+ * End-user software should never call this, and instead should call rcutils_log()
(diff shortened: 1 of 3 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17043/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11602/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17468/)

fujitatomoya:
related to https://github.com/ros2/rcutils/issues/364

fujitatomoya (approved):
lgtm! but adding test would be ideal.

clalancette:
> lgtm! but adding test would be ideal.

I ended up not adding tests because I think everything here is already covered elsewhere.  That is, the tests in https://github.com/ros2/rcutils/blob/rolling/test/test_logging.cpp end up testing `rcutils_log`, and the tests in https://github.com/ros2/rcutils/blob/rolling/test/test_logging_macros.cpp test the logging macros (which indirectly test `rcutils_log_internal`).

That said, let me know if you still think I should add additional tests.

fujitatomoya:
@clalancette thanks for the comment, i am good to go with green CI.

clalancette:
Here's another CI try:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17085/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11638/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17506/)

PR #374, opened 2022-08-14
Add in a series of micro-optimizations in rcutils

clalancette:
These were all discovered as part of optimizing the logging implementation.  By themselves, none of these changes make a huge change to performance, but when combined with some of the other optimizations (like in #372 and another upcoming optimization), they can make a small but measurable difference.

The optimizations are as follows:
1. Make hash_map_find() use a shift rather than a divide to find modulo (only works with power-of-two capacities).
2. When increasing the size of a char_array, add at least 1.5x the size to avoid another reallocation soon after.
3. If the hash_map is empty, don't bother computing the key hash when we don't need it.  In the corner case of an empty hash_map, the key hash computation can actually dominate the computation time.
4. When using glibc, use the `memrchr` function that is available with _GNU_SOURCE.  This uses optimized instructions where possible.  On all other platforms, fall back to the old implementation.
5. Make a couple of global variables in logging static.  This should allow the optimizer more freedom.  (this is honestly more a code-cleanliness change rather than performance, but I included it here anyway)

### DIFF
--- a/include/rcutils/types/hash_map.h
+++ b/include/rcutils/types/hash_map.h
@@ -156,7 +156,7 @@ rcutils_get_zero_initialized_hash_map();
  * ```
  *
  * \param[inout] hash_map rcutils_hash_map_t to be initialized
- * \param[in] initial_capacity the amount of initial capacity for the hash_map
+ * \param[in] initial_capacity the amount of initial capacity for the hash_map - this must be greater than zero and a power of 2
  * \param[in] key_size the size (in bytes) of the key used to index the data
  * \param[in] data_size the size (in bytes) of the data being stored
  * \param[in] key_hashing_func a function that returns a hashed value for a key
--- a/src/char_array.c
+++ b/src/char_array.c
@@ -139,6 +139,12 @@ rcutils_char_array_expand_as_needed(rcutils_char_array_t * char_array, size_t ne
     return RCUTILS_RET_OK;
   }
 
+  // Make sure we expand by at least 1.5x the old capacity
+  size_t minimum_size = char_array->buffer_capacity + (char_array->buffer_capacity >> 1);
+  if (new_size < minimum_size) {
+    new_size = minimum_size;
+  }
+
   return rcutils_char_array_resize(char_array, new_size);
 }
 
--- a/src/hash_map.c
+++ b/src/hash_map.c
@@ -439,6 +439,11 @@ rcutils_hash_map_unset(rcutils_hash_map_t * hash_map, const void * key)
   bool already_exists = false;
   rcutils_hash_map_entry_t * entry = NULL;
 
+  // If there is nothing in the hash map, don't bother computing the key
+  if (hash_map->impl->size == 0) {
+    return RCUTILS_RET_OK;
+  }
+
   already_exists = hash_map_find(hash_map, key, &key_hash, &map_index, &bucket_index, &entry);
 
   if (!already_exists) {
@@ -472,6 +477,11 @@ rcutils_hash_map_key_exists(const rcutils_hash_map_t * hash_map, const void * ke
   bool already_exists = false;
   rcutils_hash_map_entry_t * entry = NULL;
 
+  // If there is nothing in the hash map, don't bother computing the key
+  if (hash_map->impl->size == 0) {
+    return RCUTILS_RET_OK;
+  }
+
   already_exists = hash_map_find(hash_map, key, &key_hash, &map_index, &bucket_index, &entry);
 
   return already_exists;
@@ -487,6 +497,11 @@ rcutils_hash_map_get(const rcutils_hash_map_t * hash_map, const void * key, void
   size_t key_hash = 0, map_index = 0, bucket_index = 0;
   bool already_exists = false;
   rcutils_hash_map_entry_t * entry = NULL;
+
+  // If there is nothing in the hash map, don't bother computing the key
+  if (hash_map->impl->size == 0) {
+    return RCUTILS_RET_NOT_FOUND;
+  }
 
   already_exists = hash_map_find(hash_map, key, &key_hash, &map_index, &bucket_index, &entry);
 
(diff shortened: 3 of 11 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17134/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11681/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17552/)

gavanderhoorn:
> Make hash_map_find() use a shift rather than a divide to find modulo (only works with power-of-two capacities).

wouldn't any reasonable compiler already optimise this case?

At least GCC should recognise the cases where it can use shifts, see [this godbolt test fi](https://godbolt.org/z/GqPEo75ej) (in fact, even the non-power-of-two gets optimised to use shifts).

clalancette:
> Edit: `/ignore`: the divisor is not known at compile-time. That prevents any such optimisation.

Yeah, exactly.  I should point out that I put in these optimizations as I was actually profiling logging in rcutils.  In this particular case, I saw that the divide was taking up a not-insignificant amount of time, which is why I added the optimization here.

src/char_array.c
```diff
@@ -139,6 +139,12 @@ rcutils_char_array_expand_as_needed(rcutils_char_array_t * char_array, size_t ne
     return RCUTILS_RET_OK;
   }
 
+  // Make sure we expand by at least 1.5x the old capacity
```
fujitatomoya, line 142: I think it would be better to move this to header file doc section, not in the implementation.

src/hash_map.c
```diff
@@ -255,6 +255,13 @@ rcutils_hash_map_init(
     return RCUTILS_RET_INVALID_ARGUMENT;
   }
 
+  // Due to an optimization we use during lookup, we can currently only handle power-of-two
+  // capacities.  Enforce that here.
```
fujitatomoya, line 259: Since this is hard requirement, we should add this to doc section in header. (Example, does not even work anymore.)
besides, I think it would be nice to add this failure case in the test.

src/char_array.c
clalancette: The thing is, the change to 1.5x is an implementation detail, not something the user really needs to worry about.  That is, other than a slight increase in memory, the user won't really notice anything that changes because of this.  It should just avoid the reallocation logic on every small increase to the size.  Because of that, I prefer to leave it in the implementation.

src/hash_map.c
clalancette: Both very good points.  I'll fix those up.

clalancette: Fixed in b23ad53

fujitatomoya:

src/hash_map.c
```diff
@@ -255,6 +255,13 @@ rcutils_hash_map_init(
...
+  if ((initial_capacity % 2) != 0) {
```
iuhilnehc-ynos, line 260: Is it to handle `multiple-of-two` or `power-of-two`?
I think `initial_capacity % 2` can only handle the former. If so, maybe we can use `initial_capacity & ~(SIZE_MAX - 1)` instead.

clalancette: Very good point.  I'll fix it like you suggest.

clalancette: That's a very good point, I totally didn't think about this properly.

That said, I don't think your solution is correct.  If I run it locally, it still returns `true` for other multiple-of-two, like 6.  What does seem to work properly is the following:

```
if ((initial_capacity & (initial_capacity - 1)) == 0) {
```

I'm going to change the code to that.

clalancette: And see 404836c now.

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17138/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11685/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17560/)

iuhilnehc-ynos (approved):

clalancette:
Thanks for the detailed reviews!  With approvals and CI being green, I'm going to go ahead and merge this one.

include/rcutils/types/hash_map.h
```diff
@@ -156,7 +156,7 @@ rcutils_get_zero_initialized_hash_map();
  * ```
  *
  * \param[inout] hash_map rcutils_hash_map_t to be initialized
- * \param[in] initial_capacity the amount of initial capacity for the hash_map
+ * \param[in] initial_capacity the amount of initial capacity for the hash_map - this must be greater than zero and a power of 2
```
wjwwood, line 159: I think this is a silly thing to do, now everyone that calls it has to do this calculation (it isn't trivial btw). Instead this function should take this as a request and document that it's rounding up to the nearest power of 2, not require the user to do this...

See what @methylDragon is having to do in his new pr:

https://github.com/ros2/rosidl/pull/728/files#diff-f7fe239aa4ebdf12a882356e1ef293b766aced118d5a9b59e487efb08f08c41aR45-R57

btw, the example, a few lines above uses 10, which isn't a power of two 🤦‍♂️

clalancette: > I think this is a silly thing to do, now everyone that calls it has to do this calculation (it isn't trivial btw). Instead this function should take this as a request and document that it's rounding up to the nearest power of 2, not require the user to do this...

That's fair.  There were only two users of this before, and they were both trivial to fix, so I just did it this way.  But I'd be fine modifying this so it does the calculation internally.

> btw, the example, a few lines above uses 10, which isn't a power of two man_facepalming

I just missed that example in the documentation.  All of the code does it properly.  I can fix that as well.

clalancette: See https://github.com/ros2/rcutils/pull/420

PR #375, opened 2022-08-17
Fix a bug in hash_map_get_next_key_and_data.

clalancette:
What was happening was that we were incorrectly using the
previous_key for finding our last "spot" in the hash_map
for iteration.  That's because of a silly typo.

This wasn't picked up in the tests because the tests also
had a bug; they were reusing the same key pointer for the
previous and the current key, and because of that they
didn't pick up on the bug.  Fix both of these issues, which
fixes iteration over the hash map.

### DIFF
--- a/src/hash_map.c
+++ b/src/hash_map.c
@@ -539,7 +539,8 @@ rcutils_hash_map_get_next_key_and_data(
   }
 
   if (NULL != previous_key) {
-    already_exists = hash_map_find(hash_map, key, &key_hash, &map_index, &bucket_index, &entry);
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17142/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11689/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17566/)

fujitatomoya (approved):
sorry i missed it, lgtm.

PR #381, opened 2022-08-29
Optimize rcutils_logging_get_logger_effective_level()

clalancette:
This function is the most expensive part of
rcutils_logging_logger_is_enabled_for(), which is called on every
RCUTILS_LOG_* invocation.
    
We notice a couple of things:
    
1. The current name -> logger level lookup is using a "string map"
structure.  A "string map" isn't a map at all, but is really a linear
array of strings.  Thus searching it can be slow when there are
a lot of strings in the map.  Since this is a global map, this
can happen when many ROS 2 nodes are loaded in to the same process.
    
2. When looking up the severity of a particular name, we need to
check the full name, plus any ancestors (separated by '.') to find
the severity level.  This requires a bunch of work, including
copying a string, running strlen on it, searching through it for
periods, etc.  This can be expensive, and in the common case it isn't
needed; the fully-qualified string is usually in the map already.
    
To fix both of these, we switch to a hash map, and look up the
current logger level via the hash map.  If the full logger name
isn't in the hash map, we fall back to the slow path where we
do the additional work of finding the severity via the hierarchy.
Note that even in this slow path, once we've computed the severity
level we place that in the map as well so that subsequent lookups
will take the fast path.
    
The benchmarks I ran on this change give very good improvements
for most cases.  For cases where there are multiple set logger
names, and those logger names have ancestors, this change is
~9x faster than the current implementation.  For cases where there
are multiple set logger names and those logger names have no
ancestors (one of the most common cases in ROS 2), this change is ~3x
faster than the current implementation.  For cases where we
have a single logger name (another common case), this change is
~2x faster than the current implementation.
    
Finally, note that rcutils_logging_set_logger_level() is
more expensive than before, due to having to traverse the hash map
to find other logger names in the hierarchy.  Since we don't expect
users to change logger levels often, this tradeoff seems worth it
to speed up lookups (which are called for every log message).
    
This should solve https://github.com/ros2/rcutils/issues/364 (at least, as much as it can be solved given the current design).

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -414,6 +414,21 @@ static const char * copy_from_orig(
   return logging_output->buffer;
 }
 
+static bool add_handler(token_handler handler, size_t start_offset, size_t end_offset)
+{
+  g_handlers[g_num_log_msg_handlers].handler = handler;
+  g_handlers[g_num_log_msg_handlers].start_offset = start_offset;
+  g_handlers[g_num_log_msg_handlers].end_offset = end_offset;
+  if (g_num_log_msg_handlers >= sizeof(g_handlers) - 1) {
+    RCUTILS_SET_ERROR_MSG("Too many substitutions in the logging output format string; truncating");
+    return false;
+  }
+
+  g_num_log_msg_handlers++;
+
+  return true;
+}
+
 static void parse_and_create_handlers_list(void)
 {
   // Process the format string looking for known tokens.
@@ -435,17 +450,10 @@ static void parse_and_create_handlers_list(void)
     if (chars_to_start_delim > 0) {  // there is stuff before a token start delimiter
       size_t chars_to_copy = chars_to_start_delim >
         remaining_chars ? remaining_chars : chars_to_start_delim;
-      g_handlers[g_num_log_msg_handlers].handler = copy_from_orig;
-      g_handlers[g_num_log_msg_handlers].start_offset = i;
-      g_handlers[g_num_log_msg_handlers].end_offset = i + chars_to_copy;
-      if (g_num_log_msg_handlers >= sizeof(g_handlers) - 1) {
-        RCUTILS_SAFE_FWRITE_TO_STDERR(
-          "Too many substitutions in the logging output format string; truncating");
-        rcutils_reset_error();
-        RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
+      if (!add_handler(copy_from_orig, i, i + chars_to_copy)) {
+        // The error was already set by add_handler
         return;
       }
-      g_num_log_msg_handlers++;
 
       i += chars_to_copy;
       if (i >= size) {  // perhaps no start delimiter was found
@@ -466,17 +474,10 @@ static void parse_and_create_handlers_list(void)
     if (chars_to_end_delim > remaining_chars) {
       // No end delimiters found in the remainder of the format string;
       // there won't be any more tokens so shortcut the rest of the checking.
-      g_handlers[g_num_log_msg_handlers].handler = copy_from_orig;
-      g_handlers[g_num_log_msg_handlers].start_offset = i;
-      g_handlers[g_num_log_msg_handlers].end_offset = i + remaining_chars;
-      if (g_num_log_msg_handlers >= sizeof(g_handlers) - 1) {
-        RCUTILS_SAFE_FWRITE_TO_STDERR(
-          "Too many substitutions in the logging output format string; truncating");
-        rcutils_reset_error();
-        RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
+      if (!add_handler(copy_from_orig, i, i + remaining_chars)) {
+        // The error was already set by add_handler
         return;
       }
-      g_num_log_msg_handlers++;
       break;
     }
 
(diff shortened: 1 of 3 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17182/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17605/)

clalancette:
New CI after the latest fix:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17183/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11725/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17606/)

clalancette:
Another CI with the latest Windows fix:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17191/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11732/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17613/)

clalancette:

ivanpauno:
I have a few questions before reviewing 

> When looking up the severity of a particular name, we need to
check the full name, plus any ancestors (separated by '.') to find
the severity level. This requires a bunch of work, including
copying a string, running strlen on it, searching through it for
periods, etc. This can be expensive, and in the common case it isn't
needed; the fully-qualified string is usually in the map already.

The parent logging severity is only queried if there wasn't a severity directly set for this logger, right?
If looking up for a parent severity is needed, it would be nice to set the child logger severity it in the map, so the next time the only thing needed is a query.
The only tricky thing is that if someone changes the parent logger severity, the child should also change in that case (as the severity was actually inherited from the parent), so to make sure that happens we will need to store some extra information.

---

IMO, everything would be more efficient if we would have a logger object, instead of only using a name.
It may be a pain to change that now, but most people are using the rclcpp/rclpy logging functions instead of this, and those already have a logger object.

clalancette:
> The parent logging severity is only queried if there wasn't a severity directly set for this logger, right?

That's correct.

> If looking up for a parent severity is needed, it would be nice to set the child logger severity it in the map, so the next time the only thing needed is a query.

Yep!  That's what this PR does now; if the parent lookup is needed, then once it is calculated the child is added into the map for subsequent lookups (see the code in `rcutils_logging_get_logger_effective_level`).

> The only tricky thing is that if someone changes the parent logger severity, the child should also change in that case (as the severity was actually inherited from the parent), so to make sure that happens we will need to store some extra information.

Right.  This PR also addresses that by walking the entire hash map when loggers change severity, looking for children of the current name being changed.  For any ones that haven't been explicitly set, it drops them from the hash map until they are used again (see the code in `rcutils_logging_set_logger_level`).

> IMO, everything would be more efficient if we would have a logger object, instead of only using a name.
> It may be a pain to change that now, but most people are using the rclcpp/rclpy logging functions instead of this, and those already have a logger object.

Yes, agreed.  The performance with this PR is the best we can do (I think) with a strictly string-based hierarchy like we currently have.  I think it is valuable to get this in (as it will help performance today), and then something like https://github.com/ros2/rcl_logging/issues/92 may make this obsolete.

src/logging.c
```diff
@@ -761,6 +768,51 @@ int rcutils_logging_get_logger_level(const char * name)
...
+    // Don't report an error to the error handling machinery; some uses of this function are for
+    // caching so this is not necessarily fatal.
+    return RCUTILS_RET_ERROR;
+  }
+
+  if (set_by_user) {
+    // If the level was set by the user (rather than optimistically cached by the implementation),
+    // mark it here.  When we purge the cache, we'll make sure not to purge these.  The mark we
+    // use is setting the bottom bit; since our levels are 0, 10, 20, 30, 40, and 50, this works.
+    level |= 0x1;
```
ivanpauno, line 786: we should add a note in the RCUTILS_LOG_SEVERITY that the last bit isn't supposed to be used.
Or maybe use a struct and store both the log level and the flag independently.

src/logging.c
```diff
@@ -805,30 +853,102 @@ int rcutils_logging_get_logger_effective_level(const char * name)
...
     }
+    // If the severity is UNSET, then we go through the slow path and try to find a parent logger
+    // that is set.  Failing that, we fall back to the default logger level.
+  } else if (ret != RCUTILS_RET_NOT_FOUND) {
+    // The error message was already set by get_severity_level
+    return -1;
+  }
+
+  // Since we didn't find the name in the fast path, fall back to the slow path where we break the
+  // string into substrings based on dots and look for any part that matches.
```
ivanpauno, line 893: if we want to further optimize this, one thing we could do is to not only store the severity/created_by_user flag of each logger, but also keep a list of the children loggers.

When we create a logger, we also create the parent (and grandparents) if needed, and the new logger is added to the children list.
When the severity of a logger is changed, the severity of the children is also checked and set.
(in the children list, we don't need to keep all children only the ones that inheriting the min severity)

No need to do that in this PR, but I think something like that would work.
Basically, we're having an object for each logger with all the required info, and using a global map to match the objects to names.

ivanpauno (approved):
LGTM!

I have some minor comments

src/logging.c
clalancette: Let me know if you think we should change to a structure.  I avoided that just for the efficiency, but I can do it if you think that is better.

src/logging.c
clalancette: Yeah, I thought about this approach.

As you say, building a tree like this could significantly speed up the setting of levels.  But I worry that the cost would be slower lookup of levels, which is the thing we really want to optimize here.

Honestly, if we are going to optimize further, I would probably concentrate on your earlier idea of switching to a logger object rather than a string.  In my mind that is likely to be the fastest for both getting and setting.  All of that said, I'm not planning on doing more optimizations here in the near future.

clalancette:
@ivanpauno Thanks for the review!  I think I've addressed all of your comments; please take another look.

src/logging.c
ivanpauno: > But I worry that the cost would be slower lookup of levels, which is the thing we really want to optimize here.

How the severity level is looked up doesn't change, it's done in the same way.
The tree is used when setting a severity level, instead of iterating over all loggers in the map.
The tree is constructed when a logger is created for the first time(*), by checking all possible parents and adding the new child to the list (if the parent doesn't exist, it's added to the map as well).

Though, the most important thing is the time it takes to query a severity level, and this PR already improves that as much as it seems possible (without breaking API).
So maybe my suggestion will make the code more complicated and improve something that it's not that important.

(*) I think loggers aren't actually "created", I mean when they're queried or their severity level is set for the first time.

> Honestly, if we are going to optimize further, I would probably concentrate on your earlier idea of switching to a logger object rather than a string. In my mind that is likely to be the fastest for both getting and setting.

Agreed!
I was suggesting ideas if we still want to optimize without breaking API.

> All of that said, I'm not planning on doing more optimizations here in the near future.

Yes, that's not needed right now. This is great work!!!

ivanpauno (approved):
LGTM! Great work!!

clalancette:
One more CI on the latest changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17240/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11780/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17700/)

clalancette:
The Windows CI warning is unrelated to this change, and everything else is green.

Going ahead and merging this, thanks for the review!

PR #382, opened 2022-08-29
add portable nonnull macros

wjwwood:
I found this useful when implementing a helper header for the new `<memory_resource>` header in c++17, because some implementations are incomplete so I had to implement a few functions it helped to mark some parameters and return types as non-null for the compiler.

I will link that pr (for the helper header) here when I have it up.

### DIFF
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -145,7 +145,7 @@ extern "C"
 // Provide the compiler a hint about an argument being nonnull when possible.
 #ifndef _WIN32
 # define RCUTILS_NONNULL __attribute__((__nonnull__))
-# define RCUTILS_NONNULL_ARGS(...) __attribute__((__nonnull__ (__VA_ARGS__)))
+# define RCUTILS_NONNULL_ARGS(...) __attribute__((__nonnull__(__VA_ARGS__)))
 # define RCUTILS_RETURNS_NONNULL __attribute__((__returns_nonnull__))
 # define RCUTILS_HAS_NONNULL 1
 #else
--- a/include/rcutils/macros.h
+++ b/include/rcutils/macros.h
@@ -30,7 +30,7 @@ extern "C"
 #define RCUTILS_WARN_UNUSED _Check_return_
 #endif
 
-/// @cond Doxygen_Suppress
+/// \cond Doxygen_Suppress
 // This block either sets RCUTILS_THREAD_LOCAL or RCUTILS_THREAD_LOCAL_PTHREAD.
 #if defined _WIN32 || defined __CYGWIN__
 // Windows or Cygwin
@@ -140,6 +140,19 @@ extern "C"
  * No op since Windows doesn't support providing branch prediction information.
  */
 # define RCUTILS_UNLIKELY(x) (x)
+#endif  // _WIN32
+
+// Provide the compiler a hint about an argument being nonnull when possible.
+#ifndef _WIN32
+# define RCUTILS_NONNULL __attribute__((__nonnull__))
+# define RCUTILS_NONNULL_ARGS(...) __attribute__((__nonnull__ (__VA_ARGS__)))
+# define RCUTILS_RETURNS_NONNULL __attribute__((__returns_nonnull__))
+# define RCUTILS_HAS_NONNULL 1
+#else
+# define RCUTILS_NONNULL
+# define RCUTILS_NONNULL_ARGS(...)
+# define RCUTILS_RETURNS_NONNULL
+# define RCUTILS_HAS_NONNULL 0
 #endif  // _WIN32
 
 #if defined RCUTILS_ENABLE_FAULT_INJECTION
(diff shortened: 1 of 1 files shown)


clalancette:

include/rcutils/macros.h
```diff
@@ -142,6 +142,19 @@ extern "C"
 # define RCUTILS_UNLIKELY(x) (x)
 #endif  // _WIN32
 
+// Provide the compiler a hint about an argument being nonnull when possible.
+#ifndef _WIN32
```
clalancette, line 146: So it does look like the NONNULL attributes are supported in gcc and clang, so that's good.  My concern here is with other, non-gcc and clang compilers (maybe what QNX is using, for instance?).  Instead of making this conditional on `!WIN32`, should we make it `GCC || CLANG`?  Not sure.

include/rcutils/macros.h
```diff
@@ -142,6 +142,19 @@ extern "C"
...
+# define RCUTILS_NONNULL __attribute__((__nonnull__))
+# define RCUTILS_NONNULL_ARGS(...) __attribute__((__nonnull__ (__VA_ARGS__)))
+# define RCUTILS_RETURNS_NONNULL __attribute__((__returns_nonnull__))
+# define RCUTILS_HAS_NONNULL 1
```
clalancette, line 150: What's the reason for this particular macro?

clalancette (commented):
I've left a few things to think about.

clalancette:
(also, uncrustify is unhappy)

include/rcutils/macros.h
fujitatomoya: I do not think QNX supports it, i might be wrong. and this is compiler specific.

include/rcutils/macros.h
```diff
@@ -30,7 +30,7 @@ extern "C"
 #define RCUTILS_WARN_UNUSED _Check_return_
 #endif
 
-/// @cond Doxygen_Suppress
+/// \cond Doxygen_Suppress
```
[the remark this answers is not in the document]
wjwwood: Yeah, I didn't mean to add that to this pr, I was going to make another pr to fix all of those.

include/rcutils/macros.h
wjwwood: Yes, but I can't know what does and doesn't work on other compilers without testing. I believe this will work on all the compilers we test. If someone is testing with QNX, then they can report it and we can fix it.

My concern with doing something like `GCC || CLANG` is that I will miss some match, and future or older versions of gcc or clang may not match this statement where it should.

This (using `#ifndef _WIN32`) is how we do things in a lot of places already.

It basically boils down to:

- this is how we expect most compilers to work
- except windows... 
- potentially more exceptions in the future

Until I know what the alternative should be (for other compilers) I leave them out for now. 🤷

include/rcutils/macros.h
wjwwood: So that you can write code that differs based on whether or not the compiler supports this feature. I needed it for a pull request where I'm using these macros.

Right now the code lives here, but I might move it to something like `rcpputils` later: https://github.com/space-ros/demos/blob/e79d1a48cca84d383b3433e389799d9e5b7d3c87/space_ros_memory_allocation_demo/include/space_ros_memory_allocation_demo/memory_resource_helper.hpp#L128-L136

include/rcutils/macros.h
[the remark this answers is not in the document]
wjwwood: [328245b](https://github.com/ros2/rcutils/pull/382/commits/328245bfbaf3af6288aef507c523cdfacd5e8263)

include/rcutils/macros.h
wjwwood: Based on my comment above, and after thinking on it more, I would like to keep it as-is, but please let me know what you guys think.

If you think we really should change it, do you know of an example where we condition on gcc/clang? I was looking in our code base and didn't see it. There are lots of ways suggested online I just don't know which would be best.

fujitatomoya (approved):
good to go.

include/rcutils/macros.h
clalancette: Looking around a bit, I think that QNX uses gcc under the hood, so it probably understands these attributes.  I also looked around the codebase a bit, and I saw no instances where we were special-casing `__attribute__` for QNX.  The only places I saw special-casing for QNX is around libc-style differences.

But QNX was just an example.  Thinking about it more, there are really only 3 major compilers out there: gcc, clang, and MSVC.  While there are other minor compilers, I doubt this would be the only problem they would have.

All of that is to say; I'm fine with this as-is.

clalancette (approved):
Looks good with green CI.

include/rcutils/macros.h
wjwwood: Ok, thanks.

I kind of agree, if you want to introduce a new compiler (outside of gcc/clang and msvc) you'll have lots of things to fix. I don't necessarily want to add to that, but on the other hand I don't think I can make it better at the moment.

PR #383, opened 2022-09-08
time_unix: namespace zephyr headers

uLipe:
After version 3, the Zephyr RTOS requires to namespace all its headers with zephyr before including any other header file that belongs to it, this PR checks the version o major number of the kernel and when above 3, namespace the Zephyr specific posix timer implementation.

### DIFF
--- a/src/time_unix.c
+++ b/src/time_unix.c
@@ -30,12 +30,17 @@ extern "C"
 #include <math.h>
 
 #if defined(__ZEPHYR__)
-#include <posix/time.h>  //  Points to Zephyr toolchain posix time implementation
+#include <version.h>
+#if KERNELVERSION >= ZEPHYR_VERSION(3, 1, 0)
+#include <zephyr/posix/time.h>  //  Points to Zephyr toolchain posix time implementation
 #else
+#include <posix/time.h>  //  Points to Zephyr toolchain posix time implementation
+#endif
+#else  //  #if KERNELVERSION >= ZEPHYR_VERSION(3, 1, 0)
 #include <time.h>
 #endif  //  defined(__ZEPHYR__)
-#include <unistd.h>
 
+#include <unistd.h>
 #include "./common.h"
 #include "rcutils/allocator.h"
 #include "rcutils/error_handling.h"


uLipe:
CC: @pablogs9

uLipe:
@clalancette  @ahcorde PTAL

src/time_unix.c
```diff
@@ -30,7 +30,11 @@ extern "C"
 #include <math.h>
 
 #if defined(__ZEPHYR__)
+#if KERNEL_VERSION_MAJOR >= 3
```
pablogs9, line 33: This definition is only provided when building with https://github.com/micro-ROS/micro_ros_zephyr_module right? Is it not Zephyr standard?

uLipe: @pablogs9 , you are right, there is a "zephyrzed" way of handling this, please check the new one.

uLipe:
@pablogs9 @clalancette @ahcorde  PTAL again.

src/time_unix.c
fujitatomoya, line 33: I might be mistaken, but this should be v3.1.0 or later? see https://github.com/zephyrproject-rtos/zephyr/releases/tag/zephyr-v3.1.0

this could break with `v3.0.0`? it would be better to check minor kernel version too?

uLipe: Oh, you are right, If I'm not wrong it should be backward compatible to version 3.0.0, but it seems the best way is to check against the exact version.

Will update that.

uLipe:
@fujitatomoya PTAL again also.

pablogs9:
The new way to check version LGTM

clalancette (approved):
This is pretty messy, but if this makes it work, it is OK (actually this whole file is messy, but that's a separate problem).

Will run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17424/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11950/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17927/)

uLipe:
@clalancette , I agree on you, the solution is not the most elegant but needed to make this supported across Zephyr versions 2.7 and 3.1

the header namespacing in Zephyr is introducing the same problem in others external modules .

fujitatomoya (approved):

fujitatomoya:
good to go with green CI.

clalancette:
Basically by definition, this change can't have made the one change on Windows fail.  That one is flaky every once in a while anyway.  Going ahead and merging this.

pablogs9:
PTAL here: https://github.com/micro-ROS/micro_ros_zephyr_module/pull/85#issuecomment-1264971811
Zephyr 2.7.2 failing to find `<zephyr/posix/time.h>`

uLipe:
@pablogs9 , reason and fix here: https://github.com/ros2/rcutils/pull/390

PR #384, opened 2022-09-17
Export -latomic even if BUILD_TESTING is disabled

roehling:
On some architectures, libatomic is needed and should not be dependent on tests being enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,14 @@ endif()
 
 target_link_libraries(${PROJECT_NAME} ${CMAKE_DL_LIBS})
 
+check_library_exists(atomic __atomic_load_8 "" HAVE_LIBATOMICS)
+
+if(HAVE_LIBATOMICS AND NOT WIN32)
+  # Exporting link flag since it won't pass ament_export_libraries() existance check
+  ament_export_link_flags("-latomic")
+  target_link_libraries(${PROJECT_NAME} atomic)
+endif()
+
 # Needed if pthread is used for thread local storage.
 if(IOS AND IOS_SDK_VERSION LESS 10.0)
   ament_export_libraries(pthread)
@@ -156,13 +164,6 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 
   find_package(launch_testing_ament_cmake REQUIRED)
-
-  check_library_exists(atomic __atomic_load_8 "" HAVE_LIBATOMICS)
-
-  if(HAVE_LIBATOMICS AND NOT WIN32)
-    # Exporting link flag since it won't pass ament_export_libraries() existance check
-    ament_export_link_flags("-latomic")
-  endif()
 
   if(ament_cmake_cppcheck_FOUND)
     ament_cppcheck(


fujitatomoya:
@roehling thanks for the PR, can you explain more details why we need this and what the actual problem you have? i think that would make it easier to understand for everyone.

roehling:
The atomic operations are used by `src/testing/fault_injection.c`, which is compiled into the library unconditionally and may generate symbol references to `libatomic` on some architectures. However, as of now, the required `-latomic` linker flag is added only if `BUILD_TESTING` is enabled. This means that users may need to link against `libatomic` themselves, even if they do not use atomics at all, just because their `rcutils` library happens to be compiled with `BUILD_TESTING` disabled.

fujitatomoya:
@roehling appreciate the explanation.

> The atomic operations are used by src/testing/fault_injection.c, which is compiled into the library unconditionally and may generate symbol references to libatomic on some architectures.

is this because of the following? if that so, probably getting rid of unnecessary dependency would be solution?

https://github.com/ros2/rcutils/blob/9cf42fbb77caa4a729db7fbf6a3087f3e6dee439/include/rcutils/error_handling.h#L39

and, this should be no problem, since it includes only when it is test build.

https://github.com/ros2/rcutils/blob/9cf42fbb77caa4a729db7fbf6a3087f3e6dee439/include/rcutils/macros.h#L145-L146

thanks,

roehling:
No, it is not.

It is because `src/testing/fault_injection.c` is compiled into the rcutils library here: https://github.com/ros2/rcutils/blob/9cf42fbb77caa4a729db7fbf6a3087f3e6dee439/CMakeLists.txt#L74

And said file then uses `rcutils_atomic_` operations, which in turn generate symbol references to the `atomic` library.

I guess, if you wanted to avoid the unconditional dependency on the `atomic` library, you could also make `fault_injection.c` conditional, for example with the generator expression  `$<$<BOOL:BUILD_TESTING>:src/testing/fault_injection.c>`, but that means you probably should not install the corresponding headers either.

I believe my PR is less intrusive,  as GCC normally adds `--as-needed` to the linker call, so the `atomic` library will only be linked if it is actually needed.

fujitatomoya:
> but that means you probably should not install the corresponding headers either.

yeah right, currently it include this header, e.g) `/opt/ros/humble/include/rcutils/rcutils/testing/fault_injection.h`

> I believe my PR is less intrusive, as GCC normally adds --as-needed to the linker call, so the atomic library will only be linked if it is actually needed.

agree.

fujitatomoya (approved):

fujitatomoya:
@clalancette could you take a look?

lopsided98:
I just ran into this problem while cross-compiling to armv6, and came up with pretty much the same solution. I did find that the `ament_export_link_flags("-latomic")` call doesn't appear to be necessary, at least for the packages that I've built (mavros and all its dependencies, which should include all the core ROS packages).

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17933/)

wentasah:
If anyone needs it, the commit from this PR updated to current rolling (there are conflicts) is available in https://github.com/wentasah/rcutils/commit/acdcf805dfd8d3cf77f269ef280077d4226e6e4e.

fujitatomoya:
@Mergifyio rebase

fujitatomoya:
@roehling since i could not resolve the conflict on your dev branch, i just went ahead to create another PR with adding your author signature. it would be appreciated if you can take a look.

PR #385, opened 2022-09-20
Script empy

yashi:
Hi,

This PR simplifies empy invocation.  Hope you like it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,39 +80,26 @@ set_source_files_properties(
   ${rcutils_sources}
   PROPERTIES language "C")
 
-# "watch" template/inputs for changes
-configure_file(
-  "resource/logging_macros.h.em"
-  "logging_macros.h.em.watch"
-  COPYONLY)
-configure_file(
-  "rcutils/logging.py"
-  "logging.py.watch"
-  COPYONLY)
 # generate header with logging macros
-set(rcutils_module_path ${CMAKE_CURRENT_SOURCE_DIR})
-set(python_code
-  "import em"  # implicitly added ; between python statements due to CMake list
-  "\
-em.invoke( \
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
Thanks for the PR.  I have a couple of pieces of feedback here:

1.  Please don't include the changes to the `.gitignore` file; if you'd like that, we can do it separately.
2. As it stands, I don't believe this is going to work on Windows.  That's because Windows does not let you directly invoke a Python script, and instead you always need to do `python my_script.py`.

But before you go trying to fix this on Windows, my question is: what problem is this solving?  That is, what problem did you have that caused you to try to refactor this?

yashi:
Hi @clalancette, thanks for your feed back. I'm trying to convert ros2 libraries to Zephyr RTOS modules.  I know there is https://github.com/micro-ROS/micro_ros_zephyr_module but it has some problems. By creating proper Zephyr modules, I'm hoping to solve the problems we have with it.

While testing them I found that the current build generates files in the source dir. (I should have mentioned this in the commit log. Will fix.).   And looked at CMakeList.txt and found that it's a bit hard to follow.

Hope this explains my situation.

yashi:
Updated to run `python3 scripts/empy` for Windows.
Also,
- removed `message()` leftover from my debugging session
- removed about `/usr/bin/empy` from the commit log
- removed `.gitignore` change. (I didn't know ROS uses [squash commit strategy](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#pull-requests). Thanks @clalancette)

Regarding " the current build generates files in the source dir."  This was my mistake.  It was `micro_ros_zephyr_module` which generated files under source dir.  rcutils doesn't do anything bad.

PR #386, opened 2022-09-21
Restrict overmatching MACH ifdef to only trigger on OSX and Mach

Zopolis4:
As said on line 43, this check is only intended for OSX.
```
 // Assume clock_get_time is available on OS X.
```

Hurd uses a different version of Mach that does not have `mach/clock.h`.
https://www.gnu.org/software/hurd/hurd/porting/guidelines.html
```
#ifdef __MACH__
Some applications put Apple Darwin-specific code inside #ifdef __MACH__ guards. Such guard is clearly not enough, since not only Apple uses Mach as a kernel. This should be replaced by #if defined(__MACH__) && defined(__APPLE__)
```

Signed-off-by: Maximilian Downey Twiss <creatorsmithmdt@gmail.com>

### DIFF
--- a/src/time_unix.c
+++ b/src/time_unix.c
@@ -23,10 +23,10 @@ extern "C"
 
 #include "rcutils/time.h"
 
-#if defined(__MACH__)
+#if defined(__MACH__) && defined(__APPLE__)
 #include <mach/clock.h>
 #include <mach/mach.h>
-#endif  // defined(__MACH__)
+#endif  // defined(__MACH__) && defined(__APPLE__)
 #include <math.h>
 
 #if defined(__ZEPHYR__)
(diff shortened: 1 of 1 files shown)


src/time_unix.c
```diff
@@ -40,13 +40,13 @@ extern "C"
 #include "rcutils/allocator.h"
 #include "rcutils/error_handling.h"
 
-#if !defined(__MACH__)  // Assume clock_get_time is available on OS X.
+#if !defined(__MACH__)  && !defined(__APPLE__)   // Assume clock_get_time is available on OS X.
```
clalancette, line 43: The linter is unhappy here.

```suggestion
#if !defined(__MACH__) && !defined(__APPLE__)   // Assume clock_get_time is available on OS X.
```

clalancette (changes requested):
One fix inline, then this will be ready for CI.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17356/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11891/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17851/)

Zopolis4:
This doesn't touch windows at all, so not sure what's going on there.

clalancette:
> This doesn't touch windows at all, so not sure what's going on there.

We've been having some instability with the Windows workers, it looks like we ran into it.  I've kicked off a new job.

clalancette:
The single failing test on Windows has nothing to do with this, so I'll go ahead and merge it.

PR #387, opened 2022-09-21
.gitignore: Add Emacs backup file pattern

yashi:
Emacs uses *~ as a backup file.  Ignore them.

Signed-off-by: Yasushi SHOJI <yashi@spacecubics.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+*~
 __pycache__
 doc_output


fujitatomoya (commented):
If everyone commits environment-specific rules in local .gitignore, we will have a really long list to maintain. I do not thinks this is an good idea.

How about adding environment-specific rule to you global setting?

yashi:
I agree that adding truly environment specific patterns, such as `builddir` to `.gitignore` is nonsense.  But adding editor specifics ain't.  Relaying on users' setting isn't best practice IMHO.

I don't think it'll get that long, either.  Long `.gitignore`s are rare. I have over 500 open source git repos under `~/src`. and the situation is
```
$ for i in  ~/src/*/.gitignore; do awk 'END{print NR}' $i; done | st-console --N --min --max --q1 --q2 --q3 --sd
N	min	q1	median	q3	max	stddev
532	1	6	23	58	763	74.8845
```
But even you get that long, nobody cares.

Anyway, that's my idea.  Please reject it if you don't want it.

fujitatomoya:
yeah that is alright, we can keep this open to get more feedback. i just do not think this is a good approach. (personally, i use emacs, and set this into global config.)

clalancette:
> yeah that is alright, we can keep this open to get more feedback. i just do not think this is a good approach. (personally, i use emacs, and set this into global config.)

Personally, I dislike per-repository `.gitignore` files.  They obscure what is actually going on in the repository, so I often have to resort to additional flags to `git status`.  So I'm not in favor of expanding it either, and in fact I would be for totally deleting it.

fujitatomoya:
agree, something i do not understand why we need this editor setting is, if we set `editor config`(environment or not config, whatever...) to the global config, **it just works**, and we do not need to care about this for each repository.
Why we need to push these setting to each repository, it already works for you in better way... just cannot understand that.

mjcarroll:
Yes, this is what a global ignore file is for.  To set configurations for your specific setup, I would recommend:

```
git config --global core.excludesfile ~/.gitignore
```

Change to your liking on where to store your global ignores.

fujitatomoya:
I will go ahead to close this one, feel free to reopen if necessary.

PR #389, opened 2022-09-30
Cleanup time_unix.c

clalancette:
The file was pretty messy.  Cleanups:

1.  Remove totally unnecessary #ifdef __cplusplus; that is only necessary in headers.
2.  Remove unnecessary include to ./common.h.
3.  Change WOULD_BE_NEGATIVE to an inline function, which will at least give us some type checking.
4.  Get rid of unnecessary duplicated code for macOS.  Most of the time the code was exactly the same.
5.  Make sure to check the return value from clock_gettime.

The diff is pretty hard to read, but it makes a lot more sense if you just look at the resulting file.

### DIFF
--- a/src/time_unix.c
+++ b/src/time_unix.c
@@ -16,11 +16,6 @@
 # error time_unix.c is not intended to be used with win32 based systems
 #endif  // defined(_WIN32)
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
-
 #include "rcutils/time.h"
 
 #if defined(__MACH__) && defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17427/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11953/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17935/)

fujitatomoya (approved):

PR #391, opened 2022-10-12
Fix memory leak when adding the same key to the logger hash map multiple times

jacobperron:
The first time a key is added the hash map will retain a copy of the pointer and later will deallocate the memory during shutdown. Subsequent times the same key is added to the hash map, it will NOT retain a copy of the new pointer, and hence not deallocate the memory during shutdown.

This fixes the issue by checking if we're adding an entry for an existing key and if so, deallocate the key.

Previously, ASAN was reporting memory leaks: https://ci.ros2.org/view/nightly/job/nightly_linux_address_sanitizer/1287/
With this change, we no longer see the memory leak reported: https://ci.ros2.org/job/ci_linux/17486/

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -771,13 +771,20 @@ int rcutils_logging_get_logger_level(const char * name)
 
 static rcutils_ret_t add_key_to_hash_map(const char * name, int level, bool set_by_user)
 {
-  // Copy the name that we will store, as there is no guarantee that the caller will keep it around.
+  const char * copy_name = name;
+  // Check if key already exists, to avoid extra memory allocation
+  // If the key already exists, then rcutils_hash_map_set will not maintain the key we give it,
+  // so we do not need to copy the name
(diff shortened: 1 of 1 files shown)


jacobperron:
Makes sense to me. I've updated the logic to avoid the extra allocation/deallocation: [cf76443](https://github.com/ros2/rcutils/pull/391/commits/cf764430450d2b99b4d6c457a9684150569979cd)

jacobperron:
ASAN: [Build Status](https://ci.ros2.org/job/ci_linux/17490/)

CI (testing above rcutils):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17491/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12016/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18037/)

fujitatomoya (approved):
lgtm 👍

PR #392, opened 2022-10-13
Upgrade rcutils to C++17.

clalancette:
This upgrade was more involved than just increasing the CMAKE_CXX_STANDARD version.  In C++17, apparently 'noexcept' became part of the type system.  This is problematic because some functions that we use from the underlying libc, like 'stat', actually specify nothrow in their function declaration. This causes the type system to reject our mock shims, since the types are no longer totally compatible.

To work around this, introduce a 'remove_noexcept' type into the system which effectively removes the 'noexcept' specifier for any function that it uses.  With that in place, everything here works with C++17.  See the comments in remove_noexcept.hpp for details on the provenance of that code.

The motivation for upgrading for C++17 is to fix a memory leak down in https://github.com/osrf/osrf_testing_tools_cpp , which will require C++17 support in this package and a few others.  That will come separately.

Opening as a draft pull request for now, as we still need to figure out the licensing on remove_except.hpp.  @monkey0506, if you would be willing to explicitly license that code as Apache 2.0, MIT, or BSD, we can include it here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,9 @@ project(rcutils)
 if(NOT CMAKE_C_STANDARD)
   set(CMAKE_C_STANDARD 11)
 endif()
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 include(CheckLibraryExists)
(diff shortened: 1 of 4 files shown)


clalancette:
> Hi. Sorry for not replying to your email sooner.

No problem, thanks for the response.

> I've updated the link on the Stack Overflow answer:
> https://wandbox.org/permlink/GUqJFwMnczKEKlzX
> Attribution is appreciated, but honestly this is mostly boilerplate anyway! Hope that helps.

Excellent.  Given the text of the WTFPL license, I'm going to go ahead and relicense the code to Apache 2.0 and attribute it to you.  That way it will be compatible with the rest of our code.

Thanks again, this is very helpful.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17500/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12025/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18048/)

mjcarroll (approved):

clalancette:
Despite the fact that the Windows CI has aged out, it was successful when I ran it earlier.  So I'm going to go ahead and merge this.

PR #393, opened 2022-10-14
regression of thread-safety for logging macros

wjwwood:
This came up in one of our projects using rolling and they noticed an uptick in crashes originating in logging calls.

It turns out it is related to optimizations made in https://github.com/ros2/rcutils/pull/381, and is only triggered in specific scenarios, namely when some loggers have had their severity set by the user, but not all the ones being called. This triggers a code path that was part of https://github.com/ros2/rcutils/pull/381 which can cause a global hash map to be added to, and sometimes grown, and this seems to be the root of many of the crashes, though it crashes in many different ways.

In this pr I added a regression test that catches these kind of problems most runs, but it isn't perfect since it is a race condition.

After some investigation and talking with @clalancette, this code block seems to be the offender:

https://github.com/ros2/rcutils/blob/3485f1bdc805ead1e9c26509479dc8c1f94b6951/src/logging.c#L946-L957

After removing that, this test passes even if I use `--gtest_repeat=1000` to run it 1000 times. Whereas it fails with that code uncommented almost every time. So I don't know if this makes the logging macros completely (conceptually) thread-safe, but it does undo this regression, it seems.

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -952,7 +952,7 @@ int rcutils_logging_get_logger_effective_level(const char * name)
       // Continue on if we failed to add the key to the hashmap.
       // This will affect performance but is not fatal.
       RCUTILS_SAFE_FWRITE_TO_STDERR(
-        "Failed to cache severity; this is not fatal but will impact performance");
+        "Failed to cache severity; this is not fatal but will impact performance\n");
     }
   }
 
--- a/src/logging.c
+++ b/src/logging.c
@@ -949,9 +949,6 @@ int rcutils_logging_get_logger_effective_level(const char * name)
   // // lookup next time.  If the severity is UNSET, we don't bother because we are going to have to
   // // walk the hierarchy next time anyway.
   // if (severity != RCUTILS_LOG_SEVERITY_UNSET) {
-  //   fprintf(
-  //     stderr, "in optimization of rcutils_logging_get_logger_effective_level('%s') -> %d\n",
-  //     name, severity);
   //   ret = add_key_to_hash_map(name, severity, false);
   //   if (ret != RCUTILS_RET_OK) {
   //     // Continue on if we failed to add the key to the hashmap.
--- a/src/logging.c
+++ b/src/logging.c
@@ -943,18 +943,23 @@ int rcutils_logging_get_logger_effective_level(const char * name)
     severity = g_rcutils_logging_default_logger_level;
   }
 
-  // If the calculated severity is anything but UNSET, we place it into the hashmap for speedier
-  // lookup next time.  If the severity is UNSET, we don't bother because we are going to have to
-  // walk the hierarchy next time anyway.
-  if (severity != RCUTILS_LOG_SEVERITY_UNSET) {
-    ret = add_key_to_hash_map(name, severity, false);
-    if (ret != RCUTILS_RET_OK) {
-      // Continue on if we failed to add the key to the hashmap.
-      // This will affect performance but is not fatal.
-      RCUTILS_SAFE_FWRITE_TO_STDERR(
-        "Failed to cache severity; this is not fatal but will impact performance\n");
-    }
-  }
+  // TODO(wjwwood): restore or replace this optimization when thread-safety is addressed
+  //   see: FIXME
+  // // If the calculated severity is anything but UNSET, we place it into the hashmap for speedier
+  // // lookup next time.  If the severity is UNSET, we don't bother because we are going to have to
+  // // walk the hierarchy next time anyway.
+  // if (severity != RCUTILS_LOG_SEVERITY_UNSET) {
+  //   fprintf(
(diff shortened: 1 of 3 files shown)


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12022/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18043/)

test/test_logging.cpp
```diff
@@ -485,3 +487,52 @@ TEST(TestLogging, test_root_logger_after_nonexistent)
...
+  // One of the loggers needs to be set "by the user" to trigger the optimization.
+  ASSERT_EQ(RCUTILS_RET_OK, rcutils_logging_set_logger_level("", RCUTILS_LOG_SEVERITY_DEBUG));
+
+  // In threads, in a loop do a log call on many different logger names.
+  // The message doesn't matter.
+  std::size_t loop_count = 10;
+  auto task = [&loop_count](std::size_t thread_number) {
+      for (std::size_t i = 0; i < loop_count; ++i) {
+        rcutils_log_location_t location = {"func", "file", 42u};
+        rcutils_log(
```
fujitatomoya, line 519: `rcutils_log` is not thread-safe, this is not allowed to call with multi-thread? saying we are testing something `not supported`?

https://github.com/ros2/rcutils/blob/3485f1bdc805ead1e9c26509479dc8c1f94b6951/include/rcutils/logging.h#L480-L505

clalancette: Yes, it is tricky, and the reason that I introduced the problem to begin with.

It is definitely the case that `rcutils_log` is *not* thread-safe with respect to `rcutils_set_logger_level` (before or after this PR).  If you call the two of those from two different threads (without any additional locking), the results will be undefined.

However, prior to my optimizations, `rcutils_log` was thread-safe with itself.  That is, because `rcutils_log` was only reading data out of the underlying data structures, it didn't matter how many threads were calling it.  With the current code (before this PR), that is no longer the case because the individual threads can write to the hashmap, leading to corruption/crashing.  This PR fixes it by essentially making `rcutils_log` completely read-only again.

One other bit of information here pulls this all together.  If we are coming via the `RCLCPP_` logging macros, there is actually no problem; `rclcpp` has a global lock around all logging.  However, some of our lower-level components (like `rcl`, `rmw`, etc) use the `RCUTILS_` macros, and they do not have this global lock.

I think we should consider adding some kind of reader-writer lock here, which would solve most of the problem.  However, we have classically not done any locking in the rcutils layer, so we would have to change that policy (and figure out how to do it in a cross-platform way).

Does all of that make sense @fujitatomoya ?

clalancette (changes requested):
In short, I'm fine with disabling this optimization for now.

However, I do think we should add additional language into the doc block for `rcutils_log` saying that it is expected to be thread-safe *with itself*.  That might have saved me from introducing this problem to begin with (though the new test may also do that in the future).

With that change made, :+1: from me.

src/logging.c
```diff
@@ -943,18 +943,23 @@ int rcutils_logging_get_logger_effective_level(const char * name)
...
-  }
+  // TODO(wjwwood): restore or replace this optimization when thread-safety is addressed
+  //   see: https://github.com/ros2/rcutils/pull/393
+  // // If the calculated severity is anything but UNSET, we place it into the hashmap for speedier
+  // // lookup next time.  If the severity is UNSET, we don't bother because we are going to have to
+  // // walk the hierarchy next time anyway.
+  // if (severity != RCUTILS_LOG_SEVERITY_UNSET) {
+  //   fprintf(
+  //     stderr, "in optimization of rcutils_logging_get_logger_effective_level('%s') -> %d\n",
+  //     name, severity);
```
jacobperron, line 954:
```suggestion
```

test/test_logging.cpp
wjwwood: The problem is that this is just not how people use the logging macros, which of course use this function.

Also, there's no good place to inject this locking other than in rcutils, and we've been trying to avoid using threading in rcutils/rmw/rcl in order to use client library language specific mechanisms for threading when possible.

I think a reader-writer lock is perfect for this case and I think we could probably implement a relatively efficient (compared to actually locking with a mutex) lock-free form of this, since the writer lock is only needed when setting log levels and that should be done infrequently. It's something I intend to investigate.

Also, I think this doc block was added after the functions were written and just didn't consider how it was being used. Unfortunately, I think there will be many such inaccuracies in the doc blocks that were back filled a few years ago in the REP-2004 QD effort.

fujitatomoya: understood, thanks for the explanation guys! i am good to go.

fujitatomoya (approved):

jacobperron (approved):

wjwwood:
I'm not going to re-run ci.ros2.org because the changes were docs only and the linters are run in the Rpr job.

PR #408, opened 2023-01-24
Add SHA256 utility implementation

emersonknapp:
To be used for REP-2011 type version hashing.
Simple sha256 implementation to generate a 256-byte message digest for any data. Implementation originally copied from Brad Conte https://github.com/B-Con/crypto-algorithms, with modifications to fit code style and fix compiler warnings.

Prereq for ros2/rcl#1027
Part of https://github.com/ros2/ros2/issues/1159

### DIFF
--- a/include/rcutils/sha256.h
+++ b/include/rcutils/sha256.h
@@ -18,20 +18,17 @@
 /// Implementation originally copied from Brad Conte
 /// https://github.com/B-Con/crypto-algorithms/blob/master/sha256.c
 
-/// Simple SHA256 implementation
-/*********************************************************************
-* Filename:   sha256.c
-* Author:     Brad Conte (brad AT bradconte.com)
-* Copyright:
-* Disclaimer: This code is presented "as is" without any guarantees.
-* Details:    Implementation of the SHA-256 hashing algorithm.
-              SHA-256 is one of the three algorithms in the SHA2
-              specification. The others, SHA-384 and SHA-512, are not
-              offered in this implementation.
-              Algorithm specification can be found here:
-               * http://csrc.nist.gov/publications/fips/fips180-2/fips180-2withchangenotice.pdf
-              This implementation uses little endian byte order.
-*********************************************************************/
+/** \file sha256.h
+ *  \brief SHA256 implementation
+ *
+ *  This contains an implementation of the SHA256 algorithm
+ *  It was originally copied from Brad Conte
+ *  https://github.com/B-Con/crypto-algorithms/blob/master/sha256.c
+ *  and modified to meet ros2 code formatting and compiler warning requirements.
+ *  Algorithm specification can be found here:
+ *  http://csrc.nist.gov/publications/fips/fips180-2/fips180-2withchangenotice.pdf
+ *  This implementation uses little endian byte order.
+ */
 
 #ifndef RCUTILS__SHA256_H_
 #define RCUTILS__SHA256_H_
@@ -55,8 +52,9 @@ typedef struct
   uint32_t state[8];
 } rcutils_sha256_ctx_t;
 
+/// Initialize the sha256 algorithm context with starting state.
 /**
- *
+ * Call this on any new context before starting to input data.
  *
  * \param[inout] ctx
  * \return void
@@ -64,22 +62,24 @@ typedef struct
 RCUTILS_PUBLIC
 void rcutils_sha256_init(rcutils_sha256_ctx_t * ctx);
 
+/// Add data to the sha256 algorithm
 /**
+ * This may be called repeatedly on an initialized context.
  *
- *
- * \param ctx
- * \param data
- * \param len
+ * \param[inout] ctx Initialized sha256 context struct
+ * \param[in] data Data to add to the total message being hashed
+ * \param[in] data_len Size of the input data.
  * \return void
  */
 RCUTILS_PUBLIC
 void rcutils_sha256_update(rcutils_sha256_ctx_t * ctx, const uint8_t * data, size_t data_len);
 
+/// Finalize and output sha256 hash for all data added.
 /**
+ * Call only once on a context that has been initialized, and optionally updated with data.
  *
- *
- * \param ctx
- * \param hash
+ * \param[inout] ctx Initialized sha256 context struct
+ * \param[out] hash Calculated sha256 message digest
  * \return void
  */
 RCUTILS_PUBLIC
--- a/include/rcutils/sha256.h
+++ b/include/rcutils/sha256.h
@@ -47,7 +47,7 @@ extern "C"
 typedef struct RCUTILS_PUBLIC_TYPE rcutils_sha256_ctx_s
 {
   uint8_t data[64];
-  uint32_t datalen;
+  size_t datalen;
   uint64_t bitlen;
   uint32_t state[8];
 } rcutils_sha256_ctx_t;
@@ -83,7 +83,7 @@ void rcutils_sha256_update(rcutils_sha256_ctx_t * ctx, const uint8_t * data, siz
  * \return void
  */
 RCUTILS_PUBLIC
-void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t hash[RCUTILS_SHA256_BLOCK_SIZE]);
+void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t output_hash[RCUTILS_SHA256_BLOCK_SIZE]);
 
 #ifdef __cplusplus
 }
--- a/include/rcutils/sha256.h
+++ b/include/rcutils/sha256.h
@@ -32,11 +32,12 @@ extern "C"
 
 #define RCUTILS_SHA256_BLOCK_SIZE 32
 
-typedef struct {
-	uint8_t data[64];
-	uint32_t datalen;
-	uint64_t bitlen;
-	uint32_t state[8];
+typedef struct
+{
+  uint8_t data[64];
+  uint32_t datalen;
+  uint64_t bitlen;
+  uint32_t state[8];
 } rcutils_sha256_ctx_t;
 
 /// Simple SHA256 implementation
@@ -92,4 +93,4 @@ rcutils_ret_t sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t hash[]);
 }
 #endif
 
-#endif  // RCUTILS__QSORT_H_
+#endif  // RCUTILS__SHA256_H_
--- a/include/rcutils/sha256.h
+++ b/include/rcutils/sha256.h
@@ -49,10 +49,10 @@ extern "C"
 
 typedef struct
 {
-  unsigned char data[64];
-  unsigned int datalen;
-  unsigned long long bitlen;
-  unsigned int state[8];
+  uint8_t data[64];
+  uint32_t datalen;
+  uint64_t bitlen;
+  uint32_t state[8];
 } rcutils_sha256_ctx_t;
 
 /**
--- a/include/rcutils/sha256.h
+++ b/include/rcutils/sha256.h
@@ -83,7 +83,9 @@ void rcutils_sha256_update(rcutils_sha256_ctx_t * ctx, const uint8_t * data, siz
  * \return void
  */
 RCUTILS_PUBLIC
-void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t output_hash[RCUTILS_SHA256_BLOCK_SIZE]);
+void rcutils_sha256_final(
+  rcutils_sha256_ctx_t * ctx,
+  uint8_t output_hash[RCUTILS_SHA256_BLOCK_SIZE]);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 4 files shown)


src/sha256.c
```diff
@@ -0,0 +1,169 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
```
clalancette, line 18: This is not needed in the C file, only in the header file.

src/sha256.c
```diff
@@ -0,0 +1,169 @@
...
+
+#define ROTLEFT(a,b) (((a) << (b)) | ((a) >> (32-(b))))
+#define ROTRIGHT(a,b) (((a) >> (b)) | ((a) << (32-(b))))
+
+#define CH(x,y,z) (((x) & (y)) ^ (~(x) & (z)))
+#define MAJ(x,y,z) (((x) & (y)) ^ ((x) & (z)) ^ ((y) & (z)))
+#define EP0(x) (ROTRIGHT(x,2) ^ ROTRIGHT(x,13) ^ ROTRIGHT(x,22))
+#define EP1(x) (ROTRIGHT(x,6) ^ ROTRIGHT(x,11) ^ ROTRIGHT(x,25))
+#define SIG0(x) (ROTRIGHT(x,7) ^ ROTRIGHT(x,18) ^ ((x) >> 3))
+#define SIG1(x) (ROTRIGHT(x,17) ^ ROTRIGHT(x,19) ^ ((x) >> 10))
```
clalancette, line 32: For what it is worth, I think for most of these we could use an inline function rather than a `#define`.  That would give us better type checking, at least.

src/sha256.c
```diff
@@ -0,0 +1,169 @@
...
+	0x983e5152,0xa831c66d,0xb00327c8,0xbf597fc7,0xc6e00bf3,0xd5a79147,0x06ca6351,0x14292967,
+	0x27b70a85,0x2e1b2138,0x4d2c6dfc,0x53380d13,0x650a7354,0x766a0abb,0x81c2c92e,0x92722c85,
+	0xa2bfe8a1,0xa81a664b,0xc24b8b70,0xc76c51a3,0xd192e819,0xd6990624,0xf40e3585,0x106aa070,
+	0x19a4c116,0x1e376c08,0x2748774c,0x34b0bcb5,0x391c0cb3,0x4ed8aa4a,0x5b9cca4f,0x682e6ff3,
+	0x748f82ee,0x78a5636f,0x84c87814,0x8cc70208,0x90befffa,0xa4506ceb,0xbef9a3f7,0xc67178f2
+};
+
+void sha256_transform(rcutils_sha256_ctx_t * ctx, const uint8_t data[])
+{
+	uint32_t a, b, c, d, e, f, g, h, i, j, t1, t2, m[64];
```
clalancette, line 47: According to our style, this is over-indented; it should just be 2 spaces.  Same with all of the rest of the code below.

src/sha256.c
```diff
@@ -0,0 +1,169 @@
...
+
+	for (i = 0, j = 0; i < 16; ++i, j += 4)
```
clalancette, line 49: According to our style, this loop should use brackets.  Same for other loops below.

clalancette (changes requested):
Besides the things I've pointed out inline, we should also add tests for this implementation.

All of that said, the idea here is good.

emersonknapp:
Thanks for the comments. I hadn't finished doing code style yet, just wanted to have this open as draft for the rep2011 discussion today.

Agreed, tests necessary.

src/sha256.c
emersonknapp: Agreed - done as `static inline uint32_t` for all

fujitatomoya (commented):
@emersonknapp @clalancette quick question, why we are not using openssl to avoid re-implementing the function in rcutils? i was thinking that would be more cost effective for maintenance? Do we want to avoid the dependency on that intentionally? just a question.

emersonknapp:
Yes, the motivation is that we want to use this function in RCL, which does not have a direct dependency on OpenSSL. ros2-security does have the OpenSSL dependency, but not all installations will enable that feature, but type version hashing will be built into rcl. 

@wjwwood may also have more input on the reason

src/sha256.c
```diff
@@ -0,0 +1,180 @@
...
+  0xd807aa98, 0x12835b01, 0x243185be, 0x550c7dc3, 0x72be5d74, 0x80deb1fe, 0x9bdc06a7, 0xc19bf174,
+  0xe49b69c1, 0xefbe4786, 0x0fc19dc6, 0x240ca1cc, 0x2de92c6f, 0x4a7484aa, 0x5cb0a9dc, 0x76f988da,
+  0x983e5152, 0xa831c66d, 0xb00327c8, 0xbf597fc7, 0xc6e00bf3, 0xd5a79147, 0x06ca6351, 0x14292967,
+  0x27b70a85, 0x2e1b2138, 0x4d2c6dfc, 0x53380d13, 0x650a7354, 0x766a0abb, 0x81c2c92e, 0x92722c85,
+  0xa2bfe8a1, 0xa81a664b, 0xc24b8b70, 0xc76c51a3, 0xd192e819, 0xd6990624, 0xf40e3585, 0x106aa070,
+  0x19a4c116, 0x1e376c08, 0x2748774c, 0x34b0bcb5, 0x391c0cb3, 0x4ed8aa4a, 0x5b9cca4f, 0x682e6ff3,
+  0x748f82ee, 0x78a5636f, 0x84c87814, 0x8cc70208, 0x90befffa, 0xa4506ceb, 0xbef9a3f7, 0xc67178f2
+};
+
+void sha256_transform(rcutils_sha256_ctx_t * ctx, const uint8_t data[])
```
clalancette, line 65:
```suggestion
static void sha256_transform(rcutils_sha256_ctx_t * ctx, const uint8_t data[])
```

(it is not part of the public API)

src/sha256.c
```diff
@@ -0,0 +1,180 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string.h>
+
+#include "rcutils/sha256.h"
+
+static inline uint32_t rotright(uint32_t a, uint8_t b)
+{
+  return (a >> b) | (a << (32 - b));
```
clalancette, line 21: The `32 - b` is slightly suspect.  What if `b` is 255?  We would then end up shifting left with a negative number, which I think is undefined behavior.  I guess given how we use this function it won't happen, but maybe we should add an assert or something?  What do you think?

src/sha256.c
clalancette: Additionally, do we even need the second argument?  We already have `ctx`, and all uses of this function pass `ctx->data` as the second argument.  But we could just dereference that within this function itself.

src/sha256.c
```diff
@@ -0,0 +1,180 @@
...
+
+  for (i = 0; i < len; ++i) {
+    ctx->data[ctx->datalen] = data[i];
+    ctx->datalen++;
+    if (ctx->datalen == 64) {
+      sha256_transform(ctx, ctx->data);
+      ctx->bitlen += 512;
+      ctx->datalen = 0;
+    }
+  }
```
clalancette, line 134: I'm not sure it is worth it, but we could probably optimize this.  In particular, we could `memcpy` 64 bytes at a time into `ctx->data` (dealing with the special case of the first block and the last block), calling `sha256_transform` on each 64-byte block we complete.

I'll leave it up to you whether you think that improvement is worth it and you want to try it.  This is not a blocking comment from me.

src/sha256.c
```diff
@@ -0,0 +1,180 @@
...
+void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t hash[RCUTILS_SHA256_BLOCK_SIZE])
+{
+  uint32_t i = ctx->datalen;
+
+  // Pad whatever data is left in the buffer.
+  if (ctx->datalen < 56) {
+    ctx->data[i++] = 0x80;
+    while (i < 56) {
+      ctx->data[i++] = 0x00;
+    }
```
clalancette, line 146: Same comment here about optimizing; this would probably be faster as a `memset`.

src/sha256.c
```diff
@@ -0,0 +1,180 @@
...
+  } else {
+    ctx->data[i++] = 0x80;
+    while (i < 64) {
+      ctx->data[i++] = 0x00;
+    }
```
clalancette, line 151: Ditto.

include/rcutils/sha256.h
```diff
@@ -0,0 +1,92 @@
...
+/// Finalize and output sha256 hash for all data added.
+/**
+ * Call only once on a context that has been initialized, and optionally updated with data.
+ *
+ * \param[inout] ctx Initialized sha256 context struct
+ * \param[out] hash Calculated sha256 message digest
+ * \return void
+ */
+RCUTILS_PUBLIC
+void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t hash[RCUTILS_SHA256_BLOCK_SIZE]);
```
clalancette, line 86: I know the documentation says that `hash` is an out parameter, but it might be nice to encode that in the name:

```suggestion
void rcutils_sha256_final(rcutils_sha256_ctx_t * ctx, uint8_t output_hash[RCUTILS_SHA256_BLOCK_SIZE]);
```

test/test_sha256.cpp
```diff
@@ -0,0 +1,75 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "rcutils/sha256.h"
+
+TEST(TestSHA256, test_text1) {
+  uint8_t text1[] = {"abc"};
+  size_t text1_len = sizeof(text1) - 1;
+  uint8_t hash1[RCUTILS_SHA256_BLOCK_SIZE] = {
```
clalancette, line 22:
```suggestion
  uint8_t expected_hash1[RCUTILS_SHA256_BLOCK_SIZE] = {
```

test/test_sha256.cpp
```diff
@@ -0,0 +1,75 @@
...
+  rcutils_sha256_update(&ctx, text1, text1_len);
+  rcutils_sha256_final(&ctx, buf);
+
+  ASSERT_EQ(0, memcmp(hash1, buf, RCUTILS_SHA256_BLOCK_SIZE));
+}
+
+TEST(TestSHA256, test_text2) {
+  uint8_t text2[] = {"abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq"};
+  size_t text2_len = sizeof(text2) - 1;
+  uint8_t hash2[RCUTILS_SHA256_BLOCK_SIZE] = {
```
clalancette, line 41:
```suggestion
  uint8_t expected_hash2[RCUTILS_SHA256_BLOCK_SIZE] = {
```

test/test_sha256.cpp
```diff
@@ -0,0 +1,75 @@
...
+  rcutils_sha256_final(&ctx, buf);
+
+  ASSERT_EQ(0, memcmp(hash2, buf, RCUTILS_SHA256_BLOCK_SIZE));
+}
+
+TEST(TestSHA256, test_multi_update) {
+  uint8_t text[] = {"aaaaaaaaaa"};
+  size_t text_len = sizeof(text) - 1;
+
+  uint8_t hash[RCUTILS_SHA256_BLOCK_SIZE] = {
```
clalancette, line 60:
```suggestion
  uint8_t expected_hash[RCUTILS_SHA256_BLOCK_SIZE] = {
```

clalancette (changes requested):
I've left a few comments, but they are all pretty minor.  Overall this looks pretty good to me.

src/sha256.c
emersonknapp: Yes, this is really a `uint5_t` expected input range. I think an assert is fine as long as we're optimizing them out in release builds.

src/sha256.c
emersonknapp: I gave block-copy implementation a go. The tests still pass so it seems all right. I don't have a performance comparison but maybe it looks better this way.

src/sha256.c
emersonknapp: Done

clalancette (approved):
Thanks for iterating.  This looks good to me with green CI.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/e2b18a4fab8fcd32f47ceb057768b8e8/raw/fddd335b1bc03a0c5381e82c3b3989fae80e2d5f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11402
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17994/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12524/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18631/)

emersonknapp:
unrelated rclcpp test failed, rerunning Linux:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/17998/)

clalancette:
You can ignore that failure; it is fixed in https://github.com/ros2/rclcpp/pull/2092 .

emersonknapp:
OK then :+1:, I don't have merge access on this repo so that's up to you now! Thanks for the reviews

PR #411, opened 2023-02-27
Fix memory leak in string_map.c in rcutils

nnmm:
If realloc for new_values failed, string_map->impl->keys would have a size that is inconsistent with string_map->impl->capacity, leading to a memory leak in the best case, and invalid memory access in the worst case.

### DIFF
--- a/src/string_map.c
+++ b/src/string_map.c
@@ -172,7 +172,6 @@ rcutils_string_map_reserve(rcutils_string_map_t * string_map, size_t capacity)
     return RCUTILS_RET_OK;
   } else if (capacity == 0) {
     // if the requested capacity is zero, then make sure the existing keys and values are free'd
-    __clear_string_map(string_map);
     allocator.deallocate(string_map->impl->key_value_pairs, allocator.state);
     string_map->impl->key_value_pairs = NULL;
     // falls through to normal function end
@@ -181,17 +180,17 @@ rcutils_string_map_reserve(rcutils_string_map_t * string_map, size_t capacity)
     // note that realloc when the pointer is NULL is the same as malloc
     // note also that realloc will shrink the space if needed
 
-    // ensure that reallocate won't overflow capacity
+    // ensure that reallocate won't overflow SIZE_MAX
     if (capacity > (SIZE_MAX / sizeof(key_value_pair_t))) {
       RCUTILS_SET_ERROR_MSG("requested capacity for string_map too large");
       return RCUTILS_RET_BAD_ALLOC;
     }
 
-    // resize the keys, assigning the result only if it succeeds
+    // resize the keys and values, assigning the result only if it succeeds
     key_value_pair_t * new_key_value_pairs = allocator.reallocate(
       string_map->impl->key_value_pairs, capacity * sizeof(key_value_pair_t), allocator.state);
     if (NULL == new_key_value_pairs) {
-      RCUTILS_SET_ERROR_MSG("failed to allocate memory for string_map keys");
+      RCUTILS_SET_ERROR_MSG("failed to allocate memory for string_map key-value pairs");
       return RCUTILS_RET_BAD_ALLOC;
     }
     string_map->impl->key_value_pairs = new_key_value_pairs;
@@ -308,7 +307,7 @@ rcutils_string_map_set_no_resize(
   char * original_value = string_map->impl->key_value_pairs[key_index].value;
   char * new_value = rcutils_strdup(value, allocator);
   if (NULL == new_value) {
-    RCUTILS_SET_ERROR_MSG("failed to allocate memory for key");
+    RCUTILS_SET_ERROR_MSG("failed to allocate memory for value");
     if (should_free_key_on_error) {
       allocator.deallocate(string_map->impl->key_value_pairs[key_index].key, allocator.state);
       string_map->impl->key_value_pairs[key_index].key = NULL;
--- a/src/string_map.c
+++ b/src/string_map.c
@@ -29,10 +29,15 @@ extern "C"
 #include "rcutils/format_string.h"
 #include "rcutils/types/rcutils_ret.h"
 
+typedef struct key_value_pair
+{
+  char * key;
+  char * value;
+} key_value_pair_t;
+
 typedef struct rcutils_string_map_impl_s
 {
-  char ** keys;
-  char ** values;
+  key_value_pair_t * key_value_pairs;
   size_t capacity;
   size_t size;
   rcutils_allocator_t allocator;
@@ -64,8 +69,7 @@ rcutils_string_map_init(
     RCUTILS_SET_ERROR_MSG("failed to allocate memory for string map impl struct");
     return RCUTILS_RET_BAD_ALLOC;
   }
-  string_map->impl->keys = NULL;
-  string_map->impl->values = NULL;
+  string_map->impl->key_value_pairs = NULL;
   string_map->impl->capacity = 0;
   string_map->impl->size = 0;
   string_map->impl->allocator = allocator;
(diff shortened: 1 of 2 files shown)


clalancette (commented):
I see the problem you are pointing out, but I don't necessarily agree that we can't use `reallocate` here.

Instead of this change, if we just move the assignment of `string_map->impl->keys` until *after* we know that the `new_values` is not NULL, that would accomplish the same thing, correct?  That is (pseudo-code):

```
new_keys = reallocate();
if (new_keys == NULL) {
    return RCUTILS_RET_BAD_ALLOC;
}
new_values = reallocate();
if (new_values == NULL) {
    free(new_keys);
    return RCUTILS_RET_BAD_ALLOC;
}
string->map->impl->keys = new_keys;
string->map->impl->values = new_values;
```

(`reallocate` has performance benefits, so I don't want to move away from it if we don't have to)

nnmm:
@clalancette I thought of that too, but decided against it because reallocate is allowed to return the same pointer. If `new_keys == string->map->impl->keys`, you've freed `string->map->impl->keys`.

clalancette:
> @clalancette I thought of that too, but decided against it because reallocate is allowed to return the same pointer. If `new_keys == string->map->impl->keys`, you've freed `string->map->impl->keys`. Edit: And checking for that and not freeing is also not an option, because then `string->map->impl->keys` would have a different size from the capacity and the values.

Good point.

That said, looking at it again I'm not entirely sure where the problem is.  It is true that we may succeed in reallocating the `keys`, but not the `values`.  But we don't update the `capacity` or the `size` in that case.  That means that all other operations should still work fine on the old data, and we should never run into a memory leak or invalid memory accesses.

Can you explain a bit more about the problem that caused you to look at this?

nnmm:
> That said, looking at it again I'm not entirely sure where the problem is. It is true that we may succeed in reallocating the `keys`, but not the `values`. But we don't update the `capacity` or the `size` in that case. That means that all other operations should still work fine on the old data, and we should never run into a memory leak or invalid memory accesses.

If the new capacity is smaller than the old one, then the map will still report the old capacity even though `string->map->impl->keys` is now smaller than that. Then e.g. `rcutils_string_map_clear()` will access invalid memory.
```cpp
  for (; i < string_map->impl->capacity; ++i) {
    if (string_map->impl->keys[i] != NULL) {
      …
    }
  }
```

> Can you explain a bit more about the problem that caused you to look at this?

Apex.AI used the time bomb allocator on the string map to increase coverage, and I noticed that this test leaked memory.

clalancette:
> If the new capacity is smaller than the old one, and the value reallocation fails, then the map will still report the old capacity even though `string->map->impl->keys` is now smaller than that. Then e.g. `rcutils_string_map_clear()` will access invalid memory.

Ah!  Thanks for the explanation, now I see the issue.

That said, I'm still not that willing to give up the realloc if we don't have to.  I know that this is only for the pointer tracking, but I want to give the malloc implementation as much chance as we can for performance.

Thinking about this more, fundamentally I think the problem is that we are tracking the keys and values separately.  That is, we really should be allocating and reallocating the keys and values as pairs.  That way, they always succeed or fail as a pair.  Luckily, we are using the PIMPL pattern here, so we are free to make a change like this.  Towards that end, please see https://github.com/ApexAI/rcutils/pull/1, where I've done exactly this.  I'm interested to hear your thoughts about it.  If you agree with the direction, feel free to merge it into this PR and we can continue from there.

fujitatomoya (commented):
implementation looks good to me, either @clalancette proposal or this, i am okay to go.

nnmm:
> Thinking about this more, fundamentally I think the problem is that we are tracking the keys and values separately. That is, we really should be allocating and reallocating the keys and values as pairs. That way, they always succeed or fail as a pair. Luckily, we are using the PIMPL pattern here, so we are free to make a change like this. Towards that end, please see [ApexAI#1](https://github.com/ApexAI/rcutils/pull/1), where I've done exactly this. I'm interested to hear your thoughts about it. If you agree with the direction, feel free to merge it into this PR and we can continue from there.

I think that's a brilliant idea, it makes things much simpler. I merged it, but now I see that your commit wasn't signed off, how do we best solve that?

clalancette:
> I think that's a brilliant idea, it makes things much simpler. I merged it, but now I see that your commit wasn't signed off, how do we best solve that?

It's upset because of the merge commit, which doesn't have a sign-off.  The easiest thing to do here is to rebase and squash all of the commits together with proper Signed-off-by lines from both you and me.  I'd do it myself, but I don't have permissions to push to the ApexAI account, so you'll have to do it.

src/string_map.c
```diff
@@ -147,46 +172,34 @@ rcutils_string_map_reserve(rcutils_string_map_t * string_map, size_t capacity)
     return RCUTILS_RET_OK;
   } else if (capacity == 0) {
     // if the requested capacity is zero, then make sure the existing keys and values are free'd
-    allocator.deallocate(string_map->impl->keys, allocator.state);
-    string_map->impl->keys = NULL;
-    allocator.deallocate(string_map->impl->values, allocator.state);
-    string_map->impl->values = NULL;
+    allocator.deallocate(string_map->impl->key_value_pairs, allocator.state);
```
clalancette, line 175: Note that I'm pretty sure we need the call to `__clear_string_map` here.  Otherwise, while we will free the list of key/value pairs, we won't actually free the keys and values themselves.

clalancette (commented):
I've left one comment that I think we need to replace the `__clear_string_map` call.  Other than that, this looks good to me so I'll run CI once we come to consensus on that.

src/string_map.c
nnmm: It cannot be that `size != 0` and `capacity == 0`, because of the recursive call above. See https://github.com/ApexAI/rcutils/pull/1#discussion_r1120097853. So there are no keys and values to free.

clalancette: Oh, good point.  OK, this looks good to me then!

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18120/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12651/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18771/)

src/string_map.c
```diff
@@ -147,46 +172,34 @@ rcutils_string_map_reserve(rcutils_string_map_t * string_map, size_t capacity)
...
     // falls through to normal function end
   } else {
     // if the capacity non-zero and different, use realloc to increase/shrink the size
     // note that realloc when the pointer is NULL is the same as malloc
     // note also that realloc will shrink the space if needed
 
-    // ensure that reallocate won't overflow capacity
-    if (capacity > (SIZE_MAX / sizeof(char *))) {
+    // ensure that reallocate won't overflow SIZE_MAX
+    if (capacity > (SIZE_MAX / sizeof(key_value_pair_t))) {
```
fujitatomoya, line 184: Why capacity is down into half? originally capacity is `(SIZE_MAX / sizeof(char *))` for both entries, but current code shrinks it down to half? i might be mistaken...

clalancette: You are correct; on a 64-bit platform, the old check was:
```
capacity > (SIZE_MAX / sizeof(char *)) === capacity > (18446744073709551615UL / 8) === capacity > 2305843009213693951
```

The new check would be:
```
capacity > (SIZE_MAX / sizeof(key_value_pair_t)) === capacity > (18446744073709551615UL / 16) === capacity > 1152921504606846975
```

On a 32-bit platform (which we don't currently support), the old would be:
```
capacity > (SIZE_MAX / sizeof(char *)) === capacity > (4294967295UL / 8) === capacity > 536870911
```

The new would be:
```
capacity > (SIZE_MAX / sizeof(char *)) === capacity > (4294967295UL / 16) === capacity > 268435455
```

In all cases, this is more-or-less just a check to see if someone allocated something crazy; we don't really expect maps with that many entries.  So I don't see a huge deal here, but if you have other thoughts on what we might change this to I'm happy to hear it.

iuhilnehc-ynos (approved):
LGTM with two minor suggestions.

src/string_map.c
```diff
@@ -130,6 +134,27 @@ rcutils_string_map_get_size(const rcutils_string_map_t * string_map, size_t * si
...
+  rcutils_allocator_t allocator = string_map_impl->allocator;
+  allocator.deallocate(string_map_impl->key_value_pairs[index].key, allocator.state);
+  string_map_impl->key_value_pairs[index].key = NULL;
+  allocator.deallocate(string_map_impl->key_value_pairs[index].value, allocator.state);
+  string_map_impl->key_value_pairs[index].value = NULL;
+  string_map_impl->size--;
+}
+
+static void
+__clear_string_map(rcutils_string_map_t * string_map)
```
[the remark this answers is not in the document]
clalancette: > Yes, I should have undone this when I removed the other call to this function. @clalancette Do you think inlining that function again is worth another CI run?

Yeah, we may as well do it.  Once this is done and the other thing is addressed, I'll rerun CI.

clalancette:
Here's another CI.  Assuming this comes back clean, we just need to wait for @fujitatomoya to resolve the last outstanding conversation:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18130/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12668/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18780/)

src/string_map.c
fujitatomoya: yeah, allocating this much is crazy enough, current code should do good for checking 👍 (this was more like question.)

fujitatomoya (approved):

nnmm:
It's green with a tinge of yellow, is that okay?

clalancette:
All right, the yellow on Windows is in some other builds, so not caused by this.  I'll go ahead and merge, thanks for the contribution.

PR #414, opened 2023-03-07
Out of bounds range check

jrutgeer:
Not tested!

I think this should be ">" instead of ">= ":
https://github.com/ros2/rcutils/blob/1db1f16a49378f4693b74296449ee920ff53cffe/src/logging.c#L978-L980
As you want `level` to be in the range [0, 5] and not [0, 6]. https://github.com/ros2/rcutils/blob/1db1f16a49378f4693b74296449ee920ff53cffe/src/logging.c#L68-L75

Similar to this:
https://github.com/ros2/rcutils/blob/1db1f16a49378f4693b74296449ee920ff53cffe/src/logging.c#L720-L721

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -976,7 +976,7 @@ rcutils_ret_t rcutils_logging_set_logger_level(const char * name, int level)
 
   // Convert the severity value into a string for storage.
   if (level < 0 ||
-    level >=
+    level >
     (int)(sizeof(g_rcutils_log_severity_names) / sizeof(g_rcutils_log_severity_names[0])))
   {
     RCUTILS_SET_ERROR_MSG("Invalid severity level specified for logger");


jrutgeer:
How does this DCO thing work through for an edit through github online? Or is the only way to do the whole process (fork, clone, commit, push, etc)?

jrutgeer:
Silly me, I got confused sorry

wjwwood:
> How does this DCO thing work through for an edit through github online? Or is the only way to do the whole process (fork, clone, commit, push, etc)?

I think you can do it via online editing, but you have to make sure to manually sign-off on the commit when committing (i.e. you need to manually add something like `Signed-off-by: Humpty Dumpty <humpty.dumpty@example.com>` in your commit message in the GitHub webui when making the change).

Once you make the commit, I think you might need to use the git CLI to fix the problem using `git commit --amend ...`, but the DCO bot usually will give you a command you can run to fix it, if you click through the failure link.

Issue #269, opened 2020-07-27
Simplify error handling in failure mode or during finalization 


hidmic:
### Rationale

Error handling support in `rcutils` (and by extension, in every ROS 2 C library implementation or client code) falls short when it comes to error propagation when in failure mode or during finalization. As discussed in [ros2/rmw_fastrtps#414](https://github.com/ros2/rmw_fastrtps/pull/414#discussion_r460170546) and [ros2/rmw_cyclonedds#210](https://github.com/ros2/rmw_cyclonedds/pull/210#pullrequestreview-455037275), ensuring the initial error gets propagated, while no subsequent error goes unnoticed, clutters client code significantly.

### Proposal

Introduce a mechanism for `rcutils` to distinguish between an error overwrite and a nested error that should be handled differently (e.g. logging to `stderr` directly). For that matter, have a thread local integer to track nested error handling _scopes_ and a boolean in local storage to ensure a single error handling _scope_ per function.

With an API like:

```c
typedef struct rcutils_error_handling_scope_t {
    bool active;
} rcutils_error_handling_scope_t;

void rcutils_error_handling_scope_init(rcutils_error_handling_scope_t * scope);
void rcutils_error_handling_scope_enter(rcutils_error_handling_scope_t * scope);
void rcutils_error_handling_scope_leave(rcutils_error_handling_scope_t * scope);
```

a function may request `rcutils_set_error_state` to behave differently in between `_enter` and `_leave` calls, for its own code and the call chain that follows after it.

hidmic:
FYI @clalancette @ivanpauno

wjwwood:
My problem with a call chain is that you cannot know how deep it needs to be. If we have one then, in my opinion, it should be preallocated per thread and have a fixed depth, similar to how the error message itself has a fixed length and is preallocated per thread.

hidmic:
I'm not sure I follow. The purpose of  `rcutils_error_handling_scope_t` is not to store anything, but to aid changing error handling macros (e.g. `RCUTILS_SET_ERROR_MSG`) behavior whenever errors occur while handling another error. I guess we could generate full backtraces within each error handling "scope", but we also can start small and just print to `stderr` (like we do explicitly in quite a few places).

gbiggs:
> My problem with a call chain is that you cannot know how deep it needs to be

Wouldn't it be possible to make a reasonable guess for how much we need, at least inside the ROS API? We can take a look at our call graph and see how deep it goes.

[...]

PR #421, opened 2023-04-09
Add convenience error handling macros

methylDragon:
This adds a bunch of macros to append error strings.

So before you would do:
```
    auto old_error = rcutils_get_error_msg();
    rcutils_reset_error();
    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("New error:\n%s", old_error.str);
```

Now you can do:
```
    RCUTILS_SET_ERROR_MSG_AND_APPEND_PREV_ERROR("New error");
```

(And similarly for setting error with format strings, and also safe fprint)

If no old error was set, then the behavior is just like RCUTILS_SET_ERROR_MSG


include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
...
+#define _RCUTILS_EXPAND(x) x
+
+/// Set the error message, as well as append the current file, line number, and previous error.
+/**
+ * If there is no previous error, has same behavior as RCUTILS_SET_ERROR_MSG.
+ * \param[in] msg The error message to be set.
+ */
+#define RCUTILS_SET_ERROR_MSG_AND_APPEND_PREV_ERROR(msg) \
+  do { \
+    if (rcutils_error_is_set()) { \
```
wjwwood, line 327: I would not add this as part of the macro. I think it encourages code that doesn't know if the called function is going to set the error. Instead I think it's better to assert one way or the other. 

The reason being that just recently I was debugging why a called function was failing and it returned error not set when I appended it. That told me which of the error checking statements was the cause and let me fix that. I think pretty much every error check that returns an error code should set the error message.

include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
...
+      rcutils_error_string_t error_string = rcutils_get_error_string(); \
+      rcutils_reset_error(); \
+      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING( \
+        _RCUTILS_EXPAND(msg ":\n%s"), error_string.str); \
```
wjwwood, line 331: I would not use a new line. I would recommend just using a : and space.

include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
 void
 rcutils_reset_error(void);
 
+// Helper macro for nested macro expansion
+#define _RCUTILS_EXPAND(x) x
```
wjwwood, line 318: I'm not sure, but maybe this is what you want?

https://github.com/ros2/rcutils/blob/5bd1230eb10595cec30168667adf03f48879b96c/include/rcutils/macros.h#L68

include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
...
+/**
+ * This will reset the previous error, if it exists.
+ * If there is no previous error, has same behavior as RCUTILS_SAFE_FWRITE_TO_STDERR.
+ */
+#define RCUTILS_SAFE_FWRITE_TO_STDERR_AND_APPEND_PREV_ERROR(msg) \
+  do { \
+    if (rcutils_error_is_set()) { \
+      rcutils_error_string_t error_string = rcutils_get_error_string(); \
+      rcutils_reset_error(); \
+      RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING( \
```
wjwwood, line 371: Note that this macro has a hard length limit: https://github.com/ros2/rcutils/blob/5bd1230eb10595cec30168667adf03f48879b96c/include/rcutils/error_handling.h#L68

So as to avoid allocation. 

Using multiple calls to the non format version can be preferable. It allows for interleaving in multithreaded output but that's often a good trade off for writing and arbitrary amount of info to the screen without doing formatting. 

The format version is still appropriate when doing something other than concatenating strings. 

For this particular macro I think multiple calls might be better.

include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
...
+    } else { \
+      RCUTILS_SAFE_FWRITE_TO_STDERR(msg); \
+    } \
+  } while (0)
+
+/// Set the error message to stderr using a format string and format arguments, appending the
+/// previous error.
+/**
+ * This function sets the error message to stderr using the given format string, appending and
+ * resetting the previous error..
```
wjwwood, line 382: Don't use trailing ... just use a period.

include/rcutils/error_handling.h
```diff
@@ -314,6 +314,93 @@ RCUTILS_PUBLIC
 void
 rcutils_reset_error(void);
 
+// Helper macro for nested macro expansion
+#define _RCUTILS_EXPAND(x) x
+
+/// Set the error message, as well as append the current file, line number, and previous error.
```
wjwwood, line 320: I would just say the previous error. And then mention the get error function is used. Otherwise you're repeating its documentation here. And if we ever change what it includes this will then be out of date.

wjwwood (changes requested):

include/rcutils/error_handling.h
methylDragon: To prevent raising an error or aborting in the case where the error isn't set, how about I print this instead?
```
      RCUTILS_SAFE_FWRITE_TO_STDERR("No previous error set\n"); \
```

include/rcutils/error_handling.h
methylDragon: It's not unfortunately :(, I was having a case where something like `MACRO(INNER_MACRO(msg ": %s"))` was becoming `MACRO(msg ": %s")` instead of expanding the `msg` arg.

Stringify almost does that, but causes the expansion to become `"<expanded_msg> ": %s""` (note the double-double quotes)

include/rcutils/error_handling.h
methylDragon: Oops

include/rcutils/error_handling.h
methylDragon: I'll move this to `macros.h`, though

include/rcutils/error_handling.h
wjwwood: I don't think it errors or aborts. I think it just returns a string that says error not set.

wjwwood: https://github.com/ros2/rcutils/blob/5bd1230eb10595cec30168667adf03f48879b96c/include/rcutils/error_handling.h#L305

methylDragon: Ah, I misunderstood what you meant, I'll make the appropriate change

clalancette:
Before going too far forward with this, I think we should talk about the error handling in our C layers in a more general sense.  Mich filed #269 3 years ago to consider this issue.  I think we should come up with a general strategy before we introduce new code here.  Maybe we can talk about it at the ROS 2 weekly tomorrow?

methylDragon:
> Before going too far forward with this, I think we should talk about the error handling in our C layers in a more general sense. Mich filed #269 3 years ago to consider this issue. I think we should come up with a general strategy before we introduce new code here. Maybe we can talk about it at the ROS 2 weekly tomorrow?

@clalancette 

I think we can do that, but some of my other PRs pending the rmw freeze are using code from this PR...

I was thinking that since these are slight extensions to the existing macros, it's not too much of an issue to merge them in first, then consider a more general strategy after the freeze?

clalancette:
> I think we can do that, but some of my other PRs pending the rmw freeze are using code from this PR...

So I thought that everything remaining to be merged was "bonus", no?  If that's the case, then we can leave them out of Iron and target them for J-Turtle.  If that's not the case, can you point to what still needs to get in?

wjwwood:
> If that's not the case, can you point to what still needs to get in?

I think it's the prs attached to https://github.com/ros2/ros2/issues/1405. I think we should try to get them in. If this pr doesn't get in, I think the others can still go in, he'll just have to expand the macros in this pr manually. Which honestly is what is done in the code base today (there are lots of places where we do what these macros do already).

wjwwood:
> Mich filed https://github.com/ros2/rcutils/issues/269 3 years ago to consider this issue. I think we should come up with a general strategy before we introduce new code here.

For what it's worth, the code that these new macros do is already present in the code base, and was my suggestion for what should be done right now: https://github.com/ros2/rcutils/issues/269#issuecomment-675787511

Example of it already in the code base:

https://github.com/ros2/rcl/blob/67919d56192f12c96147bcc1d459f9aeea880b3f/rcl/src/rcl/arguments.c#L341-L345

I think #269 is more about a better way to do this stuff, which I think is a worthwhile thing to discuss, but what's in this pr is just making macros to de-duplicate existing code, at least in my opinion.

clalancette (approved):

PR #427, opened 2023-07-13
Implement thread-safe logging.

### LINKED ISSUE #397: Make logging functionality truly thread-safe
[rcutils_logging_get_logger_level](https://github.com/ros2/rcutils/blob/98d624a455e52c8bdd150c380ca302cf6a8a750c/include/rcutils/logging.h#L344) (which is called when doing any logging in ROS 2) has never been thread-safe with respect to [rcutils_logging_set_logger_level](https://github.com/ros2/rcutils/blob/98d624a455e52c8bdd150c380ca302cf6a8a750c/include/rcutils/logging.h#L392).

However, prior to PR #381, `rcutils_logging_get_logger_level` was thread-safe with respect to itself.  That is, it was only ever reading data, so it was safe for multiple threads to be reading that data at once.

PR #381 changed it so that we cached the results of calculating the logging level and stored that back in the database.  While that is an important optimization, it also caused `rcutils_logging_get_logger_level` to not be thread-safe with respect to itself.  So we disabled that optimization in #393.

However, we really should re-enable that optimization and also make `rcutils_logging_get_logger_level` thread-safe with respect to `rcutils_logging_set_logger_level`.  To do that, what we need to do is to add a reader/writer lock around the hashmap.  That will allow the very common case of multiple threads calling `rcutils_logging_get_logger_level` to continue to work like today (holding a read-lock), while also making it so that the less common case of either setting the log level or updating the cache temporarily blocks readers (holding a write lock).  The wrinkle here is that all of this is in C, so we need to come up with some cross-platform way to do this.  It needs a bit of thought; we may be able to get away with only using atomics here.

clalancette:
This commit introduces thread-safe logging to rcutils. In particular, what it does is to add in a read-write lock around accesses to the g_rcutils_logging_severities_map in logging.c.

This enables two things:
1. Thread-safe writes to the hash map so that get_logger_level and set_logger_level can be called simultaneously on two different threads.
2. The ability to cache lookups during get_logger_effective_level, so we have to do less parsing on every log call.

This introduces the concept of locking into rcutils, which we haven't had before.  However, this particular use seems to be justified, since the logging subsystem can be called outside of the client libraries that could possibly do this locking at a higher level (think about the rmw implementations, for instance).

Note that this introduces a new pthread dependency within rcutils that we've not had before.

If we take this, it will fix #397 

@wjwwood @fujitatomoya @sloretz FYI, I'm interested to hear your thoughts about this.

@JanStaschulat @pablogs9 I'm interested to hear whether taking a pthread dependency like this will cause problems for e.g. rclc.


pablogs9:
Hi @clalancette, thanks for taking us into account.

IMO, this change will break compatibility with the micro-ROS stack for sure. But `rcutils` is a very special package in that sense, there are a lot of functionalities that are not embedded-friendly so we maintain [a fork](https://github.com/micro-ROS/rcutils) that keeps track of the original repo and applies micro-ROS patches.

As far as most of the functionality is in a new .c and .h, it should be easy for us to avoid building them. Also, as far as micro-ROS does not have `rcutils` logging, this won't break any of our core functionalities.

We can make a quick test next week to ensure micro-ROS functionality with this branch.

CC: @Acuadros95

clalancette:
> IMO, this change will break compatibility with the micro-ROS stack for sure. But `rcutils` is a very special package in that sense, there are a lot of functionalities that are not embedded-friendly so we maintain [a fork](https://github.com/micro-ROS/rcutils) that keeps track of the original repo and applies micro-ROS patches.

Ah, I didn't know about that.  Thanks.  (As a side note, it looks like at least some of the changes in those patches could probably be contributed back here.  `set_default_allocator` looks easy enough to take, and the `NO_FILESYSTEM` stuff we could probably take, though I'd want to probably restructure that one a bit.  The other changes I would have to take a closer look at.  Just trying to reduce the number of patches you have to carry around :).)

> As far as most of the functionality is in a new .c and .h, it should be easy for us to avoid building them. Also, as far as micro-ROS does not have `rcutils` logging, this won't break any of our core functionalities.

Good to know, thanks.

> We can make a quick test next week to ensure micro-ROS functionality with this branch.

That would be fantastic, thanks!

wjwwood:
Yeah, this kind of goes against the idea of avoiding use of system calls (env vars, mutex, threading, etc) in the lowest level libraries. I know we're not perfect on that score (especially for env vars), but it's a good guiding goal imo.

It would be better (in my opinion), to have who ever initializes logging to provide a way to handle thread safety, a la dependency injection. On special OS/hardware combos that might mean no synchronization, and in C++ that might be std::mutex, in rclc that might be an OS specific mechanism, in another use case that might be a spin-lock (trading performance for other qualities), etc.

I'll try to give the actual changes a review after lunch, but at least for now that's my general feeling for the approach.

wjwwood:
I would be happy if your rwlock here was a feature of rcutils, and was the default if nothing else was specified, but it was possible to avoid in favor of something else at initialization time of logging.

clalancette:
> It would be better (in my opinion), to have who ever initializes logging to provide a way to handle thread safety, a la dependency injection.

I'll just point out that in our current system, this is tricky.  The problem is that every logging call starts with a call to [RCUTILS_LOGGING_AUTOINIT](https://github.com/ros2/rcutils/blob/025f78fa7c03666b5fa1057af00a32ed8bbd64a0/resource/logging_macros.h.em#L75), so whichever module logs first initializes the logging subsystem.

Maybe we should change that so we call `RCUTILS_LOGGING_AUTOINIT` during `rclcpp::init` or `rclpy.init`.  That would have the benefit that a) we would know exactly when logging is initialized, and b) we would remove some branches from the logging hot-path.  The downside would be that this would be a semantic change, so people using `RCUTILS_LOGGING` macros without calling those init functions would start failing.

Thoughts?

JanStaschulat:
Regarding micro-ROS: rclc is including `<rcutils/logging_macros.h>` for logging and uses the macros like 
`RCUTILS_LOG_DEBUG_NAMED`, `RCUTILS_LOG_INFO`, or ` RCUTILS_LOG_ERROR`. 

Introducting thread-safe logging would not be a problem the for functions in the rclc package. 

However, micro-ROS also supports non-POSIX operating systems, like AUTOSAR or bare-metal, that do not have `pthread` functions or data types, e.g. ` pthread_rwlock_t`. Those would need to be replaced by OS equivalent ones and could be maintained in a micro-ROS fork.

Secondly, this locking could lead to further multi-threading issues, when the [micro-ROS PR for multi-threading support and priority assignment](https://github.com/ros2/rclc/pull/339) will have been merged. Then, we will probably run into priority inversion or deadlocks, if user callbacks run with different operating system priorities. 

See also the pull request for introducing a priority inheritance mutex:

- https://discourse.ros.org/t/avoiding-priority-inversions-includes-draft-pull-requests/29219
- https://github.com/ros2/rcpputils/pull/174

... and the PR to introduce thread creation independent of the operating system: 

- https://discourse.ros.org/t/adding-thread-attributes-configuration-in-ros-2-framework/30701
- https://github.com/ros2/rclcpp/pull/2205

As a general comment, I think this topic is very important also regarding supporting real-time for ROS 2. This does not only require a real-time ROS 2 Executor and static memory Allocator; but also topics like real-time logging, thread creation, priority inheritance mutex, and others need to be addressed. To me, it is currently unclear, how this could be organized efficiently. Many PR have been made by the community, but they need to be orchestrated well to find their place in ROS 2 core.

clalancette:
I've rebased this onto the latest, and also added in a bunch more detection logic here.  It should now only attempt to build a pthread implementation when that is available, though it won't automatically build a "stub" implementation because that is inherently racy.  I'm curious what people think of the latest here.

rodericktaylor:
Would an implementation of lock-free thread-safe logging that uses stdatomic_helper.h instead of pthreads have a chance of passing a review?

It "should" be immune to deadlocks and priority inversion.

PR #430, opened 2023-09-18
use strnlen_s instead of strnlen if available

razr:
use `strnlen_s` instead of `strnlen` if available. VxWorks has only `strnlen_s`

### DIFF
--- a/src/error_handling.c
+++ b/src/error_handling.c
@@ -30,6 +30,7 @@ extern "C"
 #include <rcutils/allocator.h>
 #include <rcutils/macros.h>
 #include <rcutils/strdup.h>
+#include <rcutils/strnlen.h>
 
 // RCUTILS_REPORT_ERROR_HANDLING_ERRORS and RCUTILS_WARN_ON_TRUNCATION are set in the header below
 #include "./error_handling_helpers.h"
@@ -198,7 +199,7 @@ rcutils_set_error_state(
   error_state.line_number = line_number;
 #if RCUTILS_REPORT_ERROR_HANDLING_ERRORS
   // Only warn of overwritting if the new error is different from the old ones.
-  size_t characters_to_compare = strnlen(error_string, RCUTILS_ERROR_MESSAGE_MAX_LENGTH);
+  size_t characters_to_compare = rcutils_strnlen(error_string, RCUTILS_ERROR_MESSAGE_MAX_LENGTH);
   // assumption is that message length is <= max error string length
   static_assert(
     sizeof(gtls_rcutils_error_state.message) <= sizeof(gtls_rcutils_error_string.str),
--- a/src/error_handling_helpers.h
+++ b/src/error_handling_helpers.h
@@ -39,6 +39,7 @@
 #include <string.h>
 
 #include <rcutils/error_handling.h>
+#include <rcutils/strnlen.h>
 
 #ifdef __cplusplus
 extern "C"
@@ -130,7 +131,7 @@ __rcutils_convert_uint64_t_into_c_str(uint64_t number, char * buffer, size_t buf
   buffer[i] = '\0';
 
   // reverse the string in place
-  __rcutils_reverse_str(buffer, strnlen(buffer, 21));
+  __rcutils_reverse_str(buffer, rcutils_strnlen(buffer, 21));
 }
 
 // do not use externally, internal function which is only to be used by error_handling.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,7 @@ set(rcutils_sources
   src/strerror.c
   src/string_array.c
   src/string_map.c
+  src/strnlen.c
   src/testing/fault_injection.c
   src/time.c
   ${time_impl_c}
@@ -293,7 +294,7 @@ if(BUILD_TESTING)
   )
   if(TARGET test_error_handling_helpers)
     target_include_directories(test_error_handling_helpers PUBLIC include)
-    target_link_libraries(test_error_handling_helpers osrf_testing_tools_cpp::memory_tools)
+    target_link_libraries(test_error_handling_helpers ${PROJECT_NAME} osrf_testing_tools_cpp::memory_tools)
   endif()
 
   ament_add_gtest(test_split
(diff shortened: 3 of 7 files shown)


src/error_handling.c
```diff
@@ -25,6 +25,7 @@ extern "C"
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#define __STDC_WANT_LIB_EXT1__ 1
```
fujitatomoya, line 28: is this required to define here? it looks like `rcutils/error_handling.h` already has defined that.

src/error_handling_helpers.h
```diff
@@ -36,6 +36,7 @@
 #endif
 #include <assert.h>
 #include <stdlib.h>
+#define __STDC_WANT_LIB_EXT1__ 1
 #include <string.h>
 
 #include <rcutils/error_handling.h>
```
fujitatomoya, line 42: same here? there is even `#define __STDC_WANT_LIB_EXT1__ 1` in line 35. probably we need to move the `#include <rcutils/error_handling.h>` to above?

fujitatomoya (changes requested):
basically i am not inclined to take this C non-standard for portability, but there are already `__STDC_LIB_EXT1__` so that should not block this PR.

(note) in other way, how about adding `memchr(string, '\0', max_string);`? and that can avoid calling `strnlen_s`? so that we can remove `__STDC_LIB_EXT1__`.

fujitatomoya:
@razr thanks for the PR, DCO is failing. could you address it with comments?

src/error_handling.c
razr: [this example](https://en.cppreference.com/w/c/string/byte/strlen) says that the `define` shall be before the `#include <string.h>`. I'm not sure whether it is good to rely on what was defined in other headers.

````bash
#define __STDC_WANT_LIB_EXT1__ 1
#include <string.h>
#include <stdio.h>
 
int main(void)
{
    const char str[] = "How many characters does this string contain?";
 
    printf("without null character: %zu\n", strlen(str));
    printf("with null character:    %zu\n", sizeof str);
 
#ifdef __STDC_LIB_EXT1__
    printf("without null character: %zu\n", strnlen_s(str, sizeof str));
#endif
}
````

File: src/error_handling_helpers.h (hunk 2 above)
Comment (razr) [reply src/error_handling_helpers.h:42]: I don't have a strong opinion on it. I let you decide. I have looked into [ROS coding guide](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#id3). It is based on the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). And in the Google C++ Style Guide, the header order is as follows:

dir2/foo2.h (a header corresponding to your cpp file)
A blank line
C system headers (more precisely: headers in angle brackets with the .h extension), e.g., <unistd.h>, <stdlib.h>.
A blank line
C++ standard library headers (without file extension), e.g., <algorithm>, <cstddef>.
A blank line
Other libraries' .h files.
A blank line
Your project's .h files.

It looks to me that in all project headers, the other project headers shall be included in the end.

**Comment by razr** [comment]
Sorry, I've missed the DSO, will update it. Thank you for pointing it out.

File: src/error_handling_helpers.h [hunk 3]
```diff
@@ -36,6 +36,7 @@
 #endif
 #include <assert.h>
 #include <stdlib.h>
+#define __STDC_WANT_LIB_EXT1__ 1
```
clalancette, line 39: This is already available a couple of lines earlier, so I don't think we need this here.

```suggestion
```

clalancette (commented):
So I left one thing that I think we should change.

That said, `strnlen_s` doesn't seem to be a widely implemented piece of the C11 standard (it is optional, and apparently most compilers don't implement it).  Regular `strnlen` can also be trivially implemented by calling `memchr(s, '\0', n)` instead.  See https://stackoverflow.com/questions/66346502/which-is-most-standard-strnlen-or-strnlen-s for some details on all of this.

My suggestion here is we do something slightly different, and introduce `rcutils_strnlen`.  It would be implemented by calling `memchr` instead, and then we could call that everywhere (this would also let us remove some conditional code in `include/rcutils/error_handling.h`.  Thoughts?

src/error_handling_helpers.h
razr: @clalancette I like your idea.

src/error_handling_helpers.h
razr: @clalancette I like your idea

clalancette:
I ended up implementing this with `memchr` in 6349f20 .  @razr Can you take a look and see if this will work for you on VxWorks?

razr:
@clalancette it works on VxWorks. LGTM.

clalancette:
> @clalancette it works on VxWorks. LGTM.

All right, sounds good.  Can you please rebase this PR and add a `Signed-off-by` line to your commit?  Once that is done, I'll rebase this on the latest and run CI.

clalancette:
I ended up rebasing this and squashing the commits together.  Given that my later commit basically completely undid @razr's commits, I'm going to go ahead and approve this and run CI on it so we can get this in.

clalancette:
Pulls: https://github.com/ros2/rcutils/pull/430
Gist: https://gist.githubusercontent.com/clalancette/83834db031f5b1ee5c6ff87c2bd587aa/raw/65c2b45e5abaa6beeb5e18909e6c5592172a0521/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16361
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24148/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18222/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3576/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24757/)

ahcorde:
https://github.com/Mergifyio backport kilted jazzy humble

PR #441, opened 2023-11-27
feat: Add human readable date to logging formats

Kaju-Bubanja:


Kaju-Bubanja:
I guess some linters are failing, is there an autoformatter I can run on the diff?

clalancette:
> Some linters are failing, is there an autoformatter I can run on the diff?

Yes, if you run `colcon test --packages-select rcutils` locally, it will run the linters for you.

clalancette:
> but only get unrecognized argument. I checked the LastTest.log but there is nothing there. How can I show the offending file/line?

Try:

```
colcon test --event-handlers console_direct+ --packages-select rcutils
```

You can also look in `log/latest_test/rcutils/stderr.log`, which should also show the output.

Kaju-Bubanja:
> > Some linters are failing, is there an autoformatter I can run on the diff?
> 
> Yes, if you run `colcon test --packages-select rcutils` locally, it will run the linters for you.

What I meant with autoformatter is a tool, which will automatically change the format applying some standard stylesheet, so that the style conforms with the linters. I got the linter to output what it's doing with `--event-handlers console_direct+` as you suggested, but there are a few issues. Is the current workflow indeed to manually fix issues which the linter brings up?

Kaju-Bubanja:
Alright, thank you for the comment. I fixed the linting issues. Any issues remaining, where already here before

include/rcutils/time.h
```diff
@@ -23,6 +23,7 @@ extern "C"
 #endif
 
 #include <stdint.h>
+#include <time.h>
```
clalancette, line 26: This include shouldn't be necessary here; we aren't referencing it in the header file at all.  Maybe this should be moved to the `.c` file instead?

src/time.c
```diff
@@ -46,6 +46,43 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+    RCUTILS_SET_ERROR_MSG("failed to format time point into string as iso8601_date");
+    return RCUTILS_RET_ERROR;
+  }
+
+  time_t now_t = (time_t)(seconds);
+  struct tm * ptm = localtime(&now_t);
+
+  strftime(str, 32, "%Y-%m-%d %H:%M:%S", ptm);
+  int position = 19;
+  if (rcutils_snprintf(&str[position], str_size - position, ".%.3s", str2)) {
```
clalancette, line 78: This doesn't seem to correspond to ISO 8601, now that I look at it: https://en.wikipedia.org/wiki/ISO_8601

That is, there is no provision in ISO 8601 to add in the nanoseconds to the date representation.  That's why there is no representation in `strftime` for it.

There are a couple of paths that we could go down here:
1.  Leave the representation with the nanoseconds applied to the end.  If we choose this, then I think we should rename the token to something like "date_time_with_ns".
2.  Change this to be iso8601 compliant.  If we do this, then I think the token should be named something like "iso8601_date_time".

(there is also the further problem that we don't know that the incoming `str` has enough room for the string, but let's deal with the output format first)

clalancette (changes requested):

src/time.c
Kaju-Bubanja: I prefer: `date_time_with_ms` since I need the ms for my application

src/time.c
```diff
@@ -46,6 +46,44 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+  char str2[21];   // Make sure the buffer is large enough to hold the largest possible uint64_t
+
+  if (rcutils_snprintf(str2, sizeof(str2), "%" PRIu64, nanoseconds)) {
+    RCUTILS_SET_ERROR_MSG("failed to format time point into string as iso8601_date");
+    return RCUTILS_RET_ERROR;
+  }
+
+  time_t now_t = (time_t)(seconds);
+  struct tm ptm = {.tm_year = 2022, .tm_mday = 1};
+  localtime_r(&now_t, &ptm);
```
clalancette, line 75: We should check the return value of `localtime_r`, and return an error if it fails.

Further, I'm not 100% sure but I believe this is going to be a problem on Windows.  I think Windows doesn't have `localtime_r`, so there we are probably going to have to use `localtime` or `localtime_s`.  My suggestion is to do the following for now:

```suggestion
#ifdef _WIN32
    localtime_s(&ptm, &now_t);
#else
    localtime_r(&now_t, &ptm);
#endif
```

(though remember to add in error checking as well)

src/time.c
```diff
@@ -46,6 +46,44 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+  uint64_t nanoseconds = abs_time_point % (1000u * 1000u * 1000u);
+  char str2[21];   // Make sure the buffer is large enough to hold the largest possible uint64_t
+
+  if (rcutils_snprintf(str2, sizeof(str2), "%" PRIu64, nanoseconds)) {
+    RCUTILS_SET_ERROR_MSG("failed to format time point into string as iso8601_date");
+    return RCUTILS_RET_ERROR;
+  }
+
+  time_t now_t = (time_t)(seconds);
+  struct tm ptm = {.tm_year = 2022, .tm_mday = 1};
```
clalancette, line 74: Let's just set these to all 0; 2022 and 1 is really arbitrary.

src/time.c
```diff
@@ -46,6 +46,44 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+  localtime_r(&now_t, &ptm);
+
+  strftime(str, 32, "%Y-%m-%d %H:%M:%S", &ptm);
```
clalancette, line 77: Before we do this, we need to ensure that `str_size` (as passed in) is at least 32 bytes.  Also, we should check the return value from `strftime`.

src/time.c
```diff
@@ -46,6 +46,44 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+  int position = 19;
```
clalancette, line 78:
```suggestion
  static const int position = 19;
```

This could also use a comment explaining why "19", and/or renaming the variable to be more clear.

clalancette (changes requested):

src/time.c
Kaju-Bubanja: I can add the windows change, but [localtime_r](https://en.cppreference.com/w/c/chrono/localtime) doesn't return an error code to check or am I misreading the docs?

clalancette: > I can add the windows change, but [localtime_r](https://en.cppreference.com/w/c/chrono/localtime) doesn't return an error code to check or am I misreading the docs?

It returns an error; see:

> 2-3) copy of the buf pointer, or null pointer on error (which may be a runtime constraint violation or a failure to convert the specified time to local calendar time)

So we need to check for a NULL return.

src/time.c
clalancette: > Should we return an `RCUTILS_RET_ERROR` if str_size passed in is not 32bytes?

At least 32 bytes, yes.

src/time.c
```diff
@@ -46,6 +46,61 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+    return RCUTILS_RET_OK;
+  }
+  // best to abs it to avoid issues with negative values in C89, see:
+  //   https://stackoverflow.com/a/3604984/671658
+  uint64_t abs_time_point = (uint64_t)llabs(*time_point);
+  // break into two parts to avoid floating point error
+  uint64_t seconds = abs_time_point / (1000u * 1000u * 1000u);
+  uint64_t nanoseconds = abs_time_point % (1000u * 1000u * 1000u);
+  // Make sure the buffer is large enough to hold the largest possible uint64_t
+  char nanoseconds_str[21];
```
fujitatomoya, line 67: why we need to allocate largest possible `uint64_t` in this case? should be less than 10?

```suggestion
  char nanoseconds_str[10];
```

fujitatomoya (commented):
@Kaju-Bubanja can you add some test for this?

src/time.c
clalancette: The largest number a uint64 can be (UINT64_MAX) is: 18446744073709551615 .  That is 20 characters long, plus one character for a \0.  While the nanoseconds value probably shouldn't be this long, I think we should be conservative and handle it.  In any case, this is stack memory, so it is extremely cheap and will be freed automatically at the end of the function.

src/time.c
```diff
@@ -46,6 +46,61 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+
+  if (rcutils_snprintf(nanoseconds_str, sizeof(nanoseconds_str), "%" PRIu64, nanoseconds)) {
```
clalancette, line 69: This check isn't correct.  The [documentation for rcutils_snprintf](https://docs.ros.org/en/rolling/p/rcutils/generated/function_snprintf_8h_1aa3d934f197ccb63c08052a27192478ec.html#function-rcutils-snprintf) says:

```
Returns:
    the number of bytes that would have been written given enough space, or
Returns:
    a negative number if there is an error, but unlike _snprintf_s(), -1 is not returned if there is truncation.
```

So we should check that this is > 0.

src/time.c
```diff
@@ -46,6 +46,61 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+rcutils_time_point_value_as_date_string(
+  const rcutils_time_point_value_t * time_point,
+  char * str,
+  size_t str_size)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(time_point, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(str, RCUTILS_RET_INVALID_ARGUMENT);
+  if (0 == str_size) {
+    return RCUTILS_RET_OK;
+  }
```
clalancette, line 59: Out of curiosity, why is it RCUTILS_RET_OK if the str_size is 0?  I would think we could move the check to ensure that `str_size` >= 32 to here, but maybe there is a reason I'm not remember that an empty str_size is OK.

src/logging.c
```diff
@@ -379,6 +390,7 @@ static const token_map_entry_t tokens[] = {
   {.token = "function_name", .handler = expand_function_name},
   {.token = "file_name", .handler = expand_file_name},
   {.token = "time", .handler = expand_time_as_seconds},
+  {.token = "date_time_with_ms", .handler = expand_time_as_date},
```
clalancette, line 393: So this is called `date_time_with_ms`, but as far as I can tell below, we are putting "nanoseconds" into the last bit.  Which is it?  Milliseconds or nanoseconds?

clalancette (changes requested):

src/time.c
Kaju-Bubanja: I don't know, I just copied this from the other functions

src/logging.c
Kaju-Bubanja: We truncate the ns to ms using `".%.3s"` this is much more readable and for my use case I don't need ns precision only ms

src/time.c
Kaju-Bubanja: shouldn't we check if the return value is -1 or 0?

clalancette: Yes, you are right.  We can change the check to `<=0`.

Kaju-Bubanja:
I added some tests but I don't get why the zero time on the jenkins machine is `1969-12-31 16:00:00` instead of `1970-01-01 01:00:00`. Anybody got an idea how to fix this? And also what timepoint I should use for the test? Is 100ns fine?

src/time.c
```diff
@@ -39,13 +39,68 @@ rcutils_time_point_value_as_nanoseconds_string(
   if (0 == str_size) {
     return RCUTILS_RET_OK;
   }
-  if (rcutils_snprintf(str, str_size, "%.19" PRId64, *time_point) < 0) {
+  if (rcutils_snprintf(str, str_size, "%.19" PRId64, *time_point) <= 0) {
```
fujitatomoya, line 42: why this change is needed? i think negative value will be returned in any error state. it depends on `vsnprintf` which possibly returns zero? besides, this change is not related to this PR enhancement?

src/time.c
```diff
@@ -39,13 +39,68 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+#endif
+
+  if (str_size < 32 || strftime(str, 32, "%Y-%m-%d %H:%M:%S", &ptm) == 0) {
+    RCUTILS_SET_ERROR_MSG("failed to format time point into string as iso8601_date");
+    return RCUTILS_RET_ERROR;
+  }
+  static const int date_end_position = 19;
+  if (rcutils_snprintf(
+      &str[date_end_position], str_size - date_end_position, ".%.3s",
+      nanoseconds_str) <= 0)
```
fujitatomoya, line 95:
```suggestion
      nanoseconds_str) < 0)
```

fujitatomoya (changes requested):
a couple of comments.

src/time.c
Kaju-Bubanja: @clalancette what do you think, I thought you agreed that checking <= 0 makes sense. When we write only zero bytes that seems like an error aswell

src/time.c
Kaju-Bubanja: Whoops, that was a search and replace problem, will fix it together with the below comment

src/time.c
clalancette: Looking at the source for [rcutils_vsnprintf](https://github.com/ros2/rcutils/blob/0b8c10148a832f7ac5be0a9abdc10e7b73e4c51f/src/snprintf.c#L39-L72), it returns `-1` if it runs into an error (like the format being NULL).  Otherwise, at least on Linux, it returns the value that `vsnprintf` returns.  Looking at the man page for that, it says:

```
       The  functions  snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was trun‐
       cated due to this limit, then the return value is the number of characters (excluding the terminating null byte) which would have been  written  to
       the  final  string  if enough space had been available.  Thus, a return value of size or more means that the output was truncated.  (See also below
       under NOTES.)

       If an output error is encountered, a negative value is returned.
```

So again on an error, a negative is returned.  So we definitely need to check for the negative value.

The `0` value as far as I can tell can't happen.  That's because of [this line](https://github.com/ros2/rcutils/blob/0b8c10148a832f7ac5be0a9abdc10e7b73e4c51f/src/snprintf.c#L55), where if the buffer_size was 0, we would have returned an error.

So I think that @fujitatomoya is right here and we only need the `<`.

fujitatomoya (approved):
lgtm

fujitatomoya:
@clalancette @ahcorde friendly ping.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20040/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14503/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20768/)

Kaju-Bubanja:
@clalancette @ahcorde friendly ping.

ahcorde (changes requested):
tests are failing : `rcutils.TestTimeFixture.test_rcutils_time_point_value_as_date_string`

Kaju-Bubanja:
> I added some tests but I don't get why the zero time on the jenkins machine is `1969-12-31 16:00:00` instead of `1970-01-01 01:00:00`. Anybody got an idea how to fix this? And also what timepoint I should use for the test? Is 100ns fine?

Yeah because of this. Not really my domain

fujitatomoya:
> Yeah because of this. Not really my domain

@Kaju-Bubanja i think we need to set timezone explicitly during this test?

Kaju-Bubanja:
> > Yeah because of this. Not really my domain
> 
> @Kaju-Bubanja i think we need to set timezone explicitly during this test?

I checked [tzinfo](https://man7.org/linux/man-pages/man3/tzset.3.html) which says:

>        The tzset() function initializes the tzname variable from the TZ
>        environment variable.  This function is automatically called by
>        the other time conversion functions that depend on the timezone.
>        In a System-V-like environment, it will also set the variables
>        timezone (seconds West of UTC) and daylight (to 0 if this
>        timezone does not have any daylight saving time rules, or to
>        nonzero if there is a time, past, present, or future when
>        daylight saving time applies).
> 
>        If the TZ variable does not appear in the environment, the system
>        timezone is used.  The system timezone is configured by copying,
>        or linking, a file in the [tzfile(5)](https://man7.org/linux/man-pages/man5/tzfile.5.html) format to /etc/localtime.  A
>        timezone database of these files may be located in the system
>        timezone directory (see the FILES section below).
> 
>        If the TZ variable does appear in the environment, but its value
>        is empty, or its value cannot be interpreted using any of the
>        formats specified below, then Coordinated Universal Time (UTC) is
>        used.
> 
>        The value of TZ can be one of two formats.  The first format is a
>        string of characters that directly represent the timezone to be
>        used:

So it looks like the runner needs to add the TZ variable and ideally set it to UTC. I don't know how to do that or if I even have the rights to change the environment variables of the runner. It would be great to get this useful feature into ROS. But I think somebody with more knowledge and maybe access needs to help us get this going.

test/test_time.cpp
```diff
@@ -287,6 +287,37 @@ TEST_F(TestTimeFixture, test_rcutils_time_point_value_as_nanoseconds_string) {
   EXPECT_STREQ("-0000000000000000100", buffer);
 }
 
+// Tests the rcutils_time_point_value_as_date_string() function.
+TEST_F(TestTimeFixture, test_rcutils_time_point_value_as_date_string) {
+  rcutils_ret_t ret;
+  rcutils_time_point_value_t timepoint;
+  char buffer[256] = "";
+
```
fujitatomoya, line 295: > So it looks like the runner needs to add the TZ variable and ideally set it to UTC. I don't know how to do that or if I even have the rights to change the environment variables of the runner. It would be great to get this useful feature into ROS. But I think somebody with more knowledge and maybe access needs to help us get this going.

how about this?

```suggestion

  // change timezone into UTC for the test process
  ASSERT_TRUE(rcutils_set_env("TZ", "UTC"));
```

clalancette: I'm going to suggest we do something different, and just use `rcutils_system_time_now` as the timepoint.  We won't be able to compare against a known string, but we can at least validate that it has the parts we expect (i.e. YYY-MM-DD HH:MM::SS.ss), and it sidesteps this whole issue (which probably won't work on Windows anyway).

Kaju-Bubanja: I like this idea, I will get to it soonish

test/test_time.cpp
```diff
@@ -287,6 +287,38 @@ TEST_F(TestTimeFixture, test_rcutils_time_point_value_as_nanoseconds_string) {
...
+  // Typical use case.
+  timepoint = 100;
+  ret = rcutils_time_point_value_as_date_string(&timepoint, buffer, sizeof(buffer));
+  EXPECT_EQ(RCUTILS_RET_OK, ret) << rcutils_get_error_string().str;
+  struct tm tm;
+  ASSERT_TRUE(strptime(buffer, "%Y-%m-%d %H:%M:%S", &tm) != nullptr);
```
fujitatomoya, line 301: Does this work on Windows? i guess no?

Kaju-Bubanja: What is the supposed way to support both platforms? Is it to write the code using platform specific instructions and guard with preprocessor directives? Or can we use boost?

clalancette: > Or can we use boost?

We can't use boost, no.  It is too heavy of a dependency.

Given that this is in a test, and we can use C++ there, we can probably do something with https://en.cppreference.com/w/cpp/io/manip/get_time .

Kaju-Bubanja: Ok, the last version uses get_time which should work on Windows as well

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20330/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14741/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21074/)

Kaju-Bubanja:
I think it's ready @clalancette @fujitatomoya

fujitatomoya:
@Kaju-Bubanja thanks for the fix!

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20369/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14777/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21119/)

src/time.c
```diff
@@ -46,6 +46,61 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+
+  if (rcutils_snprintf(nanoseconds_str, sizeof(nanoseconds_str), "%" PRIu64, nanoseconds) < 0) {
+    RCUTILS_SET_ERROR_MSG("failed to format time point nanoseconds into string");
+    return RCUTILS_RET_ERROR;
+  }
+
+  time_t now_t = (time_t)(seconds);
+  struct tm ptm = {.tm_year = 0, .tm_mday = 0};
+#ifdef _WIN32
+  if (localtime_s(&now_t, &ptm) == NULL) {
```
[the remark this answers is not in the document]
Kaju-Bubanja: Ofc win and Linux have source/dest switched around :D

Kaju-Bubanja:
@fujitatomoya Can I do that too, I wonder:

CI:

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20377/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14786/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21127/)

src/time.c
```diff
@@ -46,6 +46,61 @@ rcutils_time_point_value_as_nanoseconds_string(
...
+
+  if (rcutils_snprintf(nanoseconds_str, sizeof(nanoseconds_str), "%" PRIu64, nanoseconds) < 0) {
+    RCUTILS_SET_ERROR_MSG("failed to format time point nanoseconds into string");
+    return RCUTILS_RET_ERROR;
+  }
+
+  time_t now_t = (time_t)(seconds);
+  struct tm ptm = {.tm_year = 0, .tm_mday = 0};
+#ifdef _WIN32
+  if (localtime_s(&ptm, &now_t) == NULL) {
```
clalancette, line 77: MSVC is saying:

```
'==': 'errno_t' differs in levels of indirection from 'void *'
```

Indeed, looking up the documentation for [localtime_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s?view=msvc-170) shows that it returns an `errno_t`, not a pointer.  So at least this needs to be fixed.

clalancette (changes requested):
Besides the comment inline, Windows testing is failing in this code: https://ci.ros2.org/job/ci_windows/21127/testReport/junit/(root)/projectroot/test_time/

Kaju-Bubanja:
> CI:
> 
>     * Linux [Build Status](http://ci.ros2.org/job/ci_linux/20377/)
> 
>     * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14786/)
> 
>     * Windows [Build Status](http://ci.ros2.org/job/ci_windows/21127/)

Can I see these jobs myself, so to not loop you in unecessarily each time I fixed something? Because the checks which are running below always show 3 successful checks. Are these CI runs manually run by you and you link to the results or are they automatically run?

clalancette:
> Are these CI runs manually run by you and you link to the results or are they automatically run?

They are unfortunately run manually.

Kaju-Bubanja:
Thank you guys, happy we got this useful feature into rolling :)

ros-discourse:

https://discourse.ros.org/t/ros-2-jazzy-jalisco-released/37862/1

PR #442, opened 2023-12-05
Add rcutils_static_assert to abstract platform

mjcarroll:
Some compiler combinations seem to have an issue with `static_assert`.  In this case, I'm specifically running into the `gcc12` version that comes by default with conda not having the macro defined at all.  Similar issues have been reported in other places (https://github.com/ggerganov/llama.cpp/issues/2024) and patched here: https://github.com/RoboStack/ros-humble/blob/main/patch/ros-humble-rcutils.patch

This seems like a cleaner approach by just making a universal RCUTILS macro and defining it as necessary.

### DIFF
--- a/include/rcutils/error_handling.h
+++ b/include/rcutils/error_handling.h
@@ -121,8 +121,7 @@ typedef struct rcutils_error_state_s
 } rcutils_error_state_t;
 
 // make sure our math is right...
-#if __STDC_VERSION__ >= 201112L
-static_assert(
+RCUTILS_STATIC_ASSERT(
   sizeof(rcutils_error_string_t) == (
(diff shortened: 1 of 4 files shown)


mjcarroll:
After discussing with @clalancette, this may be an artifact of setting the C version to GNU11 vs ISOC11, which causes the `static_assert` macro to be undefined on this particular gcc version.  I'm changing this to draft until I can fully investigate.

wolfv:
Thanks for writing this up @mjcarroll - just ran into this when building [`ros-jazzy`](https://github.com/RoboStack/ros-jazzy) for robostack :) But so far it's looking quite good!

PR #426, opened 2023-07-11
make escape characters work

iuhilnehc-ynos:
to address https://github.com/ros2/rcutils/issues/425

The test cases depends on https://github.com/ros2/launch/pull/717 .

NOTE: As the color code are different on various platforms, this PR is not intended to implement it.

PR #443, opened 2023-12-14
Allow parsing of escape sequence in log format

SammyRamone:
This is an extendion of https://github.com/ros2/rcutils/pull/426 which allowed the usage of control codes (i.e. \a, \b, \n, \r, \t) in the log format. This PR adds parsing of the escape character \x1b (also known as \033 or \e).
The motivation is that this allows using [Select Graphic Rendition parameters](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters).
Their usage allows setting of bold, faint, italic and/or underline to make the logging output of the screen better readable. This is especially interesting to highlight parts of the logging output (the actual message) while giving less focus to additional information (e.g. the function name, file name or line number).

Example: 
`RCUTILS_CONSOLE_OUTPUT_FORMAT="\x1b[4m{name}\x1b[24m \x1b[1m{message}\x1b[0m \x1b[2m{function_name} \x1b[3m{file_name}:{line_number} \x1b[0m"`
grafik

Side node: 
I chose the '\x1b' representation instead of the '\e' since this would result in the following compiler warning:
`warning: non-ISO-standard escape sequence, '\e'`
This makes the parsing a bit more complex since we have a 4 character input that is parsed to a 2 character output (while the other control codes are 2->1).
I just decided on this approach to ensure sticking to the ISO standard. However, it would also be fine for me to use '\e'.

I will also add an update to the documentation if the PR is accepted.

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -87,8 +87,10 @@ bool g_rcutils_logging_initialized = false;
 static char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
 static const char * g_rcutils_logging_default_output_format =
   "[{severity}] [{time}] [{name}]: {message}";
+#ifdef _WIN32
 static DWORD g_original_console_mode = 0;
 static bool g_consol_mode_modified = false;
+#endif 
 
 static rcutils_allocator_t g_rcutils_logging_allocator;
 
@@ -793,9 +795,11 @@ rcutils_ret_t rcutils_logging_shutdown(void)
   g_num_log_msg_handlers = 0;
   g_rcutils_logging_initialized = false;
 
+  #ifdef _WIN32
   if (g_consol_mode_modified){
     SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
   }
+  #endif
   return ret;
 }
 
--- a/src/logging.c
+++ b/src/logging.c
@@ -440,14 +440,25 @@ static void create_format_string(
       break;
     } else {
       const char * expected_char = NULL;
-      switch (logging_output_format_string[i + back_slash_index + 1]) {
-        case 'a':  expected_char = "\a"; break;  // alert
-        case 'b':  expected_char = "\b"; break;  // backspace
-        case 'n':  expected_char = "\n"; break;  // new line
-        case 'r':  expected_char = "\r"; break;  // carriage return
-        case 't':  expected_char = "\t"; break;  // horizontal tab
-        default:
-          break;
+      int skip_chars = 0;
+
+      if (logging_output_format_string[i + back_slash_index + 1] == 'x' &&
+        logging_output_format_string[i + back_slash_index + 2] == '1' &&
+        logging_output_format_string[i + back_slash_index + 3] == 'b')
+      {
+        // detect escape sequence
+        expected_char = "\x1b";
+        skip_chars = 2;
+      } else {
+        switch (logging_output_format_string[i + back_slash_index + 1]) {
+          case 'a':  expected_char = "\a"; break;  // alert
+          case 'b':  expected_char = "\b"; break;  // backspace
+          case 'n':  expected_char = "\n"; break;  // new line
+          case 'r':  expected_char = "\r"; break;  // carriage return
+          case 't':  expected_char = "\t"; break;  // horizontal tab
+          default:
+            break;
+        }
       }
 
       if (expected_char) {
@@ -466,12 +477,12 @@ static void create_format_string(
         // copy the decoded character
         g_rcutils_logging_output_format_string[dest_buffer_index] = expected_char[0];
         dest_buffer_index += 1;
-        start_offset += 2;
+        start_offset += 2 + skip_chars;
       } else {
         start_offset_previous_not_copy += (back_slash_index + 2);
       }
 
-      i += (back_slash_index + 2);
+      i += (back_slash_index + 2 + skip_chars);
     }
   }
 }
--- a/src/logging.c
+++ b/src/logging.c
@@ -90,7 +90,7 @@ static const char * g_rcutils_logging_default_output_format =
 #ifdef _WIN32
 static DWORD g_original_console_mode = 0;
 static bool g_consol_mode_modified = false;
-#endif 
+#endif
 
 static rcutils_allocator_t g_rcutils_logging_allocator;
 
@@ -423,19 +423,16 @@ static const char * copy_from_orig(
 #define ACTIVATE_VIRTUAL_TERMINAL_PROCESSING() \
   { \
     HANDLE std_error_handle = GetStdHandle(STD_ERROR_HANDLE); \
-    if (std_error_handle == INVALID_HANDLE_VALUE) \
-    { \
-        RCUTILS_SET_ERROR_MSG("Could not get error handle to activating virtual terminal."); \
-        return; \
+    if (std_error_handle == INVALID_HANDLE_VALUE) { \
+      RCUTILS_SET_ERROR_MSG("Could not get error handle to activating virtual terminal."); \
+      return; \
     } \
-    if (! GetConsoleMode(std_error_handle, &g_original_console_mode)) \
-    { \
+    if (!GetConsoleMode(std_error_handle, &g_original_console_mode)) { \
       RCUTILS_SET_ERROR_MSG("Could not get consol mode to activating virtual terminal."); \
       return; \
     } \
     DWORD newDwMode = g_original_console_mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING; \
-    if (! SetConsoleMode(std_error_handle, newDwMode)) \
-    { \
+    if (!SetConsoleMode(std_error_handle, newDwMode)) { \
       RCUTILS_SET_ERROR_MSG("Could not set consol mode to activating virtual terminal."); \
       return; \
     } \
@@ -796,7 +793,7 @@ rcutils_ret_t rcutils_logging_shutdown(void)
   g_rcutils_logging_initialized = false;
 
   #ifdef _WIN32
-  if (g_consol_mode_modified){
+  if (g_consol_mode_modified) {
     SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
   }
   #endif
--- a/src/logging.c
+++ b/src/logging.c
@@ -87,6 +87,8 @@ bool g_rcutils_logging_initialized = false;
 static char g_rcutils_logging_output_format_string[RCUTILS_LOGGING_MAX_OUTPUT_FORMAT_LEN];
 static const char * g_rcutils_logging_default_output_format =
   "[{severity}] [{time}] [{name}]: {message}";
+static DWORD g_original_console_mode = 0;
+static bool g_consol_mode_modified = false;
 
 static rcutils_allocator_t g_rcutils_logging_allocator;
 
@@ -415,6 +417,33 @@ static const char * copy_from_orig(
   return logging_output->buffer;
 }
 
+#ifdef _WIN32
+#define ACTIVATE_VIRTUAL_TERMINAL_PROCESSING() \
+  { \
+    HANDLE std_error_handle = GetStdHandle(STD_ERROR_HANDLE); \
+    if (std_error_handle == INVALID_HANDLE_VALUE) \
+    { \
+        RCUTILS_SET_ERROR_MSG("Could not get error handle to activating virtual terminal."); \
+        return; \
+    } \
+    if (! GetConsoleMode(std_error_handle, &g_original_console_mode)) \
+    { \
+      RCUTILS_SET_ERROR_MSG("Could not get consol mode to activating virtual terminal."); \
+      return; \
+    } \
+    DWORD newDwMode = g_original_console_mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING; \
+    if (! SetConsoleMode(std_error_handle, newDwMode)) \
+    { \
+      RCUTILS_SET_ERROR_MSG("Could not set consol mode to activating virtual terminal."); \
+      return; \
+    } \
+    g_consol_mode_modified = true; \
+  }
+#else
+// nothing todo for non-windows platform
+#define ACTIVATE_VIRTUAL_TERMINAL_PROCESSING()
+#endif
+
 // copy buffers and decode escape characters if they exist
 static void create_format_string(
   const char * logging_output_format_string)
@@ -447,7 +476,10 @@ static void create_format_string(
         logging_output_format_string[i + back_slash_index + 3] == 'b')
       {
         // detect escape sequence
+        ACTIVATE_VIRTUAL_TERMINAL_PROCESSING();
         expected_char = "\x1b";
+        // the 4 char long "\x1b" string literal will become a 2 char long \x1b escape sequence
+        // therefore we need to skip forward in parsing the output format string
         skip_chars = 2;
       } else {
         switch (logging_output_format_string[i + back_slash_index + 1]) {
@@ -760,6 +792,10 @@ rcutils_ret_t rcutils_logging_shutdown(void)
   }
   g_num_log_msg_handlers = 0;
   g_rcutils_logging_initialized = false;
+
+  if (g_consol_mode_modified){
+    SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
+  }
   return ret;
 }
 
--- a/test/test_logging_output_format.py
+++ b/test/test_logging_output_format.py
@@ -91,6 +91,16 @@ def generate_test_description():
     ))
     processes_to_test.append(name)
 
+    env_escape_sequence = dict(os.environ)
+    # This custom output is to check that escape characters work correctly.
+    env_escape_sequence['RCUTILS_CONSOLE_OUTPUT_FORMAT'] = \
+        r'{name} \x1b[0m {message}'
+    name = 'test_logging_output_format_escape_sequence'
+    launch_description.add_action(ExecuteProcess(
+        cmd=[executable], env=env_escape_sequence, name=name, output='screen'
+    ))
+    processes_to_test.append(name)
+
     launch_description.add_action(
         launch_testing.actions.ReadyToTest()
     )
--- a/test/test_logging_output_format.py
+++ b/test/test_logging_output_format.py
@@ -127,7 +127,8 @@ class TestLoggingOutputFormatAfterShutdown(unittest.TestCase):
                     path=os.path.join(os.path.dirname(__file__), process_name),
                     encoding='unicode_escape'
                 ),
-                process=process_name
+                process=process_name,
+                strip_ansi_escape_sequences=False
             )
 
     def test_processes_exit_codes(self, proc_info):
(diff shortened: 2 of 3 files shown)


sloretz:
@clalancette mind looking at how these escape characters would be handled on Windows? 🧇

clalancette:
So it turns out that these sequences do indeed work on Windows, but with a caveat.

The caveat is that you need to enable the processing of these sequences first.  In particular, you need to call a sequence similar to the following:

```
HANDLE myhandle = GetStdHandle(STD_ERROR_HANDLE);
DWORD dwMode = 0;
GetConsoleMode(myhandle, &dwMode);
DWORD newDwMode = dwMode;
newDwMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
SetConsoleMode(myhandle, newDwMode);
```

(obviously I've omitted error-checking and the like).  So far, so good; we can easily make a function to do that.  I think we should probably call this only when we see a '\x1b' sequence, and otherwise leave it disabled.

The hard part is disabling that virtual terminal processing during shutdown.  I guess maybe we can track what we've done above, and then disable it during `rcutils_logging_shutdown`.

So with those changes, I'd be happy to take this as a new feature.

src/logging.c
```diff
@@ -466,12 +477,12 @@ static void create_format_string(
         // copy the decoded character
         g_rcutils_logging_output_format_string[dest_buffer_index] = expected_char[0];
```
fujitatomoya, line 478: if expected_char is `\x1b`, do we need to copy the index 1 and 2 here?

fujitatomoya (changes requested):
i think we need to have test cases in addition to  https://github.com/ros2/rcutils/pull/443#issuecomment-1871243943

src/logging.c
SammyRamone: No we don't. The `\x1b` is only 4 chars long in its written literal (input) representation. When parsed as a escape sequence it is only 2 chars long. Similar to `\n` which is 2 chars long in the written representation but only 1 char long in its representation as a newline command. I added a comment to make this more clear.

SammyRamone:
> So it turns out that these sequences do indeed work on Windows, but with a caveat.
> 
> The caveat is that you need to enable the processing of these sequences first. In particular, you need to call a sequence similar to the following:
> 
> ```
> HANDLE myhandle = GetStdHandle(STD_ERROR_HANDLE);
> DWORD dwMode = 0;
> GetConsoleMode(myhandle, &dwMode);
> DWORD newDwMode = dwMode;
> newDwMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
> SetConsoleMode(myhandle, newDwMode);
> ```
> 
> (obviously I've omitted error-checking and the like). So far, so good; we can easily make a function to do that. I think we should probably call this only when we see a '\x1b' sequence, and otherwise leave it disabled.
> 
> The hard part is disabling that virtual terminal processing during shutdown. I guess maybe we can track what we've done above, and then disable it during `rcutils_logging_shutdown`.
> 
> So with those changes, I'd be happy to take this as a new feature.

It took me a while to get a Windows System with ROS2 but I implemented the proposed solution and this PR should now also work on Windows systems.

SammyRamone:
> i think we need to have test cases in addition to [#443 (comment)](https://github.com/ros2/rcutils/pull/443#issuecomment-1871243943)

I tried implementing a test in my previous commit. However, the test always keeps failing, although the feature itself works fine. I spend quiet some time trying to get the test to run, but I was able to do so. I don't have any other ideas where to look for. @fujitatomoya could you maybe take a look at the test and see if you can identify what is going wrong?

src/logging.c
```diff
@@ -749,6 +792,10 @@ rcutils_ret_t rcutils_logging_shutdown(void)
   }
   g_num_log_msg_handlers = 0;
   g_rcutils_logging_initialized = false;
+
+  if (g_consol_mode_modified){
+    SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
+  }
```
fujitatomoya, line 798: these seem the Windows specific code? probably we do not need to have this `if` statement for Non-Windows platform at all, so are the static global variables?

this actually failed to build with Linux,

```bash
Finished <<< rosidl_generator_dds_idl [0.36s]
--- stderr: rcutils                                  
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c:90:8: error: unknown type name ‘DWORD’
   90 | static DWORD g_original_console_mode = 0;
      |        ^~~~~
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c: In function ‘rcutils_logging_shutdown’:
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c:797:5: warning: implicit declaration of function ‘SetConsoleMode’ [-Wimplicit-function-declaration]
  797 |     SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
      |     ^~~~~~~~~~~~~~
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c:797:20: warning: implicit declaration of function ‘GetStdHandle’ [-Wimplicit-function-declaration]
  797 |     SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
      |                    ^~~~~~~~~~~~
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c:797:33: error: ‘STD_ERROR_HANDLE’ undeclared (first use in this function)
  797 |     SetConsoleMode(GetStdHandle(STD_ERROR_HANDLE), g_original_console_mode);
      |                                 ^~~~~~~~~~~~~~~~
/root/ros2_ws/colcon_ws/src/ros2/rcutils/src/logging.c:797:33: note: each undeclared identifier is reported only once for each function it appears in
gmake[2]: *** [CMakeFiles/rcutils.dir/build.make:222: CMakeFiles/rcutils.dir/src/logging.c.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:319: CMakeFiles/rcutils.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
gmake: *** [Makefile:146: all] Error 2
---
Failed   <<< rcutils [2.08s, exited with code 2]
```

SammyRamone: Ah sorry, I forgot to check again if it still builds on Linux after fixing the Windows version. I fixed it in the last commit.

test/test_logging_output_format.py
```diff
@@ -91,6 +91,16 @@ def generate_test_description():
     ))
     processes_to_test.append(name)
 
+    env_escape_sequence = dict(os.environ)
+    # This custom output is to check that escape characters work correctly.
+    env_escape_sequence['RCUTILS_CONSOLE_OUTPUT_FORMAT'] = \
+        r'{name} \x1b[0m {message}'
```
fujitatomoya, line 97: console outputs are,

```bash
6: [test_logging_output_format_escape_sequence-7] name1  XxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxX
6: [test_logging_output_format_escape_sequence-7] name2  X42xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxX
```

they do not match the expected output,

```bash
6: ======================================================================
6: FAIL: test_logging_output (rcutils.TestLoggingOutputFormatAfterShutdown)
6: Test all executables output against expectations.
6: ----------------------------------------------------------------------
6: Traceback (most recent call last):
6:   File "/root/ros2_ws/colcon_ws/src/ros2/rcutils/test/test_logging_output_format.py", line 124, in test_logging_output
6:     launch_testing.asserts.assertInStderr(
6:   File "/root/ros2_ws/colcon_ws/build/launch_testing/launch_testing/asserts/assert_output.py", line 139, in assertInStderr
6:     assertInStream(proc_output, expected_output, process, cmd_args, output_filter=output_filter,
6:   File "/root/ros2_ws/colcon_ws/build/launch_testing/launch_testing/asserts/assert_output.py", line 103, in assertInStream
6:     assert False, "Did not find '{}' in output for any of the matching processes: {}".format(
6: AssertionError: Did not find '[re.compile('name1 \x1b\\[0m Xx{2045}X'), re.compile('name2 \x1b\\[0m X42x{2043}X')]' in output for any of the matching processes: test_logging_output_format_escape_sequence-7
```

i did not allocated to get to the bottom but i believe we need a similar change with https://github.com/ros2/launch/pull/717/files for encoding to `launch_testing`?

fujitatomoya (commented):
besides, i think this need to rebase.

test/test_logging_output_format.py
SammyRamone: Thank you very much for the hint. This lead me to finding out that the ANSI escape sequences are by default striped from the strings. I disabled this for this test and now the test works correctly.

SammyRamone:
I fixed the test and also manually tested the final version both on Linux and Windows. 
I also rebased the commits to clean them up.
Furthermore, I added documentation about this feature in https://github.com/ros2/ros2_documentation/pull/4061

If there is anything else that should be changed please tell me.

fujitatomoya (approved):
lgtm with green CI

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14648/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20949/)

fujitatomoya:
@SammyRamone unfortunately windows still failing, https://ci.ros2.org/job/ci_windows/20949/testReport/projectroot/Release/test_logging_output_format/. could you check that?

SammyRamone:
> @SammyRamone unfortunately windows still failing, https://ci.ros2.org/job/ci_windows/20949/testReport/projectroot/Release/test_logging_output_format/. could you check that?

I had a look at it and I am not sure why this test works fine on Linux but not Windows. I will further investigate, but I might need some time since I have a lot of other things to do right now and this seems like it will take some time. It might be related to the usage of the raw string but I need to do some testing.

SammyRamone:
@fujitatomoya I changed the test to use a normal string instead of the raw string. The test now runs successfully in Linux and Windows. I spend some time researching why this makes a difference in Windows but not in Linux, but I have no idea. This environment is just passed on to the standard Python `subprocess.popen` method. Maybe Windows is interpreting the '\' differently as it also uses this in paths 🤷 
@ahcorde could you please rerun the CI jobs to make sure that this does not only work on my test system

fujitatomoya:
@SammyRamone thank you for taking time to work on this. i missed that one, but it aligns with other test cases. lgtm with green CI.

ahcorde (approved):

PR #444, opened 2024-01-23
Make rcutils_split() return RCUTILS_RET_BAD_ALLOC if alloc fails

christophebedard:
`rcutils_split()`'s documentation says that it returns `RCUTILS_RET_BAD_ALLOC` if memory allocation fails, but it doesn't. Fix it and update the tests.

Also, mention `RCUTILS_RET_BAD_ALLOC` in `rcutils_split_last()`'s doc.

Note that there only seems to be one user of `rcutils_split()` in the ROS 2 core (`rcl_get_discovery_static_peers()`) and it doesn't seem to be affected by this change. I doubt it would change a caller's behaviour that much anyway.

### DIFF
--- a/src/split.c
+++ b/src/split.c
@@ -118,7 +118,7 @@ fail:
   }
 
   RCUTILS_SET_ERROR_MSG("unable to allocate memory for string array data");
-  return RCUTILS_RET_ERROR;
+  return RCUTILS_RET_BAD_ALLOC;
 }
 
 rcutils_ret_t
--- a/include/rcutils/split.h
+++ b/include/rcutils/split.h
@@ -52,6 +52,7 @@ rcutils_split(
  * \param[in] allocator for allocating new memory for the output array
  * \param[out] string_array with the split tokens
  * \return #RCUTILS_RET_OK if successful, or
+ * \return #RCUTILS_RET_BAD_ALLOC if memory allocation fails, or
  * \return #RCUTILS_RET_ERROR if an unknown error occurs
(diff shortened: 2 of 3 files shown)


ahcorde (approved):

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14560/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20833/)

wjwwood (approved):

src/split.c
```diff
@@ -118,7 +118,7 @@ rcutils_split(
   }
 
   RCUTILS_SET_ERROR_MSG("unable to allocate memory for string array data");
-  return RCUTILS_RET_ERROR;
+  return RCUTILS_RET_BAD_ALLOC;
```
fujitatomoya, line 121: @christophebedard line 105-106 (last trailing token) should be checked if that allocation fails, that also should go to `RCUTILS_RET_BAD_ALLOC`?

christophebedard: Yep! I noticed it as well in our internal code. I'll create a PR soon.

clalancette: > Yep! I noticed it as well in our internal code. I'll create a PR soon.

It might be worthwhile to just refactor it to solve the TODO on line 67, which I think will also mostly handle this case.

PR #445, opened 2024-01-24
Use rcutils_string_array_init in rcutils_split & handle alloc fail

christophebedard:
Resolve the TODO by using `rcutils_string_array_init()` for the initial allocation in `rcutils_split()`, and properly handle allocation near the end of the function. Also, update the tests.

Kind of a follow-up to #444

### DIFF
--- a/src/split.c
+++ b/src/split.c
@@ -42,7 +42,6 @@ rcutils_split(
     *string_array = rcutils_get_zero_initialized_string_array();
     return RCUTILS_RET_OK;
   }
-  string_array->allocator = allocator;
 
   size_t string_size = strlen(str);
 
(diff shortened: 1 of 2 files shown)


src/split.c
```diff
@@ -64,9 +64,7 @@ rcutils_split(
       ++string_array->size;
     }
   }
-  // TODO(wjwwood): refactor this function so it can use rcutils_string_array_init() instead
-  string_array->data = allocator.allocate(string_array->size * sizeof(char *), allocator.state);
-  if (NULL == string_array->data) {
+  if (rcutils_string_array_init(string_array, string_array->size, &allocator) != RCUTILS_RET_OK) {
```
clalancette, line 67: It's kind of weird to me that we setup `string_array->size`, and then pass it into `rcutils_string_array_init`, which sets it back to itself.  It is not wrong, per-se, but just reads strangely.

I might suggest that we use a stack variable to compute the size, and then pass that into `rcutils_string_array_init`.  That will end up with the same result, but read better in my opinion.

clalancette: Relatedly, we can remove the assignment of `string_array->allocator` at line 45, since `rcutils_string_array_init` will do that as well.

clalancette (changes requested):

src/split.c
christophebedard: done!

fujitatomoya (approved):
lgtm @christophebedard thanks 👍

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14563/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20836/)

PR #453, opened 2024-02-15
Change 'ROS2' to 'ROS 2' in quality declaration

christophebedard:
For the first modified line, I used the same line as some other quality declaration documents and thus simply avoided mentioning "the ROS 2 developer guide" at all.

Also, update the branch name in a link.


fujitatomoya (approved):
lgtm

christophebedard:
Since this only changes the quality declaration file, I think we can skip CI.

PR #455, opened 2024-03-06
validate the allocator before use.

fujitatomoya:
minor patch to avoid possible segmentation fault in rcutils function scope.

related to https://github.com/ros2/rcl_logging/pull/116

### DIFF
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -179,7 +179,6 @@ rcutils_join_path(
   const char * right_hand_path,
   rcutils_allocator_t allocator)
 {
-  RCUTILS_CHECK_ALLOCATOR(&allocator, return NULL);
   if (NULL == left_hand_path) {
     return NULL;
   }
@@ -198,7 +197,6 @@ rcutils_to_native_path(
   const char * path,
   rcutils_allocator_t allocator)
 {
-  RCUTILS_CHECK_ALLOCATOR(&allocator, return NULL);
   if (NULL == path) {
     return NULL;
   }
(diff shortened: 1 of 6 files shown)


fujitatomoya:
CC: @clalancette @cottsay

cottsay:
A few thoughts:
1. I don't think we should check this value in a function unless the function specifically uses it. If it just passes it through to another function, that other function should be responsible for checking it.
2. This is pretty nitpicky, but IIRC the convention is/was to check arguments for validity after checking for the presence of required arguments (i.e. null checks before value checks)
3. In general, I support more argument checking like this, but these are pretty low-level functions and they do get called pretty often. I don't think we'll see any noticeable performance impact, but we should be mindful.

fujitatomoya (commented):
@cottsay thank you for your comment https://github.com/ros2/rcutils/pull/455#issuecomment-1981694132. i fixed the code accordingly.

fujitatomoya:
note:

- although we do not have particular performance profiler, some places are considered as performance concern. if the check is during initialization and finalization, that should be no problem.
- assertion debug macro would be useful during the test to replace those checks.
- for performance sensitive execution paths, we can remove the check and add it to the doc section, `allocate must not be null, otherwise unexpected behavior`.

fujitatomoya:
@ahcorde can you take a look at this?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20756/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15055/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21435/)

PR #457, opened 2024-03-19
Ros2 pr

drensber:
As the commit message says, these are changes that are mostly related to use in resource constrained systems (but shouldn't cause any harm when used in larger platforms).


drensber:
> Can you please rebase this onto the latest? It looks like there are unrelated changes in here.

Sorry, you're right. A bunch of changes from micro-ROS were dragged long (probably because I originally forked their fork of your repo, but GitHub won't let me "refork" from yours).

drensber:
> Can you please rebase this onto the latest? It looks like there are unrelated changes in here.

OK, after I rebased and pushed, the set of changes looks like the ones I meant to submit.

src/logging.c
```diff
@@ -1369,6 +1377,8 @@ rcutils_ret_t rcutils_logging_format_message(
   }
 # define SET_STANDARD_COLOR_IN_STREAM(is_colorized, status)
 #endif
+static  char msg_buf[1024] = ""; 
+static  char output_buf[1024] = "";
```
clalancette, line 1381: This is not going to work, unfortunately.

Because these are now global variables, two threads calling into `rcutils_logging_console_output_handler` will trample all over each others messages.  And we definitely have multiple threads that call the handler at the same time.  So I don' think we can reasonably make these global.

What's the original probably you are trying to solve?

clalancette (changes requested):

src/logging.c
drensber: I think the problem that we were trying to solve was just to not have to allocate the buffers on the stack for microcontroller implementations where stack sizes are very limited. I'm double checking with our developer who wrote the code. Is the other change to disable the colors and eliminate the dependency no isatty() for Zephyr builds OK?

clalancette: > Is the other change to disable the colors and eliminate the dependency no isatty() for Zephyr builds OK?

Yeah, I have no real problem with that (just a couple of nits, which I'll add reviews for once we figure out this issue).

drensber: It sounds like the problem that we were solving (it was a real problem, not just a theoretical one) was that the stack allocation of the buffer pretty quickly blows through any reasonable sized stack on a Zephyr/microcontroller instance. I don't think the thread safety was an issue on micro-ROS, because there aren't multiple threads calling the function. We could possibly solve it by allocating on the heap for each invocation of the function, or we could possibly just make this a micro-ROS-specific change (the micro-ROS developers seem to be hesitant to do that, but this seems like a case that warrants it, IMHO).

clalancette: > It sounds like the problem that we were solving (it was a real problem, not just a theoretical one) was that the stack allocation of the buffer pretty quickly blows through any reasonable sized stack on a Zephyr/microcontroller instance.

All right, that was my suspicion, but good to have it confirmed.

For "normal" systems, without a very restricted stack, I would really prefer to keep using the stack allocations.  It is faster than allocating from the heap, the CPU time is more deterministic, and it still works even if the system has run out of heap (which is not really a problem on Linux, but ironically might be more of a problem in a microcontroller environment).

For systems with very small stacks, we could either a) use a smaller stack allocation, or b) use the heap, as you suggest.  If you want to go with b), then we have to come up with a way so that we use the stack by default, but switch over to a heap allocation in certain situations.  If you do decide to give that a try, then I'll suggest some refactoring here so we don't have `#if` sections littered throughout the code.

drensber: I don't really love the idea of heap allocation either (especially on slower microcontrollers where the non-determinism of heap allocation time could be even more significant). We could also just do it the way I did it here, but keep that on the micro-ROS fork (and verify that they have no intentions of ever making this multi-threaded in micro-ROS (might not hurt to have a comment explicitly stating that it's not thread safe too).

fujitatomoya: @drensber as you mentioned, you can make a PR against https://github.com/micro-ROS/rcutils? i think micro-ROS would be the best place to take this change?

drensber: Hi. I removed those changes and will either address them in the microROS fork or continue to include them in a patch for my own purposes.  The other changes that remove the dependency on isatty() for Zephyr and that eliminate the warning for the unused zero_semantic and one_semantic variables are still in this PR.

drensber: I don't understand where all of these extra changes in my changeset came from. My original PR only had changes to 1 file, and then I scaled it back further to remove the parts of the changeset dealing with the log buffer allocation. Suddenly, the the changeset grew to include 25 files.

clalancette: You'll have to rebase onto the `rolling`, or merge in from `rolling`.  That should sort it.

drensber: THat's what I just did but it still looks weird and has a lot of unrelated changes that aren't mine. Can I just close this PR and start over? I think it would be easier.

clalancette: Sure.

PR #460, opened 2024-04-16
Creating More Accuracy in NS_TO_S

CursedRock17:
This Pull Request is done in preparation for [this geometry2 issue](https://github.com/ros2/geometry2/issues/68) which reutilizes logic increasing accuracy in the conversion from nanoseconds to seconds. I assumed we wanted to keep these as Macros opposed to creating functions, but this would mean the Macros continue to take multiple "types" when we just want nanoseconds (int64_t). So should they stay as macros and we just remove all tests which don't take an int64 type and begin enforcing this behavior?

### DIFF
--- a/include/rcutils/time.h
+++ b/include/rcutils/time.h
@@ -36,7 +36,10 @@ extern "C"
 #define RCUTILS_US_TO_NS(microseconds) ((microseconds) * 1000LL)
 
 /// Convenience macro to convert nanoseconds to seconds.
-#define RCUTILS_NS_TO_S(nanoseconds) ((nanoseconds) / (1000LL * 1000LL * 1000LL))
+#define RCUTILS_NS_TO_S(nanoseconds) \
+((1e-9 * (double)((int32_t)(nanoseconds % 1000000000l))) + \
+((double)((int32_t)((nanoseconds - ((int32_t)(nanoseconds % 1000000000l))) / 1000000000l))))
+
 /// Convenience macro to convert nanoseconds to milliseconds.
 #define RCUTILS_NS_TO_MS(nanoseconds) ((nanoseconds) / (1000LL * 1000LL))
 /// Convenience macro to convert nanoseconds to microseconds.
--- a/test/test_time.cpp
+++ b/test/test_time.cpp
@@ -94,11 +94,6 @@ TEST_F(TestTimeFixture, test_rcutils_time_conversion_macros) {
   EXPECT_EQ(RCUTILS_NS_TO_S(1000000000ll), 1ll);  // int64_t
   EXPECT_EQ(RCUTILS_NS_TO_S(1000000042ll), 1ll);  // int64_t (truncated)
   EXPECT_EQ(RCUTILS_NS_TO_S(-1999999999ll), -1ll);  // int64_t (truncated)
-  EXPECT_EQ(RCUTILS_NS_TO_S(200000000.), 0.2);  // double
-  EXPECT_EQ(RCUTILS_NS_TO_S(1.0 + 1.0), 0.000000002);  // sum of doubles
-  EXPECT_EQ(
-    RCUTILS_NS_TO_S(9007199254740992.),
-    9007199.254740992);  // maximum precision double (53 bits)
 
   // nanoseconds to milliseconds
   EXPECT_EQ(RCUTILS_NS_TO_MS(1000000ll), 1ll);  // int64_t


ahcorde (changes requested):
Looks good to me with green CI.

Let's put this on the queue to merge after the freeze is lifted.

ahcorde (approved):

include/rcutils/time.h
```diff
@@ -36,7 +36,10 @@ extern "C"
 #define RCUTILS_US_TO_NS(microseconds) ((microseconds) * 1000LL)
 
 /// Convenience macro to convert nanoseconds to seconds.
-#define RCUTILS_NS_TO_S(nanoseconds) ((nanoseconds) / (1000LL * 1000LL * 1000LL))
+#define RCUTILS_NS_TO_S(nanoseconds) \
+  ((1e-9 * (double)((int32_t)(nanoseconds % 1000000000l))) + \
+  ((double)((int32_t)((nanoseconds - ((int32_t)(nanoseconds % 1000000000l))) / 1000000000l))))
```
fujitatomoya, line 41: just curious, is there any accuracy difference from this relying on FPU? i did not check the assembly files though.

```suggestion
#define RCUTILS_NS_TO_S(nanoseconds) ((double)(nanoseconds) / (1e9))
```

the following sample gives me the exact same floating value so far.

```cpp
#include <stdio.h>
#include <stdint.h>

#define NANO_TO_SEC(nanoseconds) ((double)(nanoseconds) / 1e9)

#define RCUTILS_NS_TO_S(nanoseconds)                         \
  ((1e-9 * (double)((int32_t)(nanoseconds % 1000000000l))) + \
  ((double)((int32_t)((nanoseconds - ((int32_t)(nanoseconds % 1000000000l))) / 1000000000l))))

int main() {
  int64_t nanoseconds = 1234567890123456789; // Example nanoseconds
  double seconds = NANO_TO_SEC(nanoseconds);
  printf("Nanoseconds: %ld \nSeconds: %lf\n", nanoseconds, seconds);

  seconds = RCUTILS_NS_TO_S(nanoseconds);
  printf("Nanoseconds: %ld \nSeconds: %lf\n", nanoseconds, seconds);

  return 0;
}
```

CursedRock17: [This `geometry2` issue](https://github.com/ros2/geometry2/issues/34) should explain the whole reasoning as to why this improves accuracy, but pretty much scaling each set of nanoseconds, prevents excess rounding, in case the user needs additional accuracy. I'm almost positive `printf` will do some rounding for visual help, but technically if you make this test:

```
 EXPECT_EQ(RCUTILS_NS_TO_S(-1999999999ll), -1.999999999);  // int64_t
```
You'll get a result of:

>  Expected equality of these values:
        ((1e-9 * (double)((int32_t)(-1999999999ll % 1000000000l))) + ((double)((int32_t)((-199999999       9ll - ((int32_t)(-1999999999ll % 1000000000l))) / 1000000000l))))
          Which is: -2
        -2

When you should've just [left this](https://github.com/ros2/rcutils/blob/431e7f6cd19fe207247b1619ff12c17401339cfa/test/test_time.cpp#L96)

clalancette:
I think that before we merge this, we should have a companion PR in `geometry2` (or somewhere else) that actually uses it.  Otherwise, we can run into the situation where we have an API that nothing uses, which isn't great.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20946/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15238/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21620/)

ahcorde (changes requested):
CI is failing

CursedRock17:
I'm going to roll out a fix soon, I assume the MACRO is returning `long` instead of a `double`, but it's weird that it's not throwing an error on my local tests, which require doubles as results. Also, is there something I missing with the CI having semi-random `K's` in the error message?
> [K/home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/qos.cpp:145:[Kerro[Kinvalid operands of type[Kdou[K’ an[Klong [K’ to binar[Koperat[K’

clalancette:
> It seems like the error is not on our part, but rather rmw_fastrtps. [This call to RCUTILS_NS_TO_S](https://github.com/ros2/rmw_fastrtps/blob/c48406a5f87fd7ebe5a576fbf627b6c9efcc786e/rmw_fastrtps_shared_cpp/src/qos.cpp#L145) is causing the problem where it calls incorrect types between double to long, because the nanoseconds that are being passed, are in the form of a double already. The line before is:

> double period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2.0 / 3.0;

Agreed with your analysis; that line of code should probably be:

```
int64_t period_in_ns = entity_qos.liveliness().lease_duration.to_ns() * 2 / 3;
```

(though there is some risk of losing a bit of precision there).

If you'd like to open a PR over at `rmw_fastrtps` to change that, we can review it over there and unblock this.

include/rcutils/time.h
clalancette, line 41: See my comment in https://github.com/ros2/rmw_fastrtps/pull/755 on why I'm very nervous about making this change.  Basically things that were expecting integer division before are now going to get FPU division.  While it may be more accurate, it also changes the expectations.

This actually leads to another observation, which is that the type of `nanoseconds` is ill-defined because this is a macro.

All of that leads me to another thought.  Maybe what we should do here is to make a new function called `double rcutils_nanoseconds_to_seconds(double)`.  And then we (slowly) convert uses of the macro over to that new function.  That way we aren't making this change for things that don't expect it, and we can get the types right on the new function.

Thoughts?

clalancette (changes requested):

include/rcutils/time.h
CursedRock17: > While it may be more accurate, it also changes the expectations.

Yeah as we found in [this review](https://github.com/ros2/rcutils/pull/460/files#r1573147454), changing the nature of the macro induces errors in testing. To be fair, it's a beneficial change, but if it causes errors in testing there's a high chance it could lead to breakages on the users' end. Though I don't think going the route of new typed function is a better option as it just adds more to the file and alters the way code is written, when I would assume users just want to use a macro, design-wise. Unless most of the macros are rebranded to typed functions, I would feel that it's just an awkward change when it comes to DX.

> This actually leads to another observation, which is that the type of nanoseconds is ill-defined because this is a macro.

Yeah while [rcutils](https://github.com/ros2/rcutils/blob/cf3e9986a557e4c893eb718b90c77ce189494a11/include/rcutils/time.h#L48) and [dds implementations](https://github.com/eProsima/Fast-DDS/blob/8cb447bb25032f646685c923da866b733fd78aff/src/cpp/rtps/common/Time_t.cpp#L114) use `int64_t` it's not necessarily common knowledge for users.

I would like to create more accuracy, but since there's not a effective way to deprecate macros, I doubt it's worth breaking user behavior to do so, nor is it worth expanding on.

mjcarroll: > I would like to create more accuracy, but since there's not a effective way to deprecate macros, I doubt it's worth breaking user behavior to do so, nor is it worth expanding on.

I think that adding more accurate function calls (especially if they are constexpr) and migrating where it makes sense would work?

clalancette: > I think that adding more accurate function calls (especially if they are constexpr) and migrating where it makes sense would work?

Yeah, that's what I'm thinking we should do.

CursedRock17: > (especially if they are constexpr) and migrating where it makes sense would work?

Seems fair enough to make more calls, but since `rcutils` is essentially C repository, did we need these split functions in a different repo?

clalancette: > Seems fair enough to make more calls, but since `rcutils` is essentially C repository, did we need these split functions in a different repo?

Good point.  It can't be constexpr here, but it can at least be a function, be const, and use types.

Alternatively, if we wanted to use C++ and constexpr, we could move it to `rcpputils`, but then there are certain places it can't be used (like `rcutils`, `rcl`, `rcl_action`, and `rmw`).

CursedRock17:
In response to [this comment](https://github.com/ros2/rcutils/pull/460#discussion_r1671245700) I went with the first option of keeping everything within `rcutils` as to broaden these new functions to as many repositories as possible as and continue to solve the original issue. I went ahead and made functions of all the macros that magnify the time as that's where most of our lost precision comes from, I felt the other 3 macros don't necessarily need functions of the their own (more bloat in the file). 

As I was breaking down the tests I would run into oddball tests in which a value would be too large for the original `uint32_t` conversion, but after removing the casts to said type, all of my tests were passing with the correct values to pass. After digging the original logic comes from [here](https://github.com/televator/roscpp_core/commit/b3269c96c0e831028fe4296cb1fb07bdf80121da#diff-ebee6258ac942f57a13230dfb5f2298a6785a16684fcc9eb18a077e168bb7a19) with no way to find out why said [conversions](https://github.com/televator/roscpp_core/blob/742eb23194404e2b435564dc9a5d9e8a0dbded6a/rostime/include/ros/time.h#L154) were used. Just wanted to know if my analysis is wrong here

src/time.c
```diff
@@ -28,6 +28,48 @@ extern "C"
 #include "rcutils/error_handling.h"
 #include "rcutils/snprintf.h"
 
+double
+rcutils_nanoseconds_to_seconds(const int64_t nanoseconds)
+{
+  // scale the nanoseconds separately for improved accuracy
+  int64_t sec, nsec;
+  nsec = (nanoseconds % 1000000000l);
```
fujitatomoya, line 36: just a minor suggestion. can we have like `RCUTILS_NANOSECS_PER_SEC` for `1000000000` and use it else where? e.g `RCUTILS_S_TO_NS`, `RCUTILS_NS_TO_S`. the same thing can go to miliseconds and microseconds.

CursedRock17: Yeah seems like a great metric for users to have, especially if they design their own time-based algorithms and want standardization. I would figure it'll get used enough to be worth having 3 definitions.

PR #469, opened 2024-06-04
Fix filesystem iteration on Windows

clalancette:
While looking through the logs of https://ci.ros2.org/view/nightly/job/nightly_win_deb/3114/ , I noticed that there was an "access violation".  This didn't actually cause any tests to fail, but it is a bit weird so I decided to investigate.

What I found was that we were unnecessarily calling `FindClose` on Windows during `rcutils_dir_iter_next` when we failed to find another entry.  Because we always call `rcutils_dir_iter_close` after use of the iterator, which *also* calls `FindClose`, we were running into UB on Windows.  Fix that here by just removing the `FindClose` during `rcutils_dir_iter_next`.

While I was in here, I noticed that we were unnecessarily doing a bunch of casts from void pointers to the state pointers.  But we can just use forward declaration for that, and remove the casts.

### DIFF
--- a/include/rcutils/filesystem.h
+++ b/include/rcutils/filesystem.h
@@ -245,6 +245,8 @@ RCUTILS_PUBLIC
 size_t
 rcutils_get_file_size(const char * file_path);
 
+struct rcutils_dir_iter_state_s;
+
 /// An iterator used for enumerating directory contents
 typedef struct rcutils_dir_iter_s
 {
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21214/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15523/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21935/)

PR #478, opened 2024-09-06
Fix setting allocator to NULL.

### LINKED ISSUE #477: rcutils_get_zero_initialized_string_array has (potential) data race
hello, during thread sanitizer test, I've found multi thread calling into rcutils_get_zero_initialized_string_array gives data race warnings from thread sanitizer

https://github.com/ros2/rcutils/blob/rolling/src/string_array.c#L35

```c
rcutils_string_array_t
rcutils_get_zero_initialized_string_array(void)
{
  static rcutils_string_array_t array = {
    .size = 0,
    .data = NULL,
  };
  array.allocator = rcutils_get_zero_initialized_allocator();  //.  <---- thread sanitizer report data race this line because caller on some other thread may read it, but every threading running this line is write to it, without synchronize
  return array;
}
```
this may behave correctly on some architect, but still a data race in C standard

clalancette:
That is, previously the code was setting the allocator at runtime, not during static initialization.  Switch to static initialization, which should fix a possible race.

This should fix #477 .

A draft until we get confirmation from the reporter that it fixes the issue.

### DIFF
--- a/src/char_array.c
+++ b/src/char_array.c
@@ -25,9 +25,15 @@ rcutils_get_zero_initialized_char_array(void)
     .buffer = NULL,
     .owns_buffer = true,
     .buffer_length = 0u,
-    .buffer_capacity = 0u
+    .buffer_capacity = 0u,
+    .allocator = {
+      .allocate = NULL,
+      .deallocate = NULL,
+      .reallocate = NULL,
+      .zero_allocate = NULL,
+      .state = NULL
+    }
   };
-  char_array.allocator = rcutils_get_zero_initialized_allocator();
   return char_array;
 }
 
--- a/src/char_array.c
+++ b/src/char_array.c
@@ -21,19 +21,7 @@
 rcutils_char_array_t
 rcutils_get_zero_initialized_char_array(void)
 {
-  static rcutils_char_array_t char_array = {
-    .buffer = NULL,
-    .owns_buffer = true,
-    .buffer_length = 0u,
-    .buffer_capacity = 0u,
-    .allocator = {
-      .allocate = NULL,
-      .deallocate = NULL,
-      .reallocate = NULL,
-      .zero_allocate = NULL,
-      .state = NULL
-    }
-  };
+  static rcutils_char_array_t char_array = {0};
   return char_array;
 }
 
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -65,13 +65,7 @@ __default_zero_allocate(size_t number_of_elements, size_t size_of_element, void 
 rcutils_allocator_t
 rcutils_get_zero_initialized_allocator(void)
 {
-  static rcutils_allocator_t zero_allocator = {
-    .allocate = NULL,
-    .deallocate = NULL,
-    .reallocate = NULL,
-    .zero_allocate = NULL,
-    .state = NULL,
-  };
+  static rcutils_allocator_t zero_allocator = {0};
   return zero_allocator;
 }
 
--- a/src/array_list.c
+++ b/src/array_list.c
@@ -44,7 +44,7 @@ typedef struct rcutils_array_list_impl_s
 rcutils_array_list_t
 rcutils_get_zero_initialized_array_list(void)
 {
-  static rcutils_array_list_t zero_initialized_array_list = {NULL};
+  static rcutils_array_list_t zero_initialized_array_list = {0};
   return zero_initialized_array_list;
 }
 
--- a/src/hash_map.c
+++ b/src/hash_map.c
@@ -77,7 +77,7 @@ int rcutils_hash_map_string_cmp_func(const void * val1, const void * val2)
 rcutils_hash_map_t
 rcutils_get_zero_initialized_hash_map(void)
 {
-  static rcutils_hash_map_t zero_initialized_hash_map = {NULL};
+  static rcutils_hash_map_t zero_initialized_hash_map = {0};
   return zero_initialized_hash_map;
 }
 
--- a/src/shared_library.c
+++ b/src/shared_library.c
@@ -56,18 +56,7 @@ C_ASSERT(sizeof(char) == sizeof(TCHAR));
 rcutils_shared_library_t
 rcutils_get_zero_initialized_shared_library(void)
 {
-  static rcutils_shared_library_t zero_initialized_shared_library = {
-    .library_path = NULL,
-    .lib_pointer = NULL,
-    .allocator = {
-      .allocate = NULL,
-      .deallocate = NULL,
-      .reallocate = NULL,
-      .zero_allocate = NULL,
-      .state = NULL
-    }
-  };
-
+  static rcutils_shared_library_t zero_initialized_shared_library = {0};
   return zero_initialized_shared_library;
 }
 
--- a/src/shared_library.c
+++ b/src/shared_library.c
@@ -56,10 +56,18 @@ C_ASSERT(sizeof(char) == sizeof(TCHAR));
 rcutils_shared_library_t
 rcutils_get_zero_initialized_shared_library(void)
 {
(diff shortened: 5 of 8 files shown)


clalancette:
Pulls: https://github.com/ros2/rcutils/pull/478
Gist: https://gist.githubusercontent.com/clalancette/0d8c51702f95917fc2b7fdbafa436af3/raw/e7af239fbc2785887fc1e6ee1f32787191ab063f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14537
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21813/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16084/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1316/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22514/)

fujitatomoya (commented):
@clalancette i agree with you, this should fix the warning.

if this works, can we also apply the same patch for `rcutils_get_zero_initialized_char_array` and `rcutils_get_zero_initialized_uint8_array`?

clalancette:
> if this works, can we also apply the same patch for `rcutils_get_zero_initialized_char_array` and `rcutils_get_zero_initialized_uint8_array`?

Good call.  I also found that `rcutils_get_zero_initialized_shared_library` didn't use this pattern at all, so I've now converted it over as well.  With that done, and confirmation that this fixes the sanitizer, I'm going to switch this to a "real" PR and run full CI on it.

clalancette:
Pulls: https://github.com/ros2/rcutils/pull/478
Gist: https://gist.githubusercontent.com/clalancette/f02ff4e3f3fc6d8241350f341deeea9f/raw/e7af239fbc2785887fc1e6ee1f32787191ab063f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14544
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21821/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16091/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1323/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22521/)

fujitatomoya (approved):
lgtm!

src/char_array.c
```diff
@@ -25,9 +25,15 @@ rcutils_get_zero_initialized_char_array(void)
...
+    .buffer_capacity = 0u,
+    .allocator = {
+      .allocate = NULL,
+      .deallocate = NULL,
+      .reallocate = NULL,
+      .zero_allocate = NULL,
+      .state = NULL
+    }
   };
-  char_array.allocator = rcutils_get_zero_initialized_allocator();
```
fujitatomoya, line 30: @clalancette i may be mistaken but i was thinking that maybe we should use stack memory and call `rcutils_get_zero_initialized_allocator`? i think this is not good approach for the structure maintenance. please see https://github.com/ros2/rmw/pull/378#discussion_r1751214891 too.

fujitatomoya (changes requested):

src/char_array.c
fujitatomoya: okay, i think having all members static variable can work the most effectively.

> The problem with doing it on the stack is that it reintroduces the race that we are originally trying to fix. In point of fact, that was exactly what the previous code did.

i was thinking a few things.

- specifying all member variables is not well-maintained code, this is already discussed above.
- remove `static` above, so that it can create the zero init structure with calling member specific zero init function in stack and then return the copy of that. this also should solve the racy condition. (compared to static initialization, this is less performance, but since this is initialization, i thought that was okay and we can rely on the each structure initialization call.)
- some member variables have `default` (not zero initialized) setting. the following code is actually initialized all 0, but having default setting (can be changed and updated in the future) in zero initialized call is mislead. So i think this is another issue.

https://github.com/ros2/rmw/blob/ac4f9afccca3dc305f2b519fdb0fb9131250f2a5/rmw/src/init_options.c#L35

clalancette: > * remove `static` above, so that it can create the zero init structure with calling member specific zero init function in stack and then return the copy of that. this also should solve the racy condition. (compared to static initialization, this is less performance, but since this is initialization, i thought that was okay and we can rely on the each structure initialization call.)

Ah, I see, I misunderstood your previous comment.  Yeah, doing it as a stack variable would fix both the thread problems and make it still work.  Since we've used the static initialization to this point, I'd prefer to just keep using that, but I don't feel that strongly about it.

> * some member variables have `default` (not zero initialized) setting. the following code is actually initialized all 0, but having default setting (can be changed and updated in the future) in zero initialized call is mislead. So i think this is another issue.

Yeah, agreed.  That looks like a bug.

methylDragon: I'd prefer static for performance reasons, especially since these structs and data types get used a ton everywhere.

clalancette: Yeah, agreed with @methylDragon here.

With that all said, I think I just need an approval on this now, and I can go ahead and merge it.  Thanks!

fujitatomoya (approved):
thanks for advice, lgtm.

clalancette:
The latest CI here was successful (the yellow builds were known flakes).  Given that, I'm going to go ahead and merge, thanks everyone!

clalancette:
Actually, arg.  I'm very sorry.  We agreed to put this in the BSS via *not* specifying all of the fields.  Thus, I'm actually going to change this to get rid of the explicit initialization.  Changes upcoming.

clalancette:
Pulls: https://github.com/ros2/rcutils/pull/478
Gist: https://gist.githubusercontent.com/clalancette/a386de13a9a1af1cc5804a72f192a0c1/raw/e7af239fbc2785887fc1e6ee1f32787191ab063f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14556
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21833/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16103/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1335/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22541/)

fujitatomoya (approved):
lgtm with green CI

clalancette:
Hm, the failures here are ones I haven't seen before.  So I'm going to have to look at it closer.

clalancette:
This has actually exposed a bug in rcl.  I'll open up a PR for that which will need to be merged first.

clalancette:
All right, this now depends on https://github.com/ros2/rcl/pull/1182 going in first.

clalancette:
Pulls: https://github.com/ros2/rcutils/pull/478
Gist: https://gist.githubusercontent.com/clalancette/1e055d6a7e132f4fd947d2d7bb367895/raw/e7af239fbc2785887fc1e6ee1f32787191ab063f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14573
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16120/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1356/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22561/)

clalancette:
All right, with the other fixes in, this is now clean.  Going ahead and merging this one, thanks for the review!

PR #458, opened 2024-04-03
Remove dependency on isatty() for Zephyr

drensber:
The Zephyr OS doesn't have an isatty() function, so that dependency needs to be conditionally removed for Zephyr.

PR #508, opened 2025-07-23
feat: ability to override the default allocator

russkel:
## Description

Added a function to set the default allocator, and modified the `rcutils_get_default_allocator` function to return it if it is set and valid.

I needed this for use with Unreal Engine in order to expose UE's allocator API to ROS 2 so that the UE garbage cleaner could free ROS 2 objects successfully.

I have also seen references to a similar function in another PR (https://github.com/ros2/rcutils/pull/458/files#diff-8352803836a54313eb08fa96e6b6590eae0cfad976ae3187ae45d322d6f50551R101), so I assume I am not the only one to have added this functionality.

Yes, an additional function in the rcutils allocator API.

No.

### DIFF
--- a/include/rcutils/allocator.h
+++ b/include/rcutils/allocator.h
@@ -108,6 +108,21 @@ RCUTILS_WARN_UNUSED
 rcutils_allocator_t
 rcutils_get_default_allocator(void);
 
+/// Override the default allocator returned by rcutils_get_default_allocator.
+/**
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] override_allocator The allocator to set as the default.
+ */
+RCUTILS_PUBLIC
+void
+rcutils_set_default_allocator(rcutils_allocator_t override_allocator);
+
 /// Return true if the given allocator has non-null function pointers.
 /**
  * \param[in] allocator to be checked by the function
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -69,9 +69,23 @@ rcutils_get_zero_initialized_allocator(void)
   return zero_allocator;
 }
 
+static rcutils_allocator_t rcutils_override_default_allocator = {0};
+
+void
+rcutils_set_default_allocator(rcutils_allocator_t override_allocator)
+{
+  if (rcutils_allocator_is_valid(&override_allocator)) {
+    rcutils_override_default_allocator = override_allocator;
+  }
+}
+
 rcutils_allocator_t
 rcutils_get_default_allocator(void)
 {
+  if (rcutils_allocator_is_valid(&rcutils_override_default_allocator)) {
+    return rcutils_override_default_allocator;
+  }
+
   static rcutils_allocator_t default_allocator = {
     .allocate = __default_allocate,
     .deallocate = __default_deallocate,
@@ -79,6 +93,7 @@ rcutils_get_default_allocator(void)
     .zero_allocate = __default_zero_allocate,
     .state = NULL,
   };
+
   return default_allocator;
 }
 


fujitatomoya (commented):
@russkel thanks for creating PR.

could you tell a bit more how this API is used by the application?

when this API is expected to be called? there are many packages depend on rcutils to get the default allocator such as rclcpp, rclpy, rmw and idl libraries. if this is up to the user application to set the default allocator, some uses default allocator (because they already use the default allocator before this API is called) and some uses use specified default allocator. i am not sure if that is what we want to do here, can you explain the user application behavior and expectation with this API?

i think that would be more useful to allow the user application to pass the allocator argument to be used by each API instead of changing the default allocator for everyone? and if not specified, it falls back to the default allocator as currently some APIs are implemented so?

russkel:
Hi @fujitatomoya,

> could you tell a bit more how this API is used by the application?

Code here https://github.com/Greenroom-Robotics/rclUE/blob/greenroom/Source/rclUE/Private/ROS2Subsystem.cpp#L63

In this use case, I set the allocator after `rclc_support_init` call (this maybe should be before now that I am looking at it again). This is so all the libs allocate using the Unreal Engine memory allocator to resolve double free errors I was getting. In reality I probably only want to set rmw allocator so all the messages from the middleware are freeable by Unreal garbage collector. I will freely admit I am not familiar with the specifics of Unreal/rmw allocation/deallocation/garbage collection, but this did seem to fix those issues.

> when this API is expected to be called? there are many packages depend on rcutils to get the default allocator such as rclcpp, rclpy, rmw and idl libraries.

I wrote this some years ago - but it is intended to set the allocator for all the libraries before they are initialised. If I remember correctly, I was unable to individually set them.

> i think that would be more useful to allow the user application to pass the allocator argument to be used by each API instead of changing the default allocator for everyone? and if not specified, it falls back to the default allocator as currently some APIs are implemented so?

Happy to follow your lead with what you think would be a better solution. Maintaining patches on top of a custom `rcutils` isn't what I want to be doing.

russkel:
I dug through the code again, and I realise this is probably why I did it this way:

https://github.com/ros2/rmw/blob/e6addf2411b8ee8a2ac43d691533b8c05ae8f1b6/rmw/src/allocators.c#L28

I was never able to get `rmw` to use a different allocator because it doesn't seem to use the allocator passed in in the `options` struct.

russkel:
Is there any actions for me on this? Or is this a bigger issue?

russkel:
Friendly ping.

include/rcutils/allocator.h
```diff
@@ -108,6 +108,21 @@ RCUTILS_WARN_UNUSED
...
 rcutils_get_default_allocator(void);
 
+/// Override the default allocator returned by rcutils_get_default_allocator.
+/**
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
```
wjwwood, line 118: I'm a bit concerned about having this be non-thread-safe, because it's unlikely that calling code will have the ability to synchronize with other calling threads and/or with other threads reading the value. My guess is that for this to be useful, it needs to be thread-safe and that to do so, using atomics and no locks (assuming the atomics are lock-free) would be ideal.

src/allocator.c
```diff
@@ -69,16 +69,31 @@ rcutils_get_zero_initialized_allocator(void)
   return zero_allocator;
 }
 
+static rcutils_allocator_t rcutils_override_default_allocator = {0};
+
+void
+rcutils_set_default_allocator(rcutils_allocator_t override_allocator)
+{
+  if (rcutils_allocator_is_valid(&override_allocator)) {
```
wjwwood, line 77: With this check on the setter, is it impossible to reset the default allocator to be the "real default"? I would imagine, you could do something like `rcutils_set_default_allocator({0});`, but not with this check. Maybe it should just assign the allocator value no matter what?

src/allocator.c
```diff
@@ -69,16 +69,31 @@ rcutils_get_zero_initialized_allocator(void)
...
+    rcutils_override_default_allocator = override_allocator;
+  }
+}
+
 rcutils_allocator_t
 rcutils_get_default_allocator(void)
 {
+  if (rcutils_allocator_is_valid(&rcutils_override_default_allocator)) {
+    return rcutils_override_default_allocator;
+  }
```
wjwwood, line 87: Three concerns here:

- Thread-safety: what happens if some other thread calls `rcutils_set_default_allocator()` between lines 85 and 85? Not to mention line 86 being run in the middle of the assignment that happens on line 78.
- Performance: this check will be called potentially many, many times, even in critical code sections, and having a branch here might be expensive. It would be great to either eliminate the check, or use some kind of branch prediction hints, like `__builtin_expect` or something similar. I believe we use this kind of thing in the console logging code in this repository.
- Is this redundant? calling code is likely to check that the allocator is valid and right now the setter for this first checks that the allocator is valid, so this check is probably redundant (unless we remove the check in the setter).

wjwwood (changes requested):
Sorry for taking a very long time to get back to you on this.

I left some technical comments that would need to be addressed, but taking a step back, I actually think you'd be better off overriding the default allocator functions using a mechanism like `LD_PRELOAD` or a similar mechanism, which would set the global allocator very early, and due to its nature, could do so without having to change the existing implementation. This is the approach that allocators like jemalloc recommend (https://github.com/jemalloc/jemalloc/wiki/getting-started).

Setting the allocator so early (dynamic linking step), and never unsetting it, also helps you avoid a potential problem with having the allocators mismatch. For example, consider:

- a routine uses `rcutils_get_default_allocator()` very early in the program (let's say it returns an address `0x01...`), and calls `.allocate()`
- you set the default allocator, which is stored in a different address, like `0x02...`
- the program uses the default allocator (now `0x02...`) to allocate and deallocate memory freely
- the very early routine finally gets cleaned up by calling `rcutils_get_default_allocator()` again and then `.deallocate()` on it
- 🌩️ you run into a problem because that early routine called `.allocate()` on `0x01...` and `.deallocate()` on `0x02...`

It will be very hard to avoid this situation in practice.

The allocator interface here was meant more for use with arena allocators and other special use cases, in which cases you'd create an allocator and then pass it in every where you use it, rather that changing the default allocator.

As you've discovered, it isn't passed all the way down the interface in every situation, due to a variety of technical reasons and in some cases oversights in the code.

For all of these reasons, I'm not convinced this is a good API to add, due to the numerous pitfalls you can run into with it.

russkel:
> I left some technical comments that would need to be addressed, but taking a step back, I actually think you'd be better off overriding the default allocator functions using a mechanism like `LD_PRELOAD` or a similar mechanism, which would set the global allocator very early, and due to its nature, could do so without having to change the existing implementation. This is the approach that allocators like jemalloc recommend (https://github.com/jemalloc/jemalloc/wiki/getting-started).

This is a great hint. Thanks for that.

> For all of these reasons, I'm not convinced this is a good API to add, due to the numerous pitfalls you can run into with it.

Okay, I will not pursue this as it sounds like it wont work nicely and probably will not be merged. I hope the LD_PRELOAD does the trick instead, which is quite a bit simpler.

wjwwood:
Thanks for your understanding. I believe the `LD_PRELOAD` should work (at least for Unix machines), but if it doesn't I can try to help you if you reply here.

PR #526, opened 2025-11-12
Adds {severity_with_color} as a log format option.

alexmaclean6:
## Description

This change adds a new log formatting token option: `{severity_with_color}`. When used in the log format string, the severity level will be colorized in the log output, but the rest of the line will not be affected. If `RCUTILS_COLORIZED_OUTPUT=1`, causing the entire line to be colorized, `{severity_with_color}` behaves the same as `{severity}`.

Example usage:
```
[{severity_with_color}] [{name}]: {message} ({function_name}() at {file_name}:{line_number})
```

### Is this user-facing behavior change?

Yes. It allows users to use `{severity_with_color}` in their log format.

### Did you use Generative AI?

No

PR #541, opened 2026-02-11
Add {short_file_name} as log format option

tonynajjar:
## Description

Add a new `{short_file_name}` format token to the rcutils logging system that outputs only the basename of the source file (e.g., `my_node.cpp`) instead of the full path (e.g., `/opt/ros_ws/src/my_package/src/my_node.cpp`).

The existing `{file_name}` token continues to output the full path as before. The new token uses `strrchr` to find the last path separator (`/` on POSIX, `\` on Windows) and returns everything after it.

### Changed files

- **`src/logging.c`** — Added `expand_short_file_name()` handler and registered the `{short_file_name}` token in the `tokens[]` array
- **`include/rcutils/logging.h`** — Added `short_file_name` to the documented list of available format tokens
- **`test/test_logging_console_output_handler.cpp`** — Added unit tests verifying basename extraction from full paths and passthrough for bare filenames
- **`test/test_logging_output_format.py`** — Added integration test launching with `{short_file_name}:{line_number}` format and validating output
- **`test/test_logging_output_format_short_file_name.txt`** — Expected output fixture for the integration test

Yes. Users can now use `{short_file_name}` in the `RCUTILS_CONSOLE_OUTPUT_FORMAT` environment variable to display only the source file basename in log output, reducing log line length and improving readability.

Yes — GitHub Copilot (Claude Opus 4.6)

### Additional Information

- The implementation handles both POSIX (`/`) and Windows (`\`) path separators via `#ifdef _WIN32`. Tested on Ubuntu Linux but I don't have a way to test on Windows.
- If the file name contains no path separator, the full string is returned unchanged
- No changes to `rcutils_log_location_t` struct — the basename is extracted at format-expansion time
- Docs update PR: https://github.com/ros2/ros2_documentation/pull/6226


tonynajjar:
I'd love to have it backported to Kilted when merged, it doesn't break ABI

tonynajjar:
@ahcorde this is of similar nature as https://github.com/ros2/rcutils/pull/526 so you might want to take it. FYI @kscottz if you want to share with the community

kscottz:
@tonynajjar this slaps, thanks so much. Let me see what I can cook up when I make the social media rounds tomorrow. We haven't had a "what's been merged recently post" perhaps we're past due.

fujitatomoya (commented):
@tonynajjar thanks for the PR.

i have a question about this approach. what is the major background and requirement for this PR?

assuming that full absolute paths in log output can be noisy and hurt readability.
however, I'm concerned that showing only the basename isn't quite enough. in a typical ROS 2 workspace, many packages can have identically named files (main.cpp, node.cpp, etc.), so {short_file_name} alone would make it hard to tell where a log line is actually coming from.
would it be possible to instead show the path relative to the package root? for example, my_package/src/my_node.cpp instead of just my_node.cpp. that would cut out the workspace-specific noise while still keeping enough context to identify the package. something like {relative_file_name} or {package_file_name} might be a better middle ground.
curious to hear your thoughts — what was the use case that motivated going with just the basename?

tonynajjar:
@fujitatomoya you're right, the motivation is to reduce the unnecessary noise and cluttering from the full path (in our case it's quite long)

> in a typical ROS 2 workspace, many packages can have identically named files (main.cpp, node.cpp, etc.), so {short_file_name} alone would make it hard to tell where a log line is actually coming from.

I looked into this and although I first though you have a very good point, it turns out that `{short_file_name}` as implemented will actually show the file where the `RCLCPP_INFO` macro is called, and not the original file where the node is created, which is actually convenient. I would say that typically the files with the actual node implementation (where most macro calls are) have distinct names.

Take this example from Nav2:

with `export RCUTILS_CONSOLE_OUTPUT_FORMAT="[{severity} {time}] [{short_file_name}:{line_number}] [{name}]: {message}"`

I get

```
ros2 run nav2_controller controller_server

[INFO 1770887000.671869363] [controller_server.cpp:45] [controller_server]: Creating controller server
```

although as you can see [here](https://github.com/ros-navigation/navigation2/blob/c72993c6f63aca7d6ac9723de6e730c8fb5e695e/nav2_controller/CMakeLists.txt#L57), the `controller_server` node is started in `main.cpp`.

Additionally when I look at our full stack of 100+ nodes, I never see `main.cpp` or `node.cpp` which is extra assurance. I'm happy to get counter-examples though

fujitatomoya:
thanks for the explanation! i was just curious about the background for this additional token. but i am not sure how much this is useful against full file path including performance overhead. l would like to see more feedback on this 👂

tonynajjar:
> but i am not sure how much this is useful against full file path

For us here is a before and after:

```
[INFO 1770889009.971313974] [/opt/auto_ws/src/auto-sandbox/src/vendor/navigation2/nav2_controller/src/controller_server.cpp:45] [controller_server]: Creating controller server
```

```
[INFO 1770887000.671869363] [controller_server.cpp:45] [controller_server]: Creating controller server
```

With one line it might not seem like a big deal but with thousands it compounds pretty badly and makes the logs unreadable

fujitatomoya:
that is good example 👍

src/logging.c
```diff
@@ -393,6 +428,7 @@ static const token_map_entry_t tokens[] = {
   {.token = "message", .handler = expand_message},
   {.token = "function_name", .handler = expand_function_name},
   {.token = "file_name", .handler = expand_file_name},
+  {.token = "short_file_name", .handler = expand_short_file_name},
```
fujitatomoya, line 430: i would register "short_file_name" before "file_name", because if it does prefix matching with the shorter token's length, "short_file_name" could match "file_name" first. but i do not think this is a real issue here, because it uses `strcmp` to complete match check.

src/logging.c
```diff
@@ -381,6 +381,41 @@ static const char * expand_file_name(
...
+static const char * expand_short_file_name(
+  const logging_input_t * logging_input,
+  rcutils_char_array_t * logging_output,
+  size_t start_offset, size_t end_offset)
+{
+  (void)start_offset;
+  (void)end_offset;
+
+  if (logging_input->location) {
+    const char * file_name = logging_input->location->file_name;
```
fujitatomoya, line 393: should we check if this is not NULL to avoid undefined behavior?

src/logging.c
```diff
@@ -381,6 +381,41 @@ static const char * expand_file_name(
...
+    const char * basename = file_name;
+    const char * last_sep = strrchr(file_name, '/');
+    if (last_sep != NULL) {
+      basename = last_sep + 1;
+    }
+#ifdef _WIN32
+    const char * last_backslash = strrchr(file_name, '\\');
+    if (last_backslash != NULL && last_backslash > last_sep) {
```
fujitatomoya: last_sep could be null?

```suggestion
    if (last_backslash != NULL && (last_sep == NULL || last_backslash > last_sep)) {
```

fujitatomoya (commented):
overall lgtm, i got a few comments.

@tonynajjar i can see @kscottz 's approval including doc support, so let's proceed !!! thanks for waiting 🙇

kscottz:
@tonynajjar, looks like this is ready to go pending resolved suggestions. Can you resolve the suggestions so we can ship this to the community. 

Also, thanks a bunch! This is great.

tonynajjar:
I somehow completely missed @fujitatomoya's review! Thanks for the ping, i'll address the review

src/logging.c
tonynajjar: Defensive programming, good idea :)

src/logging.c
tonynajjar: sure, I added it in `expand_file_name` and `expand_function_name` to match

fujitatomoya (approved):

fujitatomoya:
Pulls: ros2/rcutils#541
Gist: https://gist.githubusercontent.com/fujitatomoya/f8472cc6d61becdacf4f94a820ccd71f/raw/6f4ef84e9b0ba2f65db32e85b58ee38b463005d8/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18269
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27752/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20787/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7618/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26751/)

tonynajjar:
@fujitatomoya is the test failing because of this PR? As far as I see it failed because of another reason, can you restart it?

```
rclcpp
02:17:07 Starting >>> rclcpp
02:17:07 FATAL: command execution failed
FATAL: Unable to delete script file /tmp/jenkins8445192309629836075.sh
```

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27763/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20793/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26756/)

test/test_logging_console_output_handler.cpp
```diff
@@ -95,3 +97,82 @@ TEST(TestLoggingConsoleOutputHandler, bad_inputs) {
...
+    << "Expected basename in output: " << output;
+  EXPECT_EQ(std::string::npos, output.find("/some/long/path/to/"))
+    << "Full path should not appear in output: " << output;
+}
+
+TEST(TestLoggingConsoleOutputHandler, short_file_name_without_path_unchanged) {
+  setenv("RCUTILS_CONSOLE_OUTPUT_FORMAT", "{short_file_name}", 1);
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    unsetenv("RCUTILS_CONSOLE_OUTPUT_FORMAT");
```
ahcorde: This is not working on Windows

ahcorde (changes requested):

test/test_logging_console_output_handler.cpp
tonynajjar: Pushed a fix, can you try again

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27767/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20796/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7629/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26760/)

tonynajjar:
What do we do with "unstable"?

fujitatomoya:
@tonynajjar https://ci.ros2.org/job/ci_windows/26760/ is totally unrelated to this PR, i will go ahead to merge this. thanks for your contribution!

tonynajjar:
@fujitatomoya can you please trigger a backport to Kilted? Should be ABI/API compatible right?

fujitatomoya:
@Mergifyio backport kilted

PR #550, opened 2026-04-10
Use less common variable name in macro

sloretz:
## Description

Part of https://github.com/ros2/rcl/pull/1312

The variable name `ret` was colliding with other variables named `ret` in the above PR's new macro, and that caused warnings on Windows.

Yes, I used gemini to find this issue


fujitatomoya (approved):
good catch 👍

PR #570, opened 2026-05-28
Fix date_time_with_ms millisecond formatting

ahuazuipiaoliang:
## Summary

Fix `{date_time_with_ms}` formatting so the subsecond field is computed as milliseconds instead of taking the first three digits from an unpadded nanosecond string.

## Details

The previous implementation converted the nanosecond remainder to a decimal string without zero-padding and then used the first three characters. For values with leading zeros, such as `085860000 ns`, this produced `.858` instead of `.085`.

This PR computes milliseconds directly with `nanoseconds / 1000000` and formats the value with `%03`.

## Testing

Added coverage for a timestamp whose nanosecond remainder has leading zeros.

Closes #569

PR #571, opened 2026-05-28
Fix date_time_with_ms millisecond formatting

ahuazuipiaoliang:
## Description
refer to #570


fujitatomoya:
@ahuazuipiaoliang we do not have to open the backport PR something like this. we can backport https://github.com/ros2/rcutils/pull/570 once it gets merged in our development branch rolling to the downstream branches including humble.

PR #586, opened 2026-08-14
Fix atomic assignment for clang on MacOS

wentasah:
## Description

Without this change, I get the following error on MacOS with clang 21.1.8:

    /nix/var/nix/builds/nix-30990-3110187581/rcutils-release-release-lyrical-rcutils-7.1.2-1/src/testing/fault_injection.c:19:63:
    error: illegal initializer type 'atomic_int_least64_t' (aka '_Atomic(int_least64_t)')
       19 | static atomic_int_least64_t g_rcutils_fault_injection_count = {-1};

no

### DIFF
--- a/src/testing/fault_injection.c
+++ b/src/testing/fault_injection.c
@@ -16,7 +16,7 @@
 
 #include "rcutils/stdatomic_helper.h"
 
-static atomic_int_least64_t g_rcutils_fault_injection_count = {-1};
+static atomic_int_least64_t g_rcutils_fault_injection_count = -1;
 
 void rcutils_fault_injection_set_count(int_least64_t count)
 {


yukaitung:
I've tested it on my mac. This PR resolves the issue.✅

needed to backport to lyrical

fujitatomoya (commented):
@wentasah @yukaitung thanks for creating PR and verification on macOS. but i am almost sure that this PR breaks the windows, i think using structure there is intentionally. see the following code.

https://github.com/ros2/rcutils/blob/049298ff8c19cbe0611b5d782048e0e949f808d5/include/rcutils/stdatomic_helper/win32/stdatomic.h#L86-L87

after all, i guess right thing to do here is that using the initializer macro the helpers already provide instead of picking one literal form.

fujitatomoya:
@wentasah @yukaitung can you take a look at https://github.com/ros2/rcutils/pull/587? test on macOS would be appreciated.

wentasah:
Closing in favor of #587.

PR #589, opened 2026-08-23
Remove the bundled gcc stdatomic compatibility header

mjcarroll:
## Description

`include/rcutils/stdatomic_helper/gcc/stdatomic.h` is a vendored copy of the FreeBSD `stdatomic.h`, added so rcutils could build on compilers that predated a working `<stdatomic.h>`. Its own banner records the intent:

> It is a modification of the FreeBSD stdatomic header with some modifications so that it will work with gcc 4.7+ and Clang 3.1+.

The guards in `stdatomic_helper.h` select it in exactly two cases:

```c
#if !defined(__clang__) && defined(__GNUC__) && __GNUC__ <= 4 && __GNUC_MINOR__ <= 9
#   if defined(__has_feature) && !my__has_feature(c_atomic)
```

GCC 4.9 and older, or Clang that does not report the `c_atomic` feature — Clang older than 3.1. Neither has been a supported platform for years; Noble ships GCC 13. On every compiler Rolling targets, both branches are already dead and `<stdatomic.h>` is what gets included.

This deletes the header and collapses the include selection to a single choice:

```c
#if !defined(_WIN32) || defined(__MINGW64__)
# if __cplusplus
#   error "cannot be used with C++ ..."
# else
#   include <stdatomic.h>
# endif
#else
# include "./stdatomic_helper/win32/stdatomic.h"
#endif
```

The `my__has_feature` shim goes away with its only caller.

### The win32 shim stays

MSVC 14.51 does ship `<stdatomic.h>`, but including it without `/experimental:c11atomics` is a hard error:

```
vcruntime_c11_stdatomic.h(12): fatal error C1189: #error: "C atomic support is not enabled"
```

That is worth revisiting when MSVC's C11 atomics leave experimental, but not here.

No behavior change on any supported platform. Strictly, the deleted header was installed under `include/rcutils/`, so anything that included it by path would break — nothing in the core stack does.

Yes, Claude Opus 5.


wjwwood (approved):
Needs AI attribution in description.

ahcorde (approved):

ahcorde:
Pulls: ros2/rcutils#589
Gist: https://gist.githubusercontent.com/ahcorde/113fab9b42933df540a484901131fee9/raw/c5f8e1c905d807c90c9adf1d0659ba413682760f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20190
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30261/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22955/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10071/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29132/)

PR #603, opened 2026-09-21
Fix data race in concurrent first-time logging initialization

thomasmoore-torc:
## Problem

`rcutils_logging_initialize_with_allocator()` checks `g_rcutils_logging_initialized` (a plain, non-atomic `bool`) and, if false, goes straight into building `g_rcutils_logging_severities_map` via `rcutils_hash_map_init()`. This is a check-then-act race: multiple threads calling this function for the first time concurrently can all observe the flag as `false` and all build the map at once, corrupting it.

We hit this as a `SIGSEGV` inside `rcutils_hash_map_init()`, reproduced with ThreadSanitizer and a plain stress test (many threads each constructing an independent `rosbag2_compression::SequentialCompressionWriter` for the first time, simultaneously). This is easy to hit in practice because `rcutils` is typically statically linked into every `pluginlib`-loaded `.so`: each such plugin library gets its own private copy of `g_rcutils_logging_initialized`, so even one logging call from each of several plugins, made concurrently for the very first time, races.

This is a different bug from #397 / #427: those are about `rcutils_logging_get_logger_level()`/`set_logger_level()` racing with each other on the severities map *after* initialization. This is about concurrent *first-time initialization itself*, before that map safely exists at all. I checked #427's diff directly — it does not touch `rcutils_logging_initialize_with_allocator()`'s entry point, so it would not fix this even if merged.

## Fix

- Add a spinlock (`atomic_bool`, using this file's existing `rcutils_atomic_*` helpers) around the actual first-time initialization work, so only one caller ever builds the severities map.
- Convert every read/write of `g_rcutils_logging_initialized` in `logging.c` and the `RCUTILS_LOGGING_AUTOINIT` macro to the GCC/Clang `__atomic_load_n`/`__atomic_store_n` builtins, so the writes performed during initialization are properly published (acquire/release) to concurrent readers, not just the map-build step itself.
- `g_rcutils_logging_initialized` keeps its plain `bool` declared type in the public header — `rcutils/logging.h` is included from C++ translation units, which cannot include `<stdatomic.h>` / `rcutils/stdatomic_helper.h` (see the `#error` for `__cplusplus` in that header). The GCC/Clang builtins work on ordinary objects and need no such include, so this avoids that conflict entirely.

## Testing

Added `test_logging_concurrent_first_time_initialize` to `test/test_logging.cpp`. It holds a batch of threads at a barrier and releases them together so they actually land in `rcutils_logging_initialize()` at once — starting threads in a plain loop isn't enough, since a single unsynchronized call is fast enough that earlier threads finish before later ones even start, closing the race window before it can be hit. This is repeated over 20 rounds, since any single round can still get unlucky. It's a different scenario from the existing `test_logging_macro_thread_safety`, which calls `rcutils_logging_initialize()` once, single-threaded, *before* spawning threads, so it never exercises the first-time race at all.

I built and ran this locally against a `ros:rolling-ros-base` container (colcon, `-DBUILD_TESTING=ON`): unpatched, the new test crashes with `SIGSEGV` on every run (15/15); with the fix, it passes cleanly (15/15), and the full `colcon test --packages-select rcutils` suite passes 100% (52/52, 2 benchmark tests skipped as expected). While testing this way I also caught and fixed a real portability bug the syntax-check alone had missed: `ATOMIC_VAR_INIT`, used for the new spinlock's initializer, was removed in newer C standards and fails to compile on a newer glibc/GCC toolchain than the one I originally wrote this against — replaced with plain static-initializer syntax (`= false`), which is valid C11 and portable.

## Generative AI

This fix was investigated, written, and iteratively verified (ThreadSanitizer, stress-testing, and a local colcon/ROS 2 rolling build) by Claude (Anthropic), operating as Claude Code, at the direction of and reviewed by me. The regression test and this PR description were also written by Claude; I reviewed both before submitting.


thomasmoore-torc:
For reference, tracking the related issue/PR mentioned above:

- #397 (issue, still open, no fix merged)
- #427 (PR, still open/unmerged): https://github.com/ros2/rcutils/pull/427

Neither addresses the concurrent first-time initialization race this PR fixes (see description for why).

